PR #2, opened 2021-08-02
Replace rollup with tsc

bonustrack:


bonustrack:
[Remove rollup from snapshot-strategies repo and use tsc instead](https://trello.com/c/Wiq1pnDS/252-remove-rollup-from-snapshot-strategies-repo-and-use-tsc-instead)

PR #3, opened 2021-08-02
[cake] feat(cake): Minimum voting power

hachiojidev:
- Adds minimum voting power to fix issues with our subgraph


bonustrack (approved):
Looks good!

PR #5, opened 2021-08-03
[anti-whale] Add anti-whale strategy

joaomajesus:
- add an anti-whale strategy that executes a configured strategy and applies an anti-whale measure to its results to reduce the impact of big wallets in the the resulting value, reducing the effect on the voting power as the token amount increases.


src/strategies/anti-whale/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+    result = result * thresholdMultiplier
+  }
+  ```
+
+## Accepted options
+
+- **strategy.name:** name of the strategy to run
+
+- **strategy.network:** the network id to be used by the strategy.
```
bonustrack, line 24: The issue with that is that the space has already a network defined, and the snapshot block number is specific to this network, if you use Polygon network on the strategy but the space use mainnet the snapshot will be completely off. I'd rather not add any network param here.

joaomajesus: ahh I see.... good one, I guess that's why the multichain strategy made some calls to the Graph to get the blocks for each network :)

bonustrack: Exactly!

joaomajesus: It's done. Thanks for the help!

src/strategies/anti-whale/index.ts
```diff
@@ -0,0 +1,86 @@
+import getProvider from '../../utils/provider';
```
ChaituVR, line 1:
```suggestion
```

ChaituVR (commented):
LGTM

ChaituVR (approved):

ChaituVR:
Thanks for the PR :) strategy is live now https://snapshot.org/#/strategy/anti-whale

PR #11, opened 2021-08-06
[l2-deversifi] L2-deversifi strategy

nicob21:
Implementation of l2-deversifi strategy


src/strategies/l2-deversifi/index.ts
```diff
@@ -0,0 +1,49 @@
...
+        }
+      })
+    );
+  });
+
+  const results = await Promise.all(promises);
+  const resultsJson = await Promise.all(results.map((r: any) => r.json()));
+
+  const data: any = resultsJson.reduce(
+    (res: any, item: any) => [...res, ...item.score],
```
ChaituVR, line 42: There is a error here, which is failing the test cases, would be great if you can fix it
```TypeError: item.score is not iterable

      28 |     const results = await Promise.all(promises);
      29 |     const resultsJson = await Promise.all(results.map((r) => r.json()));
    > 30 |     const data = resultsJson.reduce((res, item) => [...res, ...item.score], []);
```

src/strategies/l2-deversifi/index.ts
```diff
@@ -0,0 +1,49 @@
...
+  const { api, token, limit = 300 } = options;
+
+  const pages = Math.ceil(addresses.length / limit);
+  const promises: any = [];
+
+  let api_url = api;
+  api_url += `?blockNumber=${snapshot}`;
+  api_url += `&token=${token}`;
+
+  Array.from(Array(pages)).forEach((x, i) => {
```
ChaituVR, line 25: Looks like here you are trying to make a API fetch call for each address, which won't scale for proposals with more addresses, is there any other approach you can try? Where you can get data in a single request

ChaituVR (changes requested):

src/strategies/l2-deversifi/index.ts
plutoegg: This is not per address, but per 300 address (so if its more than 300 addresses requested it splits it into multiple requests).

ChaituVR: @plutoegg Cool then :) just noticed that, Thanks for letting me know

ChaituVR (changes requested):
@plutoegg please resolve conflicts :)

plutoegg:
> @plutoegg please resolve conflicts :)

Done!

ChaituVR (commented):
LGTM

plutoegg:
Does another snapshot team member need to merge? We are keen to test it out with the UI :)

bonustrack (approved):

ChaituVR (approved):

bonustrack:
@plutoegg Thanks for your patience will get the strategy up in a moment

ChaituVR:
@plutoegg Strategy is live here: https://snapshot.org/#/strategy/l2-deversifi Thanks :)

PR #18, opened 2021-08-09
[blacklist] Add a blacklist strategy

joaomajesus:
- Add a blacklist strategy

### DIFF
--- a/src/strategies/blacklist/README.md
+++ b/src/strategies/blacklist/README.md
@@ -1,0 +1,45 @@
+# Anti-Whale strategy
+
+## Description
+
+This strategy executes a configured strategy and applies a blacklist to it's results effectively setting the result to 0 for any address set in the blacklist.
+
+## Accepted options
+
+- **strategy.name:** name of the strategy to run
+
+- **strategy.params:** the strategy specific params to execute it.
+
+- **blacklist:** An array of st6rings representing wallet addresses to be blacklisted an their result set to 0
+  
+## Examples
+
+```json
+[
+  {
+    "name": "blacklist",
+    "strategy": {
+      "name": "blacklist",
+      "params": {
+        "strategy": {
+          "name": "erc20-balance-of",
+           "params": {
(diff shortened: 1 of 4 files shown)


src/strategies/blacklist/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+This strategy executes a configured strategy and applies a blacklist to it's results effectively setting the result to 0 for any address set in the blacklist.
+
+## Accepted options
+
+- **strategy.name:** name of the strategy to run
+
+- **strategy.params:** the strategy specific params to execute it.
+
+- **blacklist:** An array of st6rings representing wallet addresses to be blacklisted an their result set to 0
```
ChaituVR, line 13:
```suggestion
- **blacklist:** An array of strings representing wallet addresses to be blacklisted an their result set to 0
```

ChaituVR (commented):
Looks good 👍

src/strategies/blacklist/index.ts
```diff
@@ -0,0 +1,38 @@
+import strategies from '..';
+
+export const author = 'joaomajesus';
+export const version = '1.0.0';
+
+let _options;
+
+function applyBlackList(address: string, result) {
+  return (_options.blacklist as Array<string>).includes(address) ? 0 : result;
```
bonustrack, line 9: Can you make it work for both checksumed and not checksumed addresses.

joaomajesus: @bonustrack  I'll check that. 
I actually have another change in mind. so that the blacklisted addresses don't go through the decorated strategy

joaomajesus: sorry for my ignorance... :) but does that mean that I should use Address instead of string?

ChaituVR: @joaomajesus use below function:
```js
const checksumedAddress = getAddress(address);
```
using this you can convert an address into a checksummed address

ChaituVR: @joaomajesus facing any issue? would love to help :)

PR #19, opened 2021-08-10
[delegation] Fix delegation strategy score calculation issue

ChaituVR:
Here  if there is no address in the last strategy score (of delgation)  `x + y[b] || 0` it replaces the score with `0` which is making score of many addresses `0`
```
- (a, b) => a + scores.reduce((x, y) => x + y[b] || 0, 0),
+ (a, b) => a + scores.reduce((x, y) => y[b] ? (x + y[b]) : (x + 0), 0),
 ```

### DIFF
--- a/src/strategies/delegation/index.ts
+++ b/src/strategies/delegation/index.ts
@@ -1,10 +1,8 @@
 import { getDelegations } from '../../utils/delegation';
 import { getScoresDirect } from '../../utils';
-import examplesFile from './examples.json';
 
 export const author = 'bonustrack';
 export const version = '0.1.0';
-export const examples = examplesFile;
 
 export async function strategy(
   space,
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by ChaituVR** [comment]
[Voting power from delegation not counted](https://trello.com/c/nQqZRHMW/266-voting-power-from-delegation-not-counted)

File: src/strategies/delegation/index.ts [hunk 1]
```diff
@@ -39,7 +39,7 @@ export async function strategy(
     addresses.map((address) => {
       const addressScore = delegations[address]
         ? delegations[address].reduce(
-            (a, b) => a + scores.reduce((x, y) => x + y[b] || 0, 0),
+            (a, b) => a + scores.reduce((x, y) => y[b] ? (x + y[b]) : (x + 0), 0),
```
bonustrack, line 42: Why x + 0 and not x directly?

ChaituVR: Oops! Yes, how did i miss that, changing

bonustrack (approved):

PR #20, opened 2021-08-10
[sushiswap] Optionally use staked LP tokens

maxaleks:
This improved strategy also uses the balances of the LP tokens sent to the Master Chef contract

### DIFF
--- a/src/strategies/sushiswap/index.ts
+++ b/src/strategies/sushiswap/index.ts
@@ -1,12 +1,123 @@
 import { getAddress } from '@ethersproject/address';
+import { formatUnits } from '@ethersproject/units';
 import { subgraphRequest } from '../../utils';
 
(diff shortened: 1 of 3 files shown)


src/strategies/sushiswap/index.ts
```diff
@@ -76,5 +194,18 @@ export async function strategy(
         });
     });
   }
+  if (options.useStakedBalances === 'true') {
```
ChaituVR, line 197: Would be great if you can add a small description in README.md of this strategy, that can explain what this option can do

ChaituVR (changes requested):

bonustrack (approved):
Looks great! Thanks for doing this @maxaleks

ChaituVR (approved):

PR #28, opened 2021-08-15
[multichain] fix params in multichain strategy example to include token name

brianfakhoury:
When you use the snapshot frontend, the parsing won't allow this strategy's params to be saved without the "symbol" property.

- add "symbol" property to multichain strategy example params.


ChaituVR:
@brianfakhoury Thanks for the PR :) Nice fix but `"<Token Symbol>"` is displayed weird in the playground UI so changed it to `MULTI`
image

bonustrack (approved):

PR #31, opened 2021-08-19
[mcn-farm] Add mcn-farm strategy

drop-out-dev:
This is the custom strategy for MCN Ventures.

MCN tokens must be staked in the [farm](https://farm.mcn.ventures) to have voting power. LP tokens and pending reward tokens are also counted for voting.

### DIFF
--- a/src/strategies/mcn-farm/index.ts
+++ b/src/strategies/mcn-farm/index.ts
@@ -25,7 +25,10 @@ export async function strategy(
   snapshot
 ) {
   const blockTag = typeof snapshot === 'number' ? snapshot : 'latest';
-  const pools = await call(provider, abi, [FARM_ADDRESS, 'getPoolList', []]);
+  const lpAddress = options.lpAddress || MCN_LP_ADDRESS;
+  const tokenAddress = options.tokenAddress || MCN_ADDRESS;
+  const farmAddress = options.stakingAddress || FARM_ADDRESS;
+  const pools = await call(provider, abi, [farmAddress, 'getPoolList', []]);
   const flatten = (arr) => [].concat.apply([], arr);
   const product = (...sets) => {
     return sets.reduce(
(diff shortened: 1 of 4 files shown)


src/strategies/mcn-farm/index.ts
```diff
@@ -0,0 +1,117 @@
...
+        internalType: 'struct IBonusRewards.User',
+        name: '',
+        type: 'tuple'
+      },
+      { internalType: 'uint256[]', name: '', type: 'uint256[]' }
+    ],
+    stateMutability: 'view',
+    type: 'function'
+  }
+];
```
ChaituVR, line 59:
```suggestion
const abi = [
  'function totalSupply() view returns (uint256)',
  'function balanceOf(address account) view returns (uint256)',
  'function getPoolList() view returns (address[])',
  'function getUser(address _lpToken, address _account) view returns (tuple(uint256 amount, uint256[] rewardsWriteoffs), uint256[])'
];
```

src/strategies/mcn-farm/index.ts
```diff
@@ -0,0 +1,117 @@
+import { formatUnits } from '@ethersproject/units';
+import { call, multicall } from '../../utils';
+import { BigNumber } from '@ethersproject/bignumber';
+
+export const author = 'drop-out-dev';
+export const version = '0.1.0';
+
+const FARM_ADDRESS = '0x15dEd15fE32EBac0b6cFb08cdAB112cca8380423';
+const MCN_ADDRESS = '0xD91E9a0fEf7C0fa4EBdAF4d0aCF55888949A2a9b';
+const MCN_LP_ADDRESS = '0x2Ef2cb6af83de4171A69EE2f7C677079fFD9BcD0';
```
ChaituVR, line 10: would it be good to have these addresses passed from params? (but also still have these as defaults) so others can pass their own addresses if they want to, or it doesn't make sense in this strategy?

ChaituVR (changes requested):

src/strategies/mcn-farm/index.ts
drop-out-dev: Yes, good idea. I've updated accordingly.

drop-out-dev:
> Hi @drop-out-dev, please resolve conflicts, I don't have access to this branch to push a commit

I know you've already removed the changes requested label, but it seems like merging is still blocked. Could you mark the change request as addressed please?

ChaituVR (approved):

ChaituVR:
Strategy is live: https://snapshot.org/#/strategy/mcn-farm

PR #32, opened 2021-08-19
[cream] Update cream strategy

jeremyHD:
- Update cream voting score strategy
- Only CREAM tokens staked in long term pool or iceCREAM contract has voting power


jeremyHD:
One silly question here, could I delete the old strategy methods?
Once proposal is uploaded to IPFS, the old strategy will also be uploaded right?

jeremyHD:
I've closed this PR since what I meant to do can be done with `erc20-balance-of`

PR #34, opened 2021-08-21
[meebitsdao] Add meebitsdao strategy

peters-josh:
Fixes # .

- Add meebitsdao voting strategy. Checks token's metadata for "Activated" property.

### DIFF
--- a/src/strategies/meebitsdao/index.ts
+++ b/src/strategies/meebitsdao/index.ts
@@ -36,7 +36,7 @@ export async function strategy(
   const resp = await fetch(options.apiUrl);
   const tokenStatus = await resp.json();
 
-  async function checkActivated(address) {
+  function checkActivated(address: any) {
     const index = ownerResponse.findIndex(
       (res: any) => res.owner.toLowerCase() === address.toLowerCase()
     );
@@ -49,10 +49,7 @@ export async function strategy(
     return 0;
   }
 
-  const results = addresses.map(async (address: any) =>
-    checkActivated(address)
-  );
-  const votes = await Promise.all(results);
+  const votes = addresses.map((address: any) => checkActivated(address));
 
   const scores = {};
   votes.map((value, i) => {
--- a/src/strategies/meebitsdao/index.ts
+++ b/src/strategies/meebitsdao/index.ts
@@ -29,13 +29,12 @@ export async function strategy(
     ownerCalls.push([options.address, 'ownerOf', [i]]);
   }
 
-  const uriResponse = await multicall(network, provider, abi, uriCalls, {
-    blockTag
-  });
-
   const ownerResponse = await multicall(network, provider, abi, ownerCalls, {
     blockTag
   });
+
+  const resp = await fetch(options.apiUrl);
+  const tokenStatus = await resp.json();
 
   async function checkActivated(address) {
     const index = ownerResponse.findIndex(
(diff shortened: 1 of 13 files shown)


src/strategies/meebitsdao/index.ts
```diff
@@ -0,0 +1,70 @@
...
+    const metadata = await resp.json();
+    const activated = metadata.attributes[2];
+    if (activated.trait_type === 'Activated' && activated.value === 'True') {
+      return 1;
+    }
+    return 0;
+  }
+
+  const results = addresses.map(async (address: any) =>
+    checkActivated(address)
```
ChaituVR, line 61: Hi @peters-josh The strategy should not send a request for each voters, this doesn't scale for more addresses . Can you find out any alternative way to get for all addresses in a single call

https://docs.snapshot.org/strategies/create#4-make-sure-you-pass-the-checklist

ChaituVR (changes requested):

src/strategies/meebitsdao/index.ts
peters-josh: Hi @ChaituVR 

This strategy is dealing with erc721 tokens — every one of which has unique metadata. For every token, have have to get the metadata from the token uri to to check if it contains the specific property we are looking for. I do not believe there is a way around this requirement.

In practice, there are only a maximum of 200 tokens we would ever call, and in reality it will be much less as every token holder will not vote on every issue. The call to fetch the metadata is only made after we confirm the address has a token as well, so this call will not be made on every address.  

What do you suggest as an alternative when creating a strategy that reads ERC721 metadata if we cannot get the metadata for the unique tokens we are dealing with?

ChaituVR: There will max 200 tokens, but problem is if 200 addresses own it and they vote, Snapshot's server will be sending 200 requests to IPFS every time someone opens the proposal, ofcourse what you said is also makes sense as every token holder will not vote on every issue. but should look at scalability at maximum possibility

peters-josh: As addresses in the Discord, this metadata is dynamic so the metadata for every token will need to be checked. It is possible to aggregate the metadata, but this introduces an element of centralization which is not ideal.

bonustrack: Do you have that data on a subgraph by chance?

peters-josh: @bonustrack @ChaituVR I updated the strategy so it no longer pulls the metadata for each address (uses one centralized API at the beginning instead).

src/strategies/meebitsdao/index.ts
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  const ownerResponse = await multicall(network, provider, abi, ownerCalls, {
+    blockTag
+  });
+
+  const resp = await fetch(options.apiUrl);
+  const tokenStatus = await resp.json();
+
+  async function checkActivated(address) {
```
bonustrack, line 39: Any reason to use `async` here? I don't see any async function, i assume you can remove it and also remove the Promise.all on line 55

peters-josh: Ah this was leftover from the old code. Removed.

bonustrack (approved):
Looks good!

ChaituVR (approved):

ChaituVR:
@peters-josh Thanks for the changes :) 
Strategy is live here: https://snapshot.org/#/strategy/meebitsdao ✨

PR #47, opened 2021-08-30
[dfyn-lp] dfyn-lp strategy added

vatsalgupta13:
Fixes # .

- dfyn-lp strategy added

### DIFF
--- a/src/strategies/dfyn-lp/index.ts
+++ b/src/strategies/dfyn-lp/index.ts
@@ -1,0 +1,81 @@
+import { getAddress } from '@ethersproject/address';
+import { subgraphRequest } from '../../utils';
+
+const DFYN_SUBGRAPH_URL = {
+  '137': 'https://api.thegraph.com/subgraphs/name/ss-sonic/dfyn-v5'
+};
+
+export const author = 'vfatouros';
+export const version = '0.1.0';
+
+export async function strategy(
+  _space,
+  network,
+  _provider,
+  addresses,
+  options,
+  snapshot
+) {
+  const params = {
+    users: {
+      __args: {
+        where: {
+          id_in: addresses.map((address) => address.toLowerCase())
+        },
+        first: 1000
+      },
+      id: true,
+      liquidityPositions: {
+        __args: {
+          where: {
+            liquidityTokenBalance_gt: 0
+          }
+        },
+        liquidityTokenBalance: true,
+        pair: {
+          id: true,
+          token0: {
+            id: true
+          },
+          reserve0: true,
+          token1: {
+            id: true
+          },
+          reserve1: true,
+          totalSupply: true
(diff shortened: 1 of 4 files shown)


src/strategies/dfyn-lp/index.ts
```diff
@@ -0,0 +1,81 @@
+import { getAddress } from '@ethersproject/address';
+import { subgraphRequest } from '../../utils';
+
+const DFYN_SUBGRAPH_URL = {
+  '137': 'https://api.thegraph.com/subgraphs/name/ss-sonic/dfyn-v5'
+};
+
+export const author = 'vfatouros';
```
ChaituVR, line 8:
```suggestion
export const author = 'vatsalgupta13';
```

ChaituVR (changes requested):
@vatsalgupta13 I see this is built on top of `infinityprotocol-liquidity-pools`  strategy, is it possible to modify `infinityprotocol-liquidity-pools`  strategy itself with `if` condition for `scoreMultiplier` and to pass subgraph url from options, which can reduce duplicate code,
image

src/strategies/dfyn-lp/index.ts
vatsalgupta13: So basically its actually based on staked-uniswap strategy. But they hardcoded the subgraph url so I had to change the code.

ChaituVR: > So basically its actually based on staked-uniswap strategy. But they hardcoded the subgraph url so I had to change the code.

Ya I can understand that hardcoding is not good, but i see this is minor change so we can always change the strategy and pass optional params as per our requirement 

example:
```
await subgraphRequest(	
    options.subGraphURL ? options.subGraphURL : INFINITYPROTOCOL_SUBGRAPH_URL[network],	
    params
```

```
if(options.scoreMultiplier) {
  score[userAddress] = score[userAddress] * options.scoreMultiplier
}
```
and also would be great if we can document these options inside README.md

vatsalgupta13: Sure we can do that. Should I make the changes in staked-uniswap strategy or infinityprotocol-liquidity-pools strategy, both of them are actually the same with different subgraph requests.

ChaituVR: Any strategy should be fine as long as the previous strategy works well and your changes are small, optional, and well documented (doesn't disturb existing voting power calculation) make sure to add some test addresses into examples.json that returns a positive score with your changes

vatsalgupta13: Sure. Will do that.

vatsalgupta13:
Made changes to the infinityprotocol-liquidity-pools strategy. Closing this PR.

PR #48, opened 2021-08-30
[dfyn-staked-in-farms] dfyn-staked-in-farms strategy added

vatsalgupta13:
Fixes # .

- new strategy to calculate score for dfyn tokens staked in dfyn farms


src/strategies/dfyn-staked-in-farms/index.ts
```diff
@@ -0,0 +1,145 @@
...
+  }
+}
+
+const getLP = async (stakedLP, network) => {
+  let lp = []
+  let results = []
+  do {
+    results = await getAllLP(lp.length, stakedLP, network);
+    lp = lp.concat(results)
+  } while (results.length === 1000);
```
ChaituVR, line 53: if `results` returned by subgraph are not equal to 1000, this will be a never-ending loop right? maybe you mean to add `lp.length < 1000`? Let me know if i am thinking in wrong way

ChaituVR (changes requested):

src/strategies/dfyn-staked-in-farms/index.ts
vatsalgupta13: On the contrary, loop will only continue if `results.length === 1000`, otherwise it will break. Since subgraph returns at max 1000 items in one call, I am checking if `results.length === 1000`. If it is, then it needs to fetch the items after those 1000 items. If less than 1000 items are returned, the loop will break since `results.length` will not be equal to 1000. Either way, this piece of code has been included to make the strategy future proof. Currently there are only around 10 Dfyn farms so only 10 liquidity pairs are returned.

ChaituVR: oh now I get it, I was overlooking :)

src/strategies/dfyn-staked-in-farms/index.ts
```diff
@@ -0,0 +1,145 @@
...
+  '137': 'https://api.thegraph.com/subgraphs/name/ss-sonic/dfyn-v5'
+};
+
+const getAllLP = async (skip, stakedLP, network) => {
+  let params =
+  {
+    pairs: {
+      __args: {
+        where: {
+          id_in: stakedLP.map((address) => address.toLowerCase())
```
bonustrack, line 19: In this query you should add a way to query specific block number if the snapshot is not "latest". Currently it may return incorrect result.

vatsalgupta13: Thanks for pointing this out. I have made the required changes.

ChaituVR:
@vatsalgupta13 Please resolve conflicts :)

vatsalgupta13:
> @vatsalgupta13 Please resolve conflicts :)

Done.

bonustrack (approved):

ChaituVR (approved):

ChaituVR:
Strategy is live: https://snapshot.org/#/strategy/dfyn-staked-in-farms

PR #56, opened 2021-09-13
Update SpookySwap strategy to include xBOO

ooGwei:
Updated strategy to add additional token and staking pools

### DIFF
--- a/src/strategies/spookyswap/index.ts
+++ b/src/strategies/spookyswap/index.ts
@@ -1,5 +1,6 @@
+import { BigNumber } from '@ethersproject/bignumber';
 import { formatUnits } from '@ethersproject/units';
-import { Multicaller } from '../../utils';
+import { Multicaller, call } from '../../utils';
 
 export const author = 'ooGwei';
 export const version = '0.1.0';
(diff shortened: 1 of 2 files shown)


src/strategies/spookyswap/index.ts
```diff
@@ -71,7 +74,33 @@ const abi = [
...
+  },
+  {
+    name: 'xBOOForBOO',
+    inputs: [
+      {
+        internalType: 'uint256',
+        name: '_xBOOAmount',
+        type: 'uint256'
+      }
+    ],
```
ChaituVR, line 86: Hello, @ooGwei please replace your abi with this: `
````
[
  'function balanceOf(address) view returns (uint256 amount)',
  'function userInfo(uint256, address) view returns (uint256 amount, uint256 rewardDebt)',
  'function xBOOForBOO(uint256 _xBOOAmount) view returns (uint256 booAmount_)',
  'function poolLength() view returns (uint256)'
]

ChaituVR (changes requested):
I see an active proposal on `spookyswap.eth` which is using this strategy, you sure these changes won't affect that?

ooGwei:
> I see an active proposal on `spookyswap.eth` which is using this strategy, you sure these changes won't affect that?

We are willing to affect the current active proposal with these changes. The bigger importance for us is to include votes that are currently not counted, but should be.

ChaituVR (changes requested):
@ooGwei Strategy fails to calculate voting power when there are more than 500 addresses

````
missing revert data in call exception (error={"reason":"processing response error","code":"SERVER_ERROR","body":"{\"jsonrpc\":\"2.0\",\"id\":46,\"error\":{\"code\":-32000,\"message\":\"execution aborted (timeout = 5s)\"}}","error":{"code":-32000},"requestBody":"{\"method\":\"eth_call\",\"params\":[{\"to\":\"0x7f6a10218264a22b4309f3896745687e712962a0\",

PR #59, opened 2021-09-15
[dextf-staked-in-vaults] DEXTF: staked tokens

dextfprotocol:
Strategy to calculate number of staked DEXTF

### DIFF
--- a/src/strategies/dextf-staked-in-vaults/README.md
+++ b/src/strategies/dextf-staked-in-vaults/README.md
@@ -1,0 +1,42 @@
+# Multiple contract call strategy
+
+This strategy allows users to call the 'balanceOf' function across multiple DEXTF contracts (vaults) and performs summation over the results. By calling 'balanceOf', DEXTF vaults return the amount of $DEXTF staked by the user in that vault. This strategy will make a single multicall which will retrieve all users' staked balances in all of DEXTF vaults.
+
+## Example
+
+The space config will look like this:
+
+```JSON
+{
+  "strategies": [
+    ["dextf-staked-in-vaults", {
+      // vault contracts across which token balance needs to be calculated
+      "contractAddresses": [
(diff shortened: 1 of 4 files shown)


bonustrack (approved):

ChaituVR:
Strategy is live: https://snapshot.org/#/strategy/dextf-staked-in-vaults

dextfprotocol:
Guys, the new `dextf-staked-in-vaults` strategy works if we run the unit test, but it doesn't if we create a proposal with it.

We created a test proposal with the new strategy only and this setting:

```
{
  "symbol": "DEXTF (Staked)",
  "address": "0x5f64ab1544d28732f0a24f4713c2c8ec0da089f0",
  "decimals": 18
}
```

When we load the proposal and connect with a wallet with staked tokens, the Java Console shows this errors (that does not occur if we use the `erc20_balance_of strategy`):

```
POST https://score.snapshot.org/api/scores 500
TypeError: Cannot read properties of undefined (reading 'scores')
    at eS (vendor.f5d3977f.js:35)
    at async index.c2dad28d.js:119
    at async Y (index.c2dad28d.js:119)
```

Any idea?

ChaituVR:
@dextfprotocol check the params in example.json it should be something like:

```
{
  "name": "DEXTF Vaults",
  "contractAddresses": [
    "0x42a05787584ec09dDDe46f8CE6a715c93049ee88"
  ],
  "scoreMultiplier": 1,
  "methodABI": {
    "inputs": [
      {
        "internalType": "address",
        "name": "account",
        "type": "address"
      }
    ],
    "name": "balanceOf",
    "outputs": [
      {
        "internalType": "uint256",
        "name": "",
        "type": "uint256"
      }
    ],
    "stateMutability": "view",
    "type": "function"
  }
}
```
play with it here https://snapshot.org/#/playground/dextf-staked-in-vaults

PR #60, opened 2021-09-15
[flexa-capacity-staking] Add Flexa Capacity staking strategy

zkmoney:
Allows for Amp stake + reward balances held within Flexa Capacity to be reflected in voting power

### DIFF
--- a/src/strategies/flexa-capacity-staking/index.ts
+++ b/src/strategies/flexa-capacity-staking/index.ts
@@ -14,18 +14,27 @@ export async function strategy(
   options,
   snapshot
 ) {
-  return Object.fromEntries(await Promise.all(addresses.map(async (address) => {
-    const response = await fetch(`${options.apiBase}/accounts/${getAddress(address)}/totals?snapshot=${snapshot}`, {
-      method: 'GET',
-      headers: {
-        Accept: 'application/vnd.flexa.capacity.v1+json'
-      }
-    });
-    const data = await response.json();
-    const { supplyTotal = 0, rewardTotal = 0 } = data;
-    return [
-      address,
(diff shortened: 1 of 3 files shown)


src/strategies/flexa-capacity-staking/index.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export async function strategy(
+  space,
+  network,
+  provider,
+  addresses,
+  options,
+  snapshot
+) {
+  return Object.fromEntries(await Promise.all(addresses.map(async (address) => {
```
ChaituVR, line 17: looks like this will send on API call for one addresses, it won't scale when there are more votes on the proposal, can u plz find an alternative to this? maybe to get all data in a single API call

ChaituVR (changes requested):

src/strategies/flexa-capacity-staking/index.ts
zkmoney: Hi @ChaituVR, thanks for the note. 

I was able to use a batched version of the API, and you can see the change here https://github.com/amptoken/snapshot-strategies/commit/920e5eb7419abe8aa01875c1f2eede67e5e4a41d

Please let me know if you need anything else.

ChaituVR:
@zkmoney plz resolve conflicts, we should be good to go :)

zkmoney:
@ChaituVR No problem, just resolved them. 

Thanks!

bonustrack (approved):

ChaituVR (approved):

ChaituVR:
Strategy is live: https://snapshot.org/#/strategy/flexa-capacity-staking

PR #66, opened 2021-09-17
docs: Install dependencies using npm ci when package-lock is present

CjHare:
Fixes # .

With a `package-lock.json` present in the repo, `npm ci` will install with the dependencies contained within, while `npm install` will grab the latest versions of dependencies that may or may not match those from the lock file (generating a modified file as tracked by git, noticed this when building locally).

https://docs.npmjs.com/cli/v7/configuring-npm/package-lock-json
https://docs.npmjs.com/cli/v7/commands/npm-ci

The alternative approach can be to have the lock file in `.gitignore`, however that adds dependency versioning risk, or to use specific version of dependencies in the `package.json`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 #### Install dependencies
 ```bash
-npm install
+npm ci
 ```
 
 #### Build package


ChaituVR:
@bonustrack is this something we want to include? I think `npm install` works for us for now at least for this repo

bonustrack:
I'm not too familiar with `ci`, is this fixing an issue someone have currently?

ChaituVR:
@CjHare Does it solve any issue for you or this change is just an improvement suggestion?

ChaituVR:
@CjHare Thanks for the PR! but closing this for now as it is not required, please feel free to reopen this if it creating some issues without this :)

PR #67, opened 2021-09-18
[synthetix] Update existing strategy with the upcoming election values and create a non-weighted version

andytcf:


bonustrack (commented):
For the strategy `synthetix-non-quadratic` it should be added here: https://github.com/snapshot-labs/snapshot-strategies/blob/41c8202393c024ab05b86224af19c98dcfcbb169/src/strategies/index.ts

For the strategy `synthetix` is there some breaking changes? The strategy will run on previous proposals too, you may not want previous proposals to break or give different results.

ChaituVR (changes requested):
Hi @andytcf I see both the strategies are the same except having
```
 Math.sqrt(scaledWeighting);
```
Maybe you can have a helper like [this](https://github.com/snapshot-labs/snapshot-strategies/blob/master/src/strategies/uniswap-v3/helper.ts) and use common code in both strategies, it can help reduce the duplicate code

bonustrack (approved):

ChaituVR (approved):

ChaituVR:
new strategies live here:
https://snapshot.org/#/strategy/synthetix-quadratic
https://snapshot.org/#/strategy/synthetix-non-quadratic

PR #74, opened 2021-09-21
[hopr-bridged-balance] HOPRnet staking and cross-chain token balance strategies

QYuQianchen:
- Create two strategies for [HOPRnet](https://hoprnet.org/): Staking program and cross-chain token balances
- Both strategies can be used on both `xDai` or `mainnet`. Conversions of block number are implemented. A converted block number can then be passed in the subgraph query to get the score on the other chain.


ChaituVR (changes requested):
@QYuQianchen Would be great if you can make separate PRs for these two strategies, will be easy to review/integrate

QYuQianchen:
Thank you @ChaituVR, PR #75 and #76 are created accordingly.

PR #79, opened 2021-09-23
Custom network test support

CjHare:
Adds the ability to run the snapshot-strategies test using a local networks file. 
This adds supports dev & CI testing of strategies prior to governance contract deployment on a public (public test) networkk.


CjHare:
This setup I've got running with a Hardhat node (Ethereum chain) and deployment of the mainnet Multicall and a ERC20 token passing the default `erc20-balanceof` strategy with local changes (not included in the commit) to the exmaple.json.

Let me know what you think for the PR 🙂

src/utils/multicaller.ts
```diff
@@ -0,0 +1,45 @@
+import { StaticJsonRpcProvider } from '@ethersproject/providers';
+import set from 'lodash.set';
+import { multicall } from '../utils';
+
+export default class Multicaller {
```
ChaituVR, line 5: @CjHare Great work on PR! but multicall and Multicaller here causes duplicate code from snapshot.js, we should find some way to make them usable from one place only ( maybe try [patching snapshot.js](https://www.npmjs.com/package/patch-package) )

ChaituVR: @bonustrack any ideas? or these can be here? 🤔

ChaituVR (changes requested):

src/utils/multicaller.ts
CjHare: @ChaituVR you're quite right, the only way I could see avoiding the duplication would be to patch `snapshot.js` in some way. 
From a design perspective I can appreciate `snapshot.js` not needing such a change (as it's a change for the purpose of testing) ...but I can always start that conversation 🙂

CjHare:
Closed off as I've messed up the Git history for the branch (and this PR was/is blocked on the `snapshot.js` changess

PR #83, opened 2021-09-28
[mcn-farm] Update MCN strategy to handle different reward token positions

drop-out-dev:
Updated the strategy to handle more contract configurations


bonustrack (approved):

ChaituVR:
@drop-out-dev Strategy should be updated by now

PR #91, opened 2021-10-02
[comp-like-votes-inclusive] Add comp-like delegation strategy that accounts for undelegated balances

d1ll0n:
**Overview**

In Comp-based tokens, accounts do not receive voting power for token balances unless they self-delegate. This strategy simply incorporates undelegated balances into the `comp-like-votes` scoring strategy.

The scoring strategy uses a [deployed query contract](https://etherscan.io/address/0x95cb39a64390994dd8c1cc5d8a29dffde4212298#code) with a multicall function to speed up execution. The function takes a token address and a list of accounts and calculates each account's score with the following logic:
```
uint votes = token.getCurrentVotes(account);
if (token.delegates(account) == address(0)) votes += token.balanceOf(account);
```

**Purpose**

Indexed Finance currently uses `erc20-balance-of` so that users can participate in off-chain temperature checks even if they don't always vote on chain. We recently introduced a feature where users deposit NDX into a smart contract to become eligible for dividends, and automatically have their voting power delegated back to them. While most of the active users are depositing into this, we still want people holding undelegated NDX to be able to engage in community decision making. This strategy solves that issue, and I expect it might be useful to some other communities.


ChaituVR:
Strategy is live 
https://snapshot.org/#/strategy/comp-like-votes-inclusive

d1ll0n:
> Strategy is live https://snapshot.org/#/strategy/comp-like-votes-inclusive

Thanks!

PR #96, opened 2021-10-07
[hashes-voting] Add hashes-voting strategy

brennanfife:
Creates the hashes-voting strategy for Hashes NFT

### DIFF
--- a/src/strategies/index.ts
+++ b/src/strategies/index.ts
@@ -158,6 +158,7 @@ import * as lootCharacterGuilds from './loot-character-guilds';
 import * as cyberkongz from './cyberkongz';
 import * as compLikeVotesInclusive from './comp-like-votes-inclusive';
 import * as mstable from './mstable';
+import * as hashesVoting from './hashes-voting';
 
 const strategies = {
   coordinape,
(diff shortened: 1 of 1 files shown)


src/strategies/hashes-voting/index.ts
```diff
@@ -0,0 +1,55 @@
+import { formatUnits } from '@ethersproject/units';
+import { multicall } from '../../utils';
+
+export const author = 'Arr00';
```
bonustrack, line 4: This should be your username
```suggestion
export const author = 'brennanfife';
```

src/strategies/hashes-voting/index.ts
```diff
@@ -0,0 +1,55 @@
...
+      {
+        internalType: 'uint256',
+        name: '',
+        type: 'uint256'
+      }
+    ],
+    stateMutability: 'view',
+    type: 'function'
+  }
+];
```
bonustrack, line 27: Can you change to this:
```suggestion
const abi = ['function getCurrentVotes(address _account) view returns (uint256)'];
```

test/addresses.json
```diff
@@ -3348,5 +3348,6 @@
   "0xCc4d08068f5F93F0A48942f72b841c43DaEA3766",
   "0x3dDEA4021C4956a5347B01C05BE72cCa2dfE5653",
   "0x6271aDd46DCbb4B4EF5b32F18704bDd716BD9be9",
-  "0xF12504E1d4917A2eBa3c27794186be17C0f02033"
+  "0xF12504E1d4917A2eBa3c27794186be17C0f02033",
```
bonustrack, line 3351: You shouldn't change this file

PR #99, opened 2021-10-11
[pod-leader] Add pod leader strategy

UrsaMaritimus:
- Add strategy that works with yield yak like farming on avalanche, like the pod leader

### DIFF
--- a/src/strategies/pod-leader/README.md
+++ b/src/strategies/pod-leader/README.md
@@ -1,5 +1,7 @@
 # PodLeader pool balance
 
+Calculates the balance of either a uniswap pair in a Yield Yak/Pod leader style farm, or the amount of each individual token in that LP which is deposited into the farm. For example, if an LP pair has 100 token1 and 200 token2, one can isolate token2 for snapshot votes. One can also give weights to each of these tokens when combined with other strategies.
+
 ## Accepted options:
 
 - chefAddress: Masterchef contract address


src/strategies/pod-leader/index.ts
```diff
@@ -0,0 +1,518 @@
...
+
+  const userInfoChunks = arrayChunk(userInfoResponse, 1);
+
+  const tokenResponse = await multicall(network, provider, abi, tokenCalls, {
+    blockTag
+  });
+
+  for (let i = 0; i < userInfoChunks.length; i++) {
+    const value = userInfoChunks[i];
+    const score = await processValues(
```
ChaituVR, line 506: Here we are calling `processValues` function for every address, which has multiple fetch calls (ex: getPrice), and it is not scalable in the case of a large number of voters, is there an alternative you can think of where you can get data from a single call?

 Here is the checklist to add any strategy:

ChaituVR (changes requested):

src/strategies/pod-leader/index.ts
UrsaMaritimus: I have removed the getPrice to simplify it, just based off token weighting now. Should be able to scale properly now.

bonustrack (approved):

ChaituVR (approved):

ChaituVR:
Strategy is live: https://snapshot.org/#/strategy/pod-leader

ChaituVR:
Hi @UrsaMaritimus We noticed this strategy is causing memory issues on server and breaks everything, we will disable this soon, please feel free to contact us on https://discord.snapshot.org

PR #114, opened 2021-10-18
[with-balance] Add alias for strategies and change erc20-with-balance to with-balance

### LINKED ISSUE #107: Find a better way to rename strategies ( Not alias )
Create an alias for strategies where one strategy can have different names, 

Example:
```
erc20-with-balance: withBalance
with-balance: withBalance
```

both erc20-with-balance and with-balance should use the with-balance strategy

ChaituVR:
Fixes #107


ChaituVR:
Closing this for now (as per discussion in discord -  we need to find a different way to rename strategies)

PR #119, opened 2021-10-21
[erc20-balance-from-graph] Retrieve balance from graph

yangzhao28:
- add a strategy to retrieve balances from graph, to support the network with no available archive node.

### DIFF
--- a/src/strategies/erc20-balance-from-graph/README.md
+++ b/src/strategies/erc20-balance-from-graph/README.md
@@ -1,0 +1,21 @@
+This is simple strategy to use balance from graph instead of archive node.
+
+The configuration would be like this:
+
+```json
+{
+    "graph": "the api address of your graph",
+    "decimals": 18 
+}
+```
+
+The schema of the graph project is:
+
+```
+Account {
+    id: ID!
+    balance: BigDecimal!
+}
+```
+
+And the `decimals` is an optional field. Ignore it if you have already handle it in graph, or set it to the exact decimals of the raw token balance.
--- a/src/strategies/erc20-balance-from-graph/examples.json
+++ b/src/strategies/erc20-balance-from-graph/examples.json
@@ -1,0 +1,18 @@
+[
+  {
+    "name": "Example query",
+    "strategy": {
+      "name": "erc20-balance-from-graph",
+      "params": {
+        "graph": "https://api.thegraph.com/subgraphs/name/renpu-mcarlo/mcb-balance"
+      }
+    },
+    "network": "1",
+    "addresses": [
+      "0xfdf6c6096e70799e53d6172d32c6c3b6f3d1f38f",
+      "0xb796a0868a8fe1f93df3ea856daa6e169f7c37db",
+      "0x0cefe050ff97872c4390bdf28cd87822185ab266"
+    ],
+    "snapshot": 2393677
+  }
+]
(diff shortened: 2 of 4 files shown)


src/strategies/erc20-balance-from-graph/examples.json
```diff
@@ -0,0 +1,18 @@
+[
+  {
+    "name": "Example query",
+    "strategy": {
+      "name": "erc20-balance-from-graph",
```
ChaituVR, line 5: @yangzhao28 There is no option to pass custom ERC20 token right, this will get mcb-balance not all erc20-balances and can confuse other user
Would it be good if we change the name to `mcb-balance-from-graph` instead of erc20 ?

src/strategies/erc20-balance-from-graph/index.ts
```diff
@@ -0,0 +1,52 @@
...
+  options,
+  snapshot
+) {
+  const query = {
+    accounts: {
+      __args: {
+        where: {
+          id_in: addresses
+        },
+        first: 1000
```
ChaituVR, line 21: Can we make it like pagination to get balance all addresses instead of the first 1000
Something like 
```
const addressSubsets = Array.apply(
    null,
    Array(Math.ceil(addresses.length / LIMIT))
  ).map((_e, i) => addresses.slice(i * LIMIT, (i + 1) * LIMIT));
  const returnedFromSubgraph = await Promise.all(
    addressSubsets.map((subset) =>
      mcnSubgraphQuery(subset, snapshotBlock)
    )
  );
  ```

File: src/strategies/erc20-balance-from-graph/index.ts [hunk 3]
```diff
@@ -0,0 +1,52 @@
+import { subgraphRequest } from '../../utils';
+import { BigNumber } from '@ethersproject/bignumber';
+
+export const author = 'yangz';
```
ChaituVR, line 4:
```suggestion
export const author = 'yangzhao28';
```

src/strategies/erc20-balance-from-graph/README.md
```diff
@@ -0,0 +1,21 @@
+This is simple strategy to use balance from graph instead of archive node.
+
+The configuration would be like this:
+
+```json
+{
+    "graph": "the api address of your graph",
+    "decimals": 18
```
ChaituVR, line 8:
```suggestion
    "symbol": "MCB",
    "graph": "the api address of your graph"
```

src/strategies/erc20-balance-from-graph/examples.json
```diff
@@ -0,0 +1,18 @@
...
+      "params": {
+        "graph": "https://api.thegraph.com/subgraphs/name/renpu-mcarlo/mcb-balance"
```
ChaituVR, line 7:
```suggestion
        "symbol": "MCB",
        "graph": "https://api.thegraph.com/subgraphs/name/renpu-mcarlo/mcb-balance"
```

src/strategies/erc20-balance-from-graph/examples.json
yangzhao28: User need to setup a graph project for a token to use it. But thanks I'll rename the strategy.

ChaituVR: @yangzhao28 you can keep it `erc20` then :) I'm not aware of this graph so, or maybe add a comment on Readme how people can setup their graph or link to it would be helpful

PR #131, opened 2021-10-30
[mcb-balance-from-graph] Fix graph query failed caused by mixed-cases address

yangzhao28:
Fixes the graph query failed caused by mixed-cases addresses.

### DIFF
--- a/src/strategies/mcb-balance-from-graph/index.ts
+++ b/src/strategies/mcb-balance-from-graph/index.ts
@@ -34,10 +34,11 @@ export async function strategy(
   options,
   snapshot
 ) {
+  const _addresses = addresses.map(x => x.toLowerCase())
   const addressSubsets = Array.apply(
     null,
-    Array(Math.ceil(addresses.length / LIMIT))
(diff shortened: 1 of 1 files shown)


yangzhao28:
By the way, I found that the two requests sent by loading and verifing are exactly the same but the addresses in the response are lowercased and mixed-case respectively.

ChaituVR:
@yangzhao28 try to use below function:
```js
const checksumedAddress = getAddress(address);
```
using this you can convert an address into a checksummed address, we should not get lower cased addresses in the output

yangzhao28:
> @yangzhao28 try to use below function:
> 
> ```js
> import { getAddress } from '@ethersproject/address';
> const checksumedAddress = getAddress(address);
> ```
> 
> using this you can convert an address into a checksummed address, we should not get lower cased addresses in the output

Thanks. I didn't change the addresses passed by arguments. The lowercased addresses are only used to query graph and match the query result.  In this case, should I still use the `getAddress` ?

ChaituVR:
Oh then i think it is not required @yangzhao28 :)

bonustrack (approved):

PR #138, opened 2021-11-02
[galaxy-nft-with-score] add strategy: galaxy-nft-with-score

alberthaotan:
1. adding galaxy-nft-with-score strategy to list of strategies.

### DIFF
--- a/src/strategies/galaxy-nft-with-score/index.ts
+++ b/src/strategies/galaxy-nft-with-score/index.ts
@@ -1,9 +1,9 @@
 import fetch from 'cross-fetch'
 
-export const author = 'albert'
+export const author = 'alberthaotan'
 export const version = '0.1.0'
 
-const GALAXY_SUBGRAPH_URL = 'https://graphigo.prd.galaxy.eco/query'
+const GALAXY_GRAPHQL_URL = 'https://graphigo.prd.galaxy.eco/query'
 
 interface Config {
   name: string,
@@ -37,7 +37,7 @@ export async function strategy(
   options,
   snapshot
 ) {
-  let fetchRes = await fetch(GALAXY_SUBGRAPH_URL, {
+  let fetchRes = await fetch(GALAXY_GRAPHQL_URL, {
     method: 'POST',
     headers: {
       'Content-Type': 'application/json',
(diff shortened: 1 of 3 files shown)


src/strategies/galaxy-nft-with-score/index.ts
```diff
@@ -0,0 +1,98 @@
...
+
+export async function strategy(
+  space,
+  network,
+  provider,
+  addresses,
+  options,
+  snapshot
+) {
+  let fetchRes = await fetch(GALAXY_SUBGRAPH_URL, {
```
ChaituVR, line 40: Hello @alberthaotan It is better to use `subgraphRequest` here instead of `fetch`

Example 
https://github.com/snapshot-labs/snapshot-strategies/blob/2528c1bda2b025853bde6194e41004fe77bcafed/src/strategies/digitalax-mona-quickswap/index.ts#L31

src/strategies/galaxy-nft-with-score/index.ts
```diff
@@ -0,0 +1,98 @@
+import fetch from 'cross-fetch'
+
+export const author = 'albert'
```
ChaituVR, line 3:
```suggestion
export const author = 'alberthaotan'
```

src/strategies/galaxy-nft-with-score/index.ts
```diff
@@ -0,0 +1,98 @@
...
+  ownersWithNfts.forEach(ownerWithNfts => {
+    ownerWithNfts.nfts.forEach(nft => {
+      if (nft.name in ownerToNftCount[ownerWithNfts.owner]) {
+        ownerToNftCount[ownerWithNfts.owner][nft.name]++
+      } else {
+        ownerToNftCount[ownerWithNfts.owner][nft.name] = 1
+      }
+    })
+  })
+  console.log(ownerToNftCount)
```
ChaituVR, line 82:
```suggestion
```

ChaituVR (changes requested):

src/strategies/galaxy-nft-with-score/index.ts
alberthaotan: Sorry, this 'GALAXY_SUBGRAPH_URL' const var name is misleading, it's actually a graphql url, I changed the name to 'GALAXY_GRAPHQL_URL'. Thank you!

ChaituVR (approved):

ChaituVR:
@alberthaotan Hello, just now realized that this strategy doesn't take `snapshot` into consideration and voting power will keep changing in future

PR #139, opened 2021-11-02
Revert "[galaxy-nft-with-score] add strategy: galaxy-nft-with-score"

ChaituVR:
Reverts snapshot-labs/snapshot-strategies#138


alberthaotan:
Hi, Chaitu. Fixed the snapshot id is not taken into consideration issue in https://github.com/snapshot-labs/snapshot-strategies/pull/143, could you have another look? Thank you so much!

PR #142, opened 2021-11-03
[vesper] Add Vesper strategy

marcelomorgado:
This strategy calls the `VesperVotingPower` in order to get accounts' voting power.

### DIFF
--- a/src/strategies/index.ts
+++ b/src/strategies/index.ts
@@ -176,6 +176,7 @@ import * as digitalaxMonaQuickswap from './digitalax-mona-quickswap';
 import * as digitalaxGenesisContribution from './digitalax-genesis-contribution';
 import * as digitalaxLPStakers from './digitalax-lp-stakers';
 import * as galaxyNftWithScore from './galaxy-nft-with-score';
+import * as vesper from './vesper';
 
 const strategies = {
   coordinape,
@@ -353,7 +354,8 @@ const strategies = {
   'digitalax-genesis-contribution': digitalaxGenesisContribution,
   'digitalax-lp-stakers': digitalaxLPStakers,
   'colony-reputation': colonyReputation,
-  'galaxy-nft-with-score': galaxyNftWithScore
+  'galaxy-nft-with-score': galaxyNftWithScore,
+  vesper
 };
 
 Object.keys(strategies).forEach(function (strategyName) {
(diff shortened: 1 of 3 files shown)


src/strategies/index.ts
```diff
@@ -348,12 +349,14 @@ const strategies = {
   'mutant-cats-stakers-and-holders': mutantCatsStakersAndHolders,
   'svs-staking': svsStaking,
   'mcb-balance-from-graph': mcbBalanceFromGraph,
+  'colony-reputation': colonyReputation,
```
ChaituVR, line 352:
```suggestion
  'mcb-balance-from-graph': mcbBalanceFromGraph,
```

src/strategies/vesper/index.ts
```diff
@@ -0,0 +1,72 @@
+import { formatUnits } from '@ethersproject/units';
+import { getBlockNumber } from '../../utils';
+import { Multicaller } from '../../utils';
+
+export const author = 'Vesper Finance';
```
ChaituVR, line 5:
```suggestion
export const author = 'marcelomorgado';
```

src/strategies/vesper/index.ts
```diff
@@ -0,0 +1,72 @@
...
+      break;
+    }
+    snapshotBlocks.push(blockTag);
+  }
+
+  const scores = await Promise.all([
+    ...snapshotBlocks.map((blockTag) =>
+      getScores(provider, addresses, options, blockTag)
+    )
+  ]);
```
ChaituVR, line 34: @marcelomorgado I think we should have a limit to `periods` param else if someone passes thousands of previous blocks, strategy won't scale 
ideally, it should be like `5` maximum blocks maybe?


The limit of the number of requests/multicall is `5` 

ChaituVR (changes requested):

src/strategies/vesper/index.ts
marcelomorgado: Done. I've hardcoded `period = 4`

src/strategies/vesper/index.ts
marcelomorgado: Done.

src/strategies/index.ts
marcelomorgado: Done.

bonustrack (approved):

ChaituVR (approved):

PR #155, opened 2021-11-10
[multichain] add optional start block for each network

TomAFrench:
This PR adds an optional startBlocks field to the multichain strategy. This avoids a space's historical vote counts changing in the situation where they add support for a new network.


bonustrack (approved):
Looks good!

PR #163, opened 2021-11-14
Error description to find out which strategy is throwing the error

ChaituVR:
Right now it is hard to find which strategy is throwing the error, adding this can help find the issue faster (knowing which strategy is causing it)

image


ChaituVR:
Closing this as stale PR

PR #165, opened 2021-11-15
[coinswap] snapshot block number validation added

PawelGozdz:
Fixes # .

- "latest" changed to "typeof snapshot === 'number' ? snapshot : 'latest'"
- 
-


PawelGozdz:
ChaituVR, shall we manually set a block number on each time we start a new voting?

ChaituVR:
@PawelGozdz Yep! your proposal selects the snapshot block number when you are creating a proposal, you can also change this block number by adding a `snapshot` param to proposal create page (maybe to select older blocks)
https://snapshot.org/#/yourspace.eth/create?snapshot=true

ChaituVR (approved):

PR #171, opened 2021-11-19
[multichain] Fix multichain strategy's BSC subgraph issue

ChaituVR:


bonustrack:
@ChaituVR isnt going to break others networks?

ChaituVR:
@bonustrack  It works fine when I tested with other network graphs

```
'137': 'https://api.thegraph.com/subgraphs/name/sameepsi/maticblocks',
  '42161':
    'https://api.thegraph.com/subgraphs/name/ianlapham/arbitrum-one-blocks'
```

But on Aribitrum there are multiple blocks at one timestamp, For example:
when i try with 
` orderBy: timestamp` i get block `3121051` and when I try `orderBy: number`, I get `3121061` could this create some problems in the future? 

If yes, we may need to find another subgraph for BSC,
https://thegraph.com/hosted-service/subgraph/realbitcoinauto/blocks,
This works fine, but we may need to change proposals already passing `graph` param to this strategy

PR #190, opened 2021-11-24
[math] Add math strategy

xJonathanLEI:
This PR adds a new strategy `math` that allows composition of different strategies, similar to `multichain`.

As an example use case, a project might want to use user ERC20 token balances as voting socre, but to also limit whales by applying a square root operation on the balances. Instead of submitting yet another `project-x-strategy` to this repository, they can just pipe `erc20-balance-of` into `math` instead, keeping the strategy directory clean and reducing (mostly) duplicate code.


bonustrack (approved):
Thanks for your contribution! That's really cool that you've made it modular / usable for everyone.

PR #193, opened 2021-11-24
[blockzerolabs-cryptonauts] Blockzerolabs: Cryptonauts Strategy v2

umarsa:
Blockzerolabs: Cryptonauts Strategy v2


src/strategies/blockzerolabs-cryptonauts/index.ts
```diff
@@ -21,70 +11,13 @@ export async function strategy(
...
-          totalBalance = totalBalance.add(vaultBalance[_nftId]);
-        }
-      }
-    }
-    // Format the balance with 18 decimals (fixed)
-    balances[address] = parseFloat(formatUnits(totalBalance, options.decimals));
   });
-
-  return balances;
+  return await response.json();
```
ChaituVR, line 22: @umarsa we don't need a separate strategy for this right? we can call the `api` strategy directly right? if API can be modified
https://github.com/snapshot-labs/snapshot-strategies/blob/master/src/strategies/api/index.ts

ChaituVR (changes requested):

src/strategies/blockzerolabs-cryptonauts/index.ts
umarsa: Yeah, we may be able to use that pre-existing strategy.

Do you know what the "addresses" field is sending over to the endpoint?
https://github.com/snapshot-labs/snapshot-strategies/blob/master/src/strategies/api/index.ts#L18

My concern is that the URL might become unnecessarily long and potentially break the RFC guidelines..

ChaituVR: Hmm yea maybe i think you can try https://snapshot.org/#/strategy/api-post where you can send all addresses in body instead, let me know if it works :)

ChaituVR:
Closing this for https://github.com/snapshot-labs/snapshot-strategies/pull/193#discussion_r757075693 . feel free to reopen or create a new PR

PR #195, opened 2021-11-25
[agave] - Add strategy

maxaleks:


src/strategies/agave/index.ts
```diff
@@ -0,0 +1,62 @@
...
+  options,
+  snapshot
+) {
+  const blockTag = typeof snapshot === 'number' ? snapshot : 'latest';
+  const [
+    availableLiquidity,
+    lpTokenTotalSupply,
+    lpTokenDecimals,
+    underlyingTokenDecimals
+  ] = await Promise.all([
```
bonustrack, line 28: You could do a multicall here instead of sending 4 requests, that would resolve faster, here is an example https://github.com/snapshot-labs/snapshot.js/blob/master/test/multicaller.ts

src/strategies/multichain/examples.json
```diff
@@ -4,58 +4,112 @@
     "strategy": {
       "name": "multichain",
       "params": {
-        "symbol": "MULTI",
+        "symbol": "STAKE",
```
bonustrack, line 7: Looks good, can you just remove change on multichain examples?

maxaleks: Sorry, didn't notice 😅

bonustrack (approved):

PR #197, opened 2021-11-25
[honeyswap] - Optionally use staked LP tokens

maxaleks:
This improved strategy also uses the balances of the LP tokens sent to the HoneyFarm contract

### DIFF
--- a/src/strategies/honeyswap/index.ts
+++ b/src/strategies/honeyswap/index.ts
@@ -1,12 +1,128 @@
 import { getAddress } from '@ethersproject/address';
+import { formatUnits } from '@ethersproject/units';
 import { subgraphRequest } from '../../utils';
 
 export const author = 'Kittyslasher';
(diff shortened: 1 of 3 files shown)


src/strategies/honeyswap/index.ts
```diff
@@ -1,13 +1,129 @@
...
+}
+
+async function getStakedBalances(network, snapshot, token, addresses) {
+  const pools = await getPools(network, snapshot, token);
+  const params = {
+    deposits: {
+      __args: {
+        where: {
+          pool_in: Object.keys(pools),
+          user_in: addresses.map((addr) => addr.toLowerCase())
```
bonustrack, line 95: This doesn't scale well, the array on arguments have limit, I don't remember how many but call will fail at some point, maybe after 1000 votes. You may need to find another way to gather this information if you expect more vote than this

bonustrack (approved):

PR #202, opened 2021-11-30
[snet-liquidity-providers] strategy: snet-liquidity-providers

Vivek205:
This strategy returns the liquidity pool balances of the voters for the SingularityNET project.

### DIFF
--- a/src/strategies/snet-liquidity-providers/index.ts
+++ b/src/strategies/snet-liquidity-providers/index.ts
@@ -2,7 +2,7 @@ import { BigNumber, BigNumberish } from '@ethersproject/bignumber';
 import { formatUnits } from '@ethersproject/units';
 import { Multicaller, call } from '../../utils';
 
-export const author = 'bonustrack';
+export const author = 'Vivek205';
 export const version = '0.1.0';
 
 type FinalResult = [
--- a/src/strategies/snet-liquidity-providers/index.ts
+++ b/src/strategies/snet-liquidity-providers/index.ts
@@ -3,7 +3,7 @@ import { formatUnits } from '@ethersproject/units';
 import { Multicaller, call } from '../../utils';
 
 export const author = 'bonustrack';
-export const version = '0.1.1';
+export const version = '0.1.0';
 
 type FinalResult = [
   Record<string, BigNumberish>,
--- a/src/strategies/snet-liquidity-providers/index.ts
+++ b/src/strategies/snet-liquidity-providers/index.ts
@@ -41,7 +41,7 @@ export async function strategy(
   options,
   snapshot
 ): Promise<Record<string, number>> {
-  const blockTag = 'latest';
+  const blockTag = typeof snapshot === 'number' ? snapshot : 'latest';
   const { address: tokenAddress, lpAddress } = options;
 
   const initMultiCaller = multiCallerFactory(network, provider, blockTag);
--- a/src/strategies/snet-liquidity-providers/index.ts
+++ b/src/strategies/snet-liquidity-providers/index.ts
@@ -1,0 +1,76 @@
+import { BigNumber, BigNumberish } from '@ethersproject/bignumber';
+import { formatUnits } from '@ethersproject/units';
+import { Multicaller, call } from '../../utils';
+
+export const author = 'bonustrack';
+export const version = '0.1.1';
+
+type FinalResult = [
+  Record<string, BigNumberish>,
+  Record<string, BigNumberish>,
+  BigNumberish
+];
+
+const erc20Abi = [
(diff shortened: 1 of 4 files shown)


src/strategies/snet-liquidity-providers/index.ts
```diff
@@ -0,0 +1,76 @@
+import { BigNumber, BigNumberish } from '@ethersproject/bignumber';
+import { formatUnits } from '@ethersproject/units';
+import { Multicaller, call } from '../../utils';
+
+export const author = 'bonustrack';
```
ChaituVR, line 5:
```suggestion
export const author = 'Vivek205';
```

src/strategies/snet-liquidity-providers/index.ts
```diff
@@ -0,0 +1,76 @@
...
+
+export async function strategy(
+  space,
+  network,
+  provider,
+  addresses,
+  options,
+  snapshot
+): Promise<Record<string, number>> {
+  const blockTag = 'latest';
```
ChaituVR, line 44: @Vivek205 block can't be latest always - it may affect the voting power in future and change the results of closed proposals, It should be something like this:
```
```

ChaituVR (changes requested):

src/strategies/snet-liquidity-providers/index.ts
```diff
@@ -0,0 +1,76 @@
+import { BigNumber, BigNumberish } from '@ethersproject/bignumber';
+import { formatUnits } from '@ethersproject/units';
+import { Multicaller, call } from '../../utils';
+
+export const author = 'bonustrack';
+export const version = '0.1.1';
```
ChaituVR, line 6:
```suggestion
export const version = '0.1.0';
```

ChaituVR (changes requested):

src/strategies/snet-liquidity-providers/index.ts
Vivek205: fixed in https://github.com/snapshot-labs/snapshot-strategies/pull/202/commits/7bb1374abb5c34f45df2a839890fafac26c14253

src/strategies/snet-liquidity-providers/index.ts
Vivek205: fixed in https://github.com/snapshot-labs/snapshot-strategies/pull/202/commits/82d0311e9678f0aa1db49d9b29a5d7f7e71c5403

src/strategies/snet-liquidity-providers/index.ts
Vivek205: Hi @ChaituVR , this is expected behaviour, we would like to have it updated for every new block.

Vivek205:
@ChaituVR 
Fixed two comments and provided a description for the other. Please take a look.

src/strategies/snet-liquidity-providers/index.ts
ChaituVR: @Vivek205 hmm but this is not the way strategies works, we need a way to get consistent results once the proposal is closed else if voting power keeps changing in the future it could break things ( also it could be hard for you to make decisions based on something that is changing in future - ex: if I vote with 1 VP and in future, I may buy 100 VP and change the results)

For your requirement, the good solution (for now) would be to use snapshot input in the proposal creation page
https://snapshot.org/#/snet.eth/create?snapshot=1
image

You can give a snapshot block that is expected to be completed on proposal closure would give you correct voting power

Vivek205:
@ChaituVR 
Updated the blocknumber change. 
Also in 
#203 
and 
#204

src/strategies/snet-liquidity-providers/index.ts
Vivek205: fixed in https://github.com/snapshot-labs/snapshot-strategies/pull/202/commits/0025f5fe0aad6ead0eae45ca1fc52914a094bc35

Vivek205:
@ChaituVR 

We have custom staking contract and we decide the users' voting power by calling a get method in the contract. 
How will snapshot.js will retrieve the past data based on the blocktag for this scenario?

Are you storing the data for the users anywhere in a DB?


reg:
> ex: if I vote with 1 VP and in future, I may buy 100 VP and change the results)

if the user has already voted with 1VP, won't it be persisted in the DB and won't the user loses his ability to vote for the proposal thereafter?
Even if the user buys more tokens after voting, it should not impact the proposal because he has already cast his vote.

Vivek205:
@ChaituVR 
The checks have been merged. Could you get this merged?

bonustrack (approved):

bonustrack (approved):

ChaituVR (approved):

PR #211, opened 2021-12-02
[biswap] UPDATE: Stategy: biswap

biswap-dex:
Fixes #Error in LP balance .

- Add total supply and balance for lp address
- Make user balance in lp token correct


biswapcom:
Hello, then you can marge it? 
Our product team want's know then we can make proposal, thanks.

biswap-dex:
Hello, then you can marge it? 
Our product team want's know then we can make proposal, thanks.

bonustrack (approved):

ChaituVR:
@biswap-dex After these changes, this strategy is causing timeouts on the server, which is making the entire snapshot slow, some logic/ some call is making it hard to compute. would be great if you can fix it and raise a PR, 

So for now we blocked this strategy, will be unblocked once the issue is fixed

PR #213, opened 2021-12-02
[api] Update index.ts

MoonVault:
added new optional parameter "additionalParameters" THis can be used for APIs that require paremeters.
Example: api.com/api.php?module=governance&network=56&......
The optional parameter in this case is additionalParameters: "module=governance"

### DIFF
--- a/src/strategies/api/index.ts
+++ b/src/strategies/api/index.ts
@@ -13,13 +13,10 @@ export async function strategy(
   snapshot
 ) {
   let api_url = options.api + '/' + options.strategy;
-  api_url += '?';
-  if (options.additionalParameters) {
-    api_url += options.additionalParameters + '&'
-  }
-  api_url += 'network=' + network;
+  api_url += '?network=' + network;
   api_url += '&snapshot=' + snapshot;
   api_url += '&addresses=' + addresses.join(',');
+  if (options.additionalParameters) api_url += '&' + options.additionalParameters;
 
   const response = await fetch(api_url, {
     method: 'GET',
--- a/src/strategies/api/index.ts
+++ b/src/strategies/api/index.ts
@@ -20,7 +20,6 @@ export async function strategy(
   api_url += 'network=' + network;
   api_url += '&snapshot=' + snapshot;
   api_url += '&addresses=' + addresses.join(',');
-  api_url += options.additionalParameters;
 
   const response = await fetch(api_url, {
     method: 'GET',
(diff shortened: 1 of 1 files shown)


src/strategies/api/index.ts
```diff
@@ -13,9 +13,14 @@ export async function strategy(
...
   let api_url = options.api + '/' + options.strategy;
-  api_url += '?network=' + network;
+  api_url += '?';
+  if (options.additionalParameters) {
+    api_url += options.additionalParameters + '&'
+  }
+  api_url += 'network=' + network;
   api_url += '&snapshot=' + snapshot;
   api_url += '&addresses=' + addresses.join(',');
+  api_url += options.additionalParameters;
```
ChaituVR, line 23: Why `additionalParameters` are adding up to api_url twice? 🤔

ChaituVR (changes requested):

src/strategies/api/index.ts
MoonVault: oh sorry line #23 had to be removed. at start it just added it in the end, but for better formet URL i thought to put it infornt instead

MoonVault:
how can i edit it again? havent used pull requests often until now

src/strategies/api/index.ts
```diff
@@ -13,7 +13,11 @@ export async function strategy(
...
 ) {
   let api_url = options.api + '/' + options.strategy;
-  api_url += '?network=' + network;
+  api_url += '?';
+  if (options.additionalParameters) {
+    api_url += options.additionalParameters + '&'
+  }
+  api_url += 'network=' + network;
   api_url += '&snapshot=' + snapshot;
   api_url += '&addresses=' + addresses.join(',');
```
bonustrack, line 22: I would rather do that way to keep order the same everywhere (first & then param)
```suggestion
  api_url += '?network=' + network;
  api_url += '&snapshot=' + snapshot;
  api_url += '&addresses=' + addresses.join(',');
  if (options.additionalParameters) api_url += '&' + options.additionalParameters;
```

MoonVault: in general it doesnt really matter.
as many APIs require stuff like ?module=example&action=whatever  i just wanted to keep the format order

i guess i also have to change the old test script to use another interface? i guess the test failed because the example used isnt online anymore

bonustrack (approved):

ChaituVR (approved):

PR #214, opened 2021-12-03
[saddle-finance] Saddle Finance Strategy

ug02fast:
Strategy for Saddle Finance added. https://github.com/saddle-finance/tickets/issues/213


src/strategies/saddle-finance/index.ts
```diff
@@ -0,0 +1,126 @@
...
+      'vestings',
+      [retroAddr]
+    ]),
+    { blockTag }
+  );
+
+  const retroUserBalances = {};
+  retroAddrs.forEach((addr, i) => {
+    const userVesting = userVestingsRes[i];
+    if (userVesting.isVerified) {
```
hammeiam, line 76: I think this should always be defined, but it may be worth doing a `?.isVerified` just in case

src/strategies/saddle-finance/index.ts
```diff
@@ -0,0 +1,126 @@
...
+      retroUserBalances[addr] = parseFloat(
```
hammeiam, line 77: Welcome to the joys of BigNumber

src/strategies/saddle-finance/index.ts
```diff
@@ -0,0 +1,126 @@
...
+  }
+  for (const [addr, amount] of Object.entries(
+    mappedBeneficiariesToVestingContract
+  )) {
+    if (userTotal[addr]) userTotal[addr] += amount;
+    else userTotal[addr] = amount;
+  }
+
+  const finalUserBalances = Object.fromEntries(
+    Object.entries(userTotal).filter(([addr]) => addresses.includes(addr))
```
hammeiam, line 122: May be worth making `addresses` into a `Set` for faster lookups.

hammeiam: Actually I guess it's faster just to do `addresses.map(addr => [addr, userTotal[addr] || 0])`

hammeiam (approved):
Logic looks good to me! I think it'd be good to include test cases for each of the 3 sources (wallet, retro, vesting), and if possible, find one or two addresses that overlap sources as well.

src/strategies/index.ts
```diff
@@ -405,7 +406,8 @@ const strategies = {
   juicebox,
   'snet-farmers': snetFarmers,
   'snet-stakers': snetStakers,
-  "snet-liquidity-providers": snetLiquidityProviders
+  'snet-liquidity-providers': snetLiquidityProviders,
```
ChaituVR, line 409: where is `snet-liquidity-providers` strategy? would be great if you can open a seperate PR for that :)

ug02fast: sir I just added our strategy to the bottom of the list :)

src/strategies/saddle-finance/index.ts
ug02fast: It's a lot more readable that's for sure.

ChaituVR:
Looks good, would be great if you can resolve conflicts :)

src/strategies/saddle-finance/index.ts
```diff
@@ -0,0 +1,136 @@
+import { formatUnits } from '@ethersproject/units';
+import { multicall } from '../../utils';
+import remappedMerkleData from './remappedMerkleData.json';
```
bonustrack, line 3: Could you load this file from IPFS directly? We would prefer not having large file within the library itself

ug02fast: @bonustrack @ChaituVR done ✔️

ChaituVR:
@ug02fast It is taking more time to resolve all the calls, is there some way to fetch data in fewer calls?

ug02fast:
@ChaituVR If this isn't sufficient then I will probably have to utilize a subgraph. Do you have any recommendations?

src/strategies/saddle-finance/index.ts
```diff
@@ -0,0 +1,148 @@
...
+    network,
+    provider,
+    abi,
+    addresses.map((address: any) => [
+      SDLTokenAddress,
+      'balanceOf',
+      [address.toLowerCase()]
+    ]),
+    { blockTag }
+  );
```
ChaituVR, line 50: Can you try moving all 4 multicall promises into one `promise.all` so all multicalls will happen at the same time and could save us some time  ( as they are not dependent on each other)

make sure to run this in your local
```bash
npm test --strategy=saddle-finance
```

ug02fast: Thanks, seems to be consistently running @ under 10s locally now, hopefully, the same holds on CI.

bonustrack (approved):

PR #219, opened 2021-12-06
[minmax-mcn-farm] Add custom strategy for minmax finance

LeifuChen:
This is a custom strategry for minmax finance.
This strategy returns the total amount of user-owned MAX tokens as the voting power, including MAX in wallets, pending rewards and MAX LP held by the [farm contract](https://farm.mcn.ventures/farms) on iotex.


ChaituVR (changes requested):
Hello @LeifuChen We need to import the strategy into this file https://github.com/snapshot-labs/snapshot-strategies/blob/master/src/strategies/index.ts

LeifuChen:
@ChaituVR Thanks for your remind. I forgot adding it into the commits after local testing. Is it ok now?

src/strategies/index.ts
```diff
@@ -405,7 +406,8 @@ const strategies = {
   juicebox,
   'snet-farmers': snetFarmers,
   'snet-stakers': snetStakers,
-  'snet-liquidity-providers': snetLiquidityProviders
+  'snet-liquidity-providers': snetLiquidityProviders，
```
ChaituVR, line 409:
```suggestion
  'snet-liquidity-providers': snetLiquidityProviders,
```

ChaituVR (changes requested):

bonustrack (approved):

ChaituVR (approved):

PR #221, opened 2021-12-08
[math] Add min and max operations to `math` strategy

xJonathanLEI:
This PR:

- changes `math` params structure to support multi-operand operations, while maintaining backward compatibility by automatically converting the old format to the new one; and
- adds two new operations: `min` and `max`; and
- bumps `math` version to `0.2.0`.

Operations like `min` and `max` are useful in cases where DAOs want to limit any single account's voting power, or to set a minimum threshold. All they have to do is to nest their original strategies into `math` as a `strategy` operand, and apply a `min`/`max` operation against a constant operand.

The original example is left untouched to demonstrate backward compatibility, and 2 new examples have been added to demonstrate the new `options` format and `math` strategy nesting.

### DIFF
--- a/src/strategies/math/README.md
+++ b/src/strategies/math/README.md
@@ -2,12 +2,18 @@
 
 Apply common mathematical operations on outputs from other strategies.
 
+## Operations
+
 Currently supported operations are:
 
-- `square-root`
-- `cube-root`
-- `min`
-- `max`
+| Operation     | Operand Count | Description                                |
+| ------------- | ------------- | ------------------------------------------ |
+| `square-root` | 1             | takes the square root of the operand       |
+| `cube-root`   | 1             | takes the cube root of the operand         |
+| `min`         | 2             | takes the smaller number of the 2 operands |
+| `max`         | 2             | takes the larger number of the 2 operands  |
+
+## Examples
 
 The following example takes the cube root of a user's DAI token balance as voting score.
 
--- a/src/strategies/math/README.md
+++ b/src/strategies/math/README.md
@@ -6,21 +6,68 @@ Currently supported operations are:
 
 - `square-root`
 - `cube-root`
+- `min`
+- `max`
 
 The following example takes the cube root of a user's DAI token balance as voting score.
 
 ```json
 {
   "symbol": "MATH",
-  "strategy": {
-    "name": "erc20-balance-of",
-    "network": "1",
-    "params": {
-      "address": "0x6b175474e89094c44da98b954eedeac495271d0f",
-      "symbol": "DAI",
-      "decimals": 18
+  "operands": [
+    {
+      "type": "strategy",
+      "strategy": {
+        "name": "erc20-balance-of",
+        "network": "1",
+        "params": {
+          "address": "0x6b175474e89094c44da98b954eedeac495271d0f",
+          "symbol": "DAI",
+          "decimals": 18
+        }
+      }
     }
-  },
+  ],
   "operation": "square-root"
+}
+```
+
+Here's another example that sets a maximum score of 100 for the result above.
+
+```json
+{
+  "symbol": "MATH",
+  "operands": [
+    {
+      "type": "strategy",
+      "strategy": {
+        "name": "math",
+        "network": "1",
+        "symbol": "MATH",
+        "params": {
+          "operands": [
+            {
+              "type": "strategy",
+              "strategy": {
+                "name": "erc20-balance-of",
+                "network": "1",
+                "params": {
+                  "address": "0x6b175474e89094c44da98b954eedeac495271d0f",
+                  "symbol": "DAI",
+                  "decimals": 18
+                }
+              }
+            }
+          ],
+          "operation": "square-root"
+        }
+      }
+    },
+    {
+      "type": "constant",
+      "value": 100
+    }
+  ],
+  "operation": "min"
 }
 ```
--- a/src/strategies/math/index.ts
+++ b/src/strategies/math/index.ts
@@ -1,48 +1,20 @@
 import { getProvider } from '../../utils';
 import strategies from '..';
 
+import {
+  ConstantOperand,
+  migrateLegacyOptions,
+  Operand,
+  OperandType,
+  Operation,
+  OptionalOptions,
+  Options,
+  StrategyOperand,
+  validateOptions
+} from './options';
+
 export const author = 'xJonathanLEI';
 export const version = '0.2.0';
-
-interface Options {
-  operands: Operand[] | undefined;
-  operation: Operation | undefined;
-}
-
-interface LegacyFields {
-  strategy: any; // Legacy option used in v0.1.0
-}
-
-type Operand = StrategyOperand | ConstantOperand;
-
-enum OperandType {
-  Strategy = 'strategy',
-  Constant = 'constant'
-}
-
-interface StrategyOperand {
-  type: OperandType.Strategy;
-  strategy: any;
-}
-
-interface ConstantOperand {
-  type: OperandType.Constant;
-  value: number;
-}
-
-enum Operation {
-  SquareRoot = 'square-root',
-  CubeRoot = 'cube-root',
-  Min = 'min',
-  Max = 'max'
-}
-
-const operandCountByOperation: Record<Operation, number> = {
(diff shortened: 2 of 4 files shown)


xJonathanLEI:
The test CI has failed with a time-out error. This doesn't seem to be related with the PR itself. Possible to re-run CI?

xJonathanLEI:
Hello guys any update on this? Can we re-run the test CI? This PR changes nothing about the actual requests made, so that error should totally be unrelated to the changes.

src/strategies/math/README.md
```diff
@@ -6,21 +6,68 @@ Currently supported operations are:
 
 - `square-root`
 - `cube-root`
+- `min`
```
ChaituVR, line 9: Can you add documentation on how these min-max work? for square root and cube root it is pretty straightforward

src/strategies/math/README.md
```diff
@@ -6,21 +6,68 @@ Currently supported operations are:
...
+        "network": "1",
+        "params": {
+          "address": "0x6b175474e89094c44da98b954eedeac495271d0f",
+          "symbol": "DAI",
+          "decimals": 18
+        }
+      }
     }
-  },
+  ],
```
ChaituVR, line 30: There should be a limit on how many operands can be passed in, else it would add additional load to the server

src/strategies/math/index.ts
```diff
@@ -12,36 +52,155 @@ export async function strategy(
...
-  const upstreamResult: Record<string, number> = await strategies[
-    options.strategy.name
-  ].strategy(
-    space,
-    options.strategy.network,
-    getProvider(options.strategy.network),
-    addresses,
-    options.strategy.params,
-    snapshot
+  const strategyOptions: Options = migrateLegacyOptions(options);
```
ChaituVR, line 55: I feel it is better to create a new strategy than change the existing options, it will confuse a lot of users

src/strategies/math/index.ts
```diff
@@ -12,36 +52,155 @@ export async function strategy(
...
+  validateOptions(strategyOptions);
+
+  // Recursively resolve operands
+  const operandPromises: Promise<
+    Record<string, number>
+  >[] = strategyOptions.operands!.map((item) =>
```
ChaituVR, line 61:
```
Forbidden non-null assertion
```

src/strategies/math/index.ts
```diff
@@ -12,36 +52,155 @@ export async function strategy(
...
+    resolveOperand(item, addresses, space, snapshot)
+  );
+  const resolvedOperands: Record<string, number>[] = await Promise.all(
+    operandPromises
+  );
+
+  const finalResult: Record<string, number> = resolveOperation(
+    strategyOptions.operation!,
```
ChaituVR, line 69:
```
Forbidden non-null assertion
```

ChaituVR (changes requested):

src/strategies/math/README.md
xJonathanLEI: Documentation improved by turning the list into a table.

src/strategies/math/README.md
xJonathanLEI: This _is_ a limit on operand count. The number of operands is limited based on the operation used. `square-root` and `cube-root` take 1 operand, while `min` and `max` take 2. Passing in more operands result in an error.

That said, this strategy is similar to `multichain` in the sense that limiting the number of operands (strategies) on the strategy level doesn't really help, as the caller can simply nest more strategies inside, forcing the server to do an unbounded number of requests. The only way to truly mitigate the issue is to set the number of requests made on the caller's side.

Also, `multichain` does not impose a limit on the number of strategies either.

src/strategies/math/index.ts
xJonathanLEI: Code refactored to remove non-null assertion by turning null-able fields into non-null-able ones after validation.

src/strategies/math/index.ts
xJonathanLEI: Code refactored to remove non-null assertion by turning null-able fields into non-null-able ones after validation.

src/strategies/math/index.ts
xJonathanLEI: I understand the concern. However, the `math` strategy was created for composibility, and would be greatly limited in terms of what can be built on top of it if multi-operand support cannot be added.

Also I just checked and there doesn't seem to be any space using this strategy yet:

```sh
$ curl "https://hub.snapshot.org/api/explore" | jq ".strategies.math"
null
```

xJonathanLEI:
Hi @ChaituVR, the changes requested have been resolved. Would you mind taking a look at the updated PR? Thanks.

The test CI seems to fail yet again. I believe the command used to run the CI is:

```sh
$ npm run test --strategy=math --more=500
```

which I don't have issues running on my side:

```
Test Suites: 1 passed, 1 total
Tests:       13 passed, 13 total
Snapshots:   0 total
Time:        5.162 s, estimated 6 s
```

xJonathanLEI:
On a side note, this strategy was created because we saw so many strategies created by projects for their own use, which sometimes are almost identical, and thought we can do better. Instead of having every project submit and keep updating their own strategy, and thus flooding the `snapshot-strategies` repository and costing a substantial amount of review effort, composibility can be introduced such that most projects can simply construct strategies by using existing ones.

The only strategy enabling composibility before `math` is `multichain`, which only allows summing up different operations. Composing strategies was thus impractical. By introducing and improving `math`, we hope to help make composing strategies using existing building blocks a more viable option.

xJonathanLEI:
@ChaituVR Sorry for being pushy but is there any update on this? Thanks.

src/strategies/math/README.md
```diff
@@ -2,25 +2,78 @@
...
 {
   "symbol": "MATH",
-  "strategy": {
-    "name": "erc20-balance-of",
-    "network": "1",
-    "params": {
-      "address": "0x6b175474e89094c44da98b954eedeac495271d0f",
-      "symbol": "DAI",
-      "decimals": 18
+  "operands": [
```
bonustrack, line 23: You shouldn't change the structure of the params here it may break results for space currently using this strategy

xJonathanLEI: I understand the concern. However, the `math` strategy was created for composibility, and would be greatly limited in terms of what can be built on top of it if multi-operand support cannot be added.

The change in structure is backward-compatible, as the strategy automatically converts the old param format into the new one.

Also there doesn't seem to be any space using this strategy yet:

```sh
$ curl "https://hub.snapshot.org/api/explore" | jq ".strategies.math" null
```

xJonathanLEI: @bonustrack Any update on this? Thank you so much.

src/strategies/math/index.ts
```diff
@@ -12,36 +24,104 @@ export async function strategy(
...
+  switch (operand.type) {
+    case OperandType.Strategy: {
+      const strategyOperand: StrategyOperand = operand as StrategyOperand;
+
+      const upstreamResult: Record<string, number> = await strategies[
+        strategyOperand.strategy.name
+      ].strategy(
+        space,
+        strategyOperand.strategy.network,
+        getProvider(strategyOperand.strategy.network),
```
bonustrack, line 111: We shouldnt be able to change the network, changing network would make "snapshot" value not correct, we should default network instead.

src/strategies/math/README.md
bonustrack: Thanks make sense, apologies for slow review

src/strategies/math/index.ts
xJonathanLEI: Hi @bonustrack, thanks for the feedback.

Yes this makes sense. If the user wants to use data from multiple chains they should simply nest a `mutlichain` strategy inside.

xJonathanLEI:
I'm not sure why but it seems that the staleness of this branch is causing all the CI failures. I'm rebasing this branch against the latest `master` to see if it helps.

src/strategies/math/index.ts
xJonathanLEI: @bonustrack Changed to use upstream `network` and `provider`.

bonustrack (approved):
Looks good!

ChaituVR (approved):

PR #224, opened 2021-12-09
[ocean-dao-brightid] Add Ocean DAO BrightID strategy

trizin:
Added a new strategy for OceanProtocol DAO that integrates BrightID with multiple strategies.


src/strategies/ocean-dao-brightid/index.ts
```diff
@@ -0,0 +1,57 @@
...
+    addresses.map((address: any) => [
+      options.registry,
+      'isVerifiedUser',
+      [address]
+    ]),
+    { blockTag: 'latest' }
+  );
+  let scores = await getScoresDirect(
+    space,
+    options.strategies,
```
ChaituVR, line 31: we should limit these strategies passed into this strategy. else this could increase the load on the server (may be one or two strategies should be good, @bonustrack any suggestions on the number? )

ChaituVR (changes requested):

src/strategies/ocean-dao-brightid/index.ts
ChaituVR: Yes @trizin we are looking to have a limit on those strategies too. but should find a way to make sure existing proposals work well :)

trizin: That's great. I wanted to point it out just in case you missed them. We planned to use this strategy as a master strategy with the five strategies in the `examples.json.` The strategies are lightweight and don't take much time to execute. Perhaps it is better to focus on the execution time rather than the number of the strategies.

ChaituVR: @trizin In that case it would be better to hard code these strategies inside the strategy than through the options because in the future some other teams may want to use this and may pass different complex strategies. Thanks for understanding :)

src/strategies/ocean-dao-brightid/index.ts
```diff
@@ -0,0 +1,71 @@
...
+  const response = await multicall(
+    options.brightIdNetwork,
+    getProvider(options.brightIdNetwork),
+    abi,
+    addresses.map((address: any) => [
+      options.registry,
+      'isVerifiedUser',
+      [address]
+    ]),
+    { blockTag: 'latest' }
```
bonustrack, line 39: This wont work, you need to use the snapshot block number otherwise the score may change anytime and proposals results would also change. The score should always be the same if you compute it now or next week unless the param "snapshot" is equal to "latest"

trizin: @bonustrack Thanks for pointing that out! I've changed it to `blockTag: snapshot`.

src/strategies/ocean-dao-brightid/index.ts
```diff
@@ -0,0 +1,71 @@
+import { getScoresDirect, multicall, getProvider } from '../../utils';
```
ChaituVR, line 1:
```suggestion
import { getScoresDirect, multicall } from '../../utils';
```

ChaituVR:
@trizin 
It is throwing this error. any idea?
```
call revert exception (method="isVerifiedUser(address)", errorArgs=null, errorName=null, errorSignature=null, reason=null, code=CALL_EXCEPTION, version=abi/5.4.0)

      at Logger.Object.<anonymous>.Logger.makeError (node_modules/@ethersproject/logger/src.ts/index.ts:213:28)
      at Logger.Object.<anonymous>.Logger.throwError (node_modules/@ethersproject/logger/src.ts/index.ts:225:20)
      at Interface.Object.<anonymous>.Interface.decodeFunctionResult (node_modules/@ethersproject/abi/src.ts/interface.ts:425:23)
      at node_modules/@snapshot-labs/snapshot.js/dist/snapshot.cjs.js:2308:40
          at Array.map (<anonymous>)
      at Object.<anonymous> (node_modules/@snapshot-labs/snapshot.js/dist/snapshot.cjs.js:2307:51)
      at step (node_modules/@snapshot-labs/snapshot.js/dist/snapshot.cjs.js:80:23)
      at Object.next (node_modules/@snapshot-labs/snapshot.js/dist/snapshot.cjs.js:61:53)
      at fulfilled (node_modules/@snapshot-labs/snapshot.js/dist/snapshot.cjs.js:51:58)
```

src/strategies/ocean-dao-brightid/index.ts
```diff
@@ -0,0 +1,71 @@
...
+  addresses,
+  options,
+  snapshot
+) {
+  const ALLOWED_STRATEGIES = [
+    'erc20-balance-of',
+    'ocean-marketplace',
+    'sushiswap',
+    'uniswap',
+    'contract-call'
```
bonustrack, line 23: Any reason you enforce specifics strategies here?

bonustrack (approved):
Looks good

PR #226, opened 2021-12-09
Feat: jade-smrt strat [jade-smrt]

drgorillamd:
Strat for jade and smrt (multichain + LP)


src/strategies/jade-smrt/index.ts
```diff
@@ -0,0 +1,182 @@
...
+
+  const chainBlocks = await getChainBlocks(
+    snapshot,
+    provider,
+    options,
+    network
+  );
+
+  let block = await provider.getBlock(typeof snapshot === 'number' ? snapshot : 'latest');
+  let timestamp = block.timestamp;
```
ChaituVR, line 85: Would be nice if we can reuse this block in lines 46 and 47

src/strategies/jade-smrt/index.ts
```diff
@@ -0,0 +1,182 @@
...
+  let jadePrice: number = await geckoPrice(JADE, timestamp, 'binance-smart-chain');
+  let results: any = [];
+
+  for (const strategy of options.strategies) {
```
ChaituVR, line 89: we should limit these strategies passed into these. else this could increase the load on the server (may be one or two strategies should be good, @bonustrack any suggestions on the number? )

ChaituVR (changes requested):

src/strategies/jade-smrt/index.ts
drgorillamd: I took it from the multichain (the example has 5 strat iirc). I just send you a message on discord with my former function prototypes, it was rather rewriting erc20-balanceOf/uni-LP (and resolve everything at once) than building on top of multichain/erc20-balanceOf/etc and iterating over strats.
Let me know if you want me to roll back to that implementation, happy to do so!

drgorillamd:
refactored : not using the multichain with other strats anymore -> every async in same resolve all

src/strategies/jade-smrt/index.ts
```diff
@@ -0,0 +1,126 @@
...
+  });
+
+  // Avax SMRTR/WAVAX pool: SMRTR balance and LP token total supply
+  const LPBalance = await call(getProvider('43114'), abi, [options.SMRTR.address, 'balanceOf', [options.SMRTRLP.address]]);
+  const LPSupply =  await call(getProvider('43114'), abi, [options.SMRTRLP.address, 'totalSupply', []]);
+  promises.push(multiAvax.execute());
+
+  // eslint-disable-next-line
+  let jadePrice, smrtPrice, smrtRPrice, resBsc, resAvax;
+  [jadePrice, smrtPrice, smrtRPrice, resBsc, resAvax] = await Promise.all(promises);
```
ChaituVR, line 58:
```suggestion
  const [jadePrice, smrtPrice, smrtRPrice, resBsc, resAvax] = await Promise.all(promises);
```

ChaituVR (changes requested):

ChaituVR:
@drgorillamd 
image

drgorillamd:
> @drgorillamd image

fixed:)

src/strategies/jade-smrt/index.ts
```diff
@@ -0,0 +1,131 @@
...
+  const multiAvax = new Multicaller('43114', getProvider('43114'), abi, { blockTag: avaxBlockTag });
+  addresses.forEach((address: string) => {
+    multiAvax.call(address+"-smrt", options.SMRT.address, 'balanceOf', [address]);
+    multiAvax.call(address+"-smrtR", options.SMRTR.address, 'balanceOf', [address]);
+    multiAvax.call(address+"-smrtRLp", options.SMRTRLP.address, 'balanceOf', [address])
+  });
+
+  // Avax SMRTR/WAVAX pool: SMRTR balance and LP token total supply
+  const LPBalance = await call(getProvider('43114'), abi, [options.SMRTR.address, 'balanceOf', [options.SMRTRLP.address]]);
+  const LPSupply =  await call(getProvider('43114'), abi, [options.SMRTRLP.address, 'totalSupply', []]);
```
ChaituVR, line 45: why not using Multicaller here also like above? 🤔 than two separate calls?

ChaituVR (changes requested):
There is a limit of 5 calls/fetch/mutlicalls 

would be great if you can reduce the number of calls :)

src/strategies/jade-smrt/index.ts
drgorillamd: Indeed, mb, totally missed the 5 call max -> fixed it (same for multiple coingecko calls), thank you for the feedback:)

bonustrack (approved):
Any reason to not use multichain strategy?

drgorillamd:
> Any reason to not use multichain strategy?

@ChaituVR changeRequest (https://github.com/snapshot-labs/snapshot-strategies/pull/226#discussion_r766050053), it was a vanilla multichain strat at first (with 5 "sub-strategies")

ChaituVR (approved):

PR #231, opened 2021-12-10
[Decentral Games] Decentral Games Strategy

SatoshiNaoki:
Strategy for Decentral Games added

### DIFF
--- a/src/strategies/decentral-games-staking/examples.json
+++ b/src/strategies/decentral-games-staking/examples.json
@@ -1,0 +1,56 @@
+[
+  {
+    "name": "Decentral Games Multichain Voting",
+    "strategy": {
+      "name": "decentral-games-staking",
+      "params": {
+        "symbol": "VOTE",
+        "strategies": [
+          {
+            "name": "erc20-balance-of",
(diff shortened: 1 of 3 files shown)


ChaituVR (changes requested):
Hi @SatoshiNaoki This looks equal to `multichain` strategy. why not use it instead?

SatoshiNaoki:
> Hi @SatoshiNaoki This looks equal to `multichain` strategy. why not use it instead?

for some contracts, voting power should be x1000.
I was going to do it by setting decimals as 15 instead of 18, but it seems it's not good.
so I just updated the calculation

ChaituVR:
Hi @SatoshiNaoki 

> for some contracts, voting power should be x1000.
I was going to do it by setting decimals as 15 instead of 18, but it seems it's not good.
so I just updated the calculation

Which line is this change? I see it is the same as multichain? 🤔 
Do you mean in examples.json? if so you can change the params from your space settings directly

ChaituVR:
Closing this as multichain strategy works for you. feel free to reopen if needed

PR #251, opened 2021-12-19
[cronaswap] Add strategy for CronaSwap

CronaChef:
[cronaswap] Add strategy for CronaSwap

### DIFF
--- a/src/strategies/cronaswap/README.md
+++ b/src/strategies/cronaswap/README.md
@@ -1,0 +1,45 @@
+# CronaSwap
+
+This is the most common strategy, it returns the balances of the voters for a balances CRONA token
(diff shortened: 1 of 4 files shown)


ChaituVR (changes requested):
Hi @CronaChef Looks good but the strategy should be imported into this file https://github.com/snapshot-labs/snapshot-strategies/blob/master/src/strategies/index.ts

CronaChef:
> Hi @CronaChef Looks good but the strategy should be imported into this file https://github.com/snapshot-labs/snapshot-strategies/blob/master/src/strategies/index.ts

OK, thanks, added. pls check.

bonustrack (approved):

ChaituVR (approved):

PR #254, opened 2021-12-20
[orange-reputation-based-voting] add orange-reputation-based-voting

joeuan:
add orange-reputation-based-voting for orange-protocol

### DIFF
--- a/src/strategies/orange-reputation-based-voting/index.ts
+++ b/src/strategies/orange-reputation-based-voting/index.ts
@@ -1,0 +1,39 @@
+import fetch from 'cross-fetch';
+
+export const author = 'orange-protocol';
+export const version = '0.1.0';
+
+export async function strategy(
+  space,
+  network,
+  provider,
+  addresses,
+  options,
+  snapshot
+): Promise<Record<string, number>> {
+  const query = `{
+    getBasedVotingStrategy(addrs: ${JSON.stringify(
+      addresses
+    )}, space: "${space}")
+    { address score }
+  }`;
+  const data = {
+    operationName: '',
+    query,
+    variables: {}
+  };
+  const rawResponse = await fetch(
+    'https://api.orangeprotocol.io/orange2c/query',
+    {
+      method: 'POST',
+      headers: {
+        Accept: 'application/json',
+        'Content-Type': 'application/json'
+      },
+      body: JSON.stringify(data)
+    }
+  );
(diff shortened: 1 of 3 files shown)


src/strategies/orange-reputation-based-voting/index.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export const author = 'orange-protocol';
+export const version = '0.1.0';
+
+export async function strategy(
+  space,
+  network,
+  provider,
+  addresses,
+  options,
```
ChaituVR, line 11: 'options' is defined but never used 🤔 but I see address and decimals being passed from examples.json?

src/strategies/orange-reputation-based-voting/index.ts
```diff
@@ -0,0 +1,39 @@
...
+  snapshot
```
ChaituVR, line 12: We need to pass snapshot to API and API should make sure it returns data based on the blocknumber else the results would be changing in future

ChaituVR (changes requested):

src/strategies/orange-reputation-based-voting/index.ts
```diff
@@ -0,0 +1,44 @@
...
+): Promise<Record<string, number>> {
+  const query = `{
+    getBasedVotingStrategy(
+      addrs: ${JSON.stringify(addresses)},
+      space: "${space}",
+      snapshot: "${snapshot}",
+      network: "${network}",
+      provider: "${provider}",
```
bonustrack, line 20: Provider is not a string or number, I assume it wont work well here, you'd rather not having it at all

ChaituVR: @bonustrack This is input for the oraganprotocol API https://api.orangeprotocol.io/orange2c/query
I can't find their documentation or schema 🤔

bonustrack: But this return the variable from the strategy with Ethers.js provider https://github.com/snapshot-labs/snapshot-strategies/pull/254/files#diff-b60e2ce70275e1efd45ba615c092306ebd2de4585a540f539a16f1ba91ed4ed0R9

bonustrack (approved):
Looks good thanks!

ChaituVR (approved):

PR #258, opened 2021-12-21
[shinobi-staked-in-tge-farms] add strategy for Shinobi DEX on Ubiq, in conjunction with the Ubiq TGE framwork

scrappydew:
this passes the checklist on my end and passes all tests (including 500+ addresses)


scrappydew:
updated to include another strategy, this now covers:

- Shinobi LP pools on their own ("shinobi" strategy name)
- Shinobi LP tokens which have been deposited into a TGE Framework based farm ("shinobi-staked-in-tge-farms")

Shinobi is compatible with Uniswap v2 and serves its own Subgraph API like Uniswap

src/strategies/index.ts
```diff
@@ -268,6 +270,8 @@ const strategies = {
   masterchef,
   sushiswap,
   uniswap,
+  shinobi,
+  'shinobi-staked-in-tge-farms': shinobiInTGEFarms,
```
ChaituVR, line 274: Would be great if you can do separate PRs for two strategies. will be easy to test/review/integrate

ChaituVR (changes requested):

src/strategies/index.ts
scrappydew: i intentionally did it as two separate, clean commits which each contain file changes for only a single one of the strategies, but i can separete this into two PRs if youd prefer just let me know

steps on creating strategies weren't clear on this

scrappydew: alright, ive reduced it back to a single strategy

sorry for the confusion

scrappydew:
this contains just a single strategy again

src/strategies/shinobi-staked-in-tge-farms/index.ts
```diff
@@ -0,0 +1,158 @@
...
+  }
+};
+
+const getLP = async (stakedLP, snapshot) => {
+  let lp = [];
+  let results = [];
+  do {
+    results = await getAllLP(lp.length, stakedLP, snapshot);
+    lp = lp.concat(results);
+  } while (results.length === 1000);
```
ChaituVR, line 53: How many times subgraph will be called here ?  There is a limit of 5 API calls for a strategy https://github.com/snapshot-labs/snapshot-strategies#checklist-for-adding-a-new-strategy

ChaituVR (changes requested):

src/strategies/shinobi-staked-in-tge-farms/index.ts
scrappydew: i copied this code from an existing strategy that uses that same code, I didnt modify it. 

i can look for another example

ChaituVR:
Ok, this is the same strategy as the `dfyn-staked-in-farms` strategy. except for graph URL change, why note pass the URL from option params to the old strategy itself?  any reason to do it as separate strategy?

scrappydew:
> 

i'll look into doing it this way instead, thanks

PR #263, opened 2021-12-21
[path-balance-staked-and-locked] fix issue with array type in schema

PathDAO:
Fixes: fix issue with array type in schema


PathDAO:
Sorry seems like my schema json had some issue for adding arrays into the schema. have fixed this

PR #264, opened 2021-12-22
[path-balance-staked-and-locked] remove schema.json to fill params manually

PathDAO:
Fixes: Removed the schema.json as it is causing issues for me to add the params for array


PathDAO:
Hi, have removed the schema.json to fill the params in manually as this will be more flexible if things were to have issues in the schema.json file

PR #268, opened 2021-12-23
[multichain] Support multichain strategy block fetching from custom API

seguido:

The existing multichain strategy requires a working subgraph on every chain that wants to be included. Some chains, such as HECO, CRONOS or HARMONY are not supported at the moment by the subgraph hosted service and are therefor unable to be added and used. As a solution to this, I propose adding an external API source as an option to use in place of subgraphs for block number fetching.

If the strategy's options contains the 'blockApi' property, it will be used as a base url from where to fetch block height numbers. This allows developers to create their own custom APIs that, given a timestamp, return the block height for every chain they support, no matter if there is a subgraph for that chain or not. **A single API call** should be done to fetch all block numbers.

In order to integrate with this, the API should
- Support GET requests to blockApi?timestamp=xxxxxxxxx, where blockApi is the url for the working endpoint.
- Return a JSON with a key named `blocks` which contains an object with chainIds as keys and blockNumber as values.
An example of a working response would be 
`{
   "timestamp":"1640185827",
   "blocks":{
      "25":678246,
      "56":13700191,
      "128":11097265,
      "137":22832200,
      "250":25715540,
      "1285":1141104,
      "42161":4021906,
      "42220":10521555,
      "43114":8573779,
      "1666600000":20812976
   }
}`

### DIFF
--- a/src/strategies/multichain/README.md
+++ b/src/strategies/multichain/README.md
@@ -4,9 +4,42 @@ If you want to calculate the balance from various chains like Ethereum, Binance 
 
 In multichain strategy, the params should define sub strategies which would use different networks mentioned in the field to combine the voting power.
 
+In order to provide multichain functionality, this strategy requires a way for calculating which block number should be used on additional chains: If a snapshot was created on block 125 on mainnet, it needs to find the timestamp for than block and go find which block number corresponds to that same timestamp on every other wanted chain. This way it can accurately represent an address' voting power at a given point in time. In order to do this, it supports 2 different mechanisms:
+
+- [DEFAULT] Querying a block subgraph. If a working block info subgraph is found for that chain, it can be passed into the strategy's options in the "graph" objects, this will allow the strategy to query that subgraph for the given chain and fetch the block number from there. An example of a graph object could be:
+```json
+ "graphs": {
+    "56": "https://api.thegraph.com/subgraphs/name/apyvision/block-info",
+    "137": "https://api.thegraph.com/subgraphs/name/sameepsi/maticblocks"
+  }
+  ```
+
+- Integrating with a custom API (overrides subgraph option if present). TheGraph doesn't support every existing chains on their hosting services, therefor finding a subgraph for them can be challenging without launching and maintaining an independant graph-node. This option comes in as an alternative for developers who wish to integrate even on chains without a subgraph. In order to integrate an API must be created to fulfill the block fetching functionality on every chain that wants to be supported. A single HTTP receiving a timestamp should return the block number for every desired chain.
+The setting that should be set is `blockApi` and it should be an url pointing to an endpoint that:
+  - Supports GET calls passing timestamp as a query parameter. (Ex: If blockApi is "https://myCustomApi.com/blocks", it should support a GET to https://myCustomApi.com/blocks?timestamp=xxxxxx)
+  - Return a json object containing at least 1 key named "blocks" with a map from chainId to block number for *every chain* desired. An example of a valid api response could be:
+```json
+     {
+      "timestamp":"1640185827",
+      "blocks":{
+          "25":678246,
+          "56":13700191,
+          "128":11097265,
+          "137":22832200,
+          "250":25715540,
+          "1285":1141104,
+          "42161":4021906,
+          "42220":10521555,
+          "43114":8573779,
+          "1666600000":20812976
+      }
+    }
+```
+
+
 Here is an example of parameters:
 
-In the below example, the tokens on the three networks namely ethereum, polygon and bsc denotes combined voting power.
+In the below example, the tokens on the three networks namely ethereum, polygon and bsc denotes combined voting power and block numbers on each chain is searched from a subgraph query.
 
 
 ```json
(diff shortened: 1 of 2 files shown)


src/strategies/multichain/index.ts
```diff
@@ -62,6 +63,43 @@ async function getChainBlocks(
   return chainBlocks;
 }
 
+async function getChainBlocksFromApi(
+  snapshot,
+  provider,
+  options
+): Promise<any> {
```
mktcode, line 70: As long as `resp.blocks` isn't a promise.... the return value here is an object I guess.

mktcode (approved):
Looks good. :)

src/strategies/multichain/index.ts
```diff
@@ -62,6 +63,43 @@ async function getChainBlocks(
...
+  const apiUrl = `${options.blockApi}?timestamp=${timestamp}`;
+  const resp = await fetch(apiUrl, {
+    method: 'GET',
+    headers: {
+      Accept: 'application/json',
+      'Content-Type': 'application/json'
+    }
+  }).then(r => {
+    return r.json()
+  });
```
mktcode, line 84:
```suggestion
  }).then(r => r.json());
```

magic

ChaituVR:
@seguido Thanks for the PR! would be great if you can add documentation to this option in [README.md](https://github.com/snapshot-labs/snapshot-strategies/blob/automated-lint/src/strategies/multichain/README.md)

seguido:
> @seguido Thanks for the PR! would be great if you can add documentation to this option in [README.md](https://github.com/snapshot-labs/snapshot-strategies/blob/automated-lint/src/strategies/multichain/README.md)

Absolutely, I just added a more detailed explanation as to how the strategy works + some additional documentation on how to support this new alternative for block number fetching. Please let me know if it's clear and detailed enough

bonustrack (commented):
@seguido How the API turn timestamp to block height, is the code open source? Do you think there is a way to have the same logic on the strategy code itself? And does it take if there is a block 100 at timestamp 12000000 and a block 101 at timestamp 12000015, if you query the API with a timestamp of 12000014 would it give the block 100 or 101?

seguido:
> @seguido How the API turn timestamp to block height, is the code open source? Do you think there is a way to have the same logic on the strategy code itself? And does it take if there is a block 100 at timestamp 12000000 and a block 101 at timestamp 12000015, if you query the API with a timestamp of 12000014 would it give the block 100 or 101?

Hey there! Absolutely I can share the code with you guys and we can look into the best way of adding this. Let me give some additional insight on how I coded our API to support this feature.

So in our case, we needed support for 10 chains atm, 7 of which have a working subgraph. So for those 7 chains with a subgraph, the timestamp to block number query is done from the subgraph itself just as how the multichain strategy does it right now.

For the 3 remaining chains, a _smart binary search_ is performed. So basically:
- We get the difference in timestamp from the current block to the desired timestamp
- We estimate how many blocks that difference would account for assuming constant block times
- We get the block that we would expect to be the target one
- We recalculate the block time based on the new data (difference in timestamps vs nr of blocks we moved)
- Repeat the process from the start.

The code for it can be found here: https://github.com/beefyfinance/beefy-vote-api/blob/master/server/helpers/blocks.ts . Feel free to check it out and ask for any insight on it.

Also, for the specific case you asked about, it would give the immediate block with timestamp equal or less to the desired timestamp. So in the case mentioned it would return the block nr 100.

Regarding adding this to the strategy itself, it clearly breaks the max 5 api calls rule that's in place + it can break the 10 seconds evaluation bit as well. Since the recursive function needs to be called 3/6 times for it to find the block on a single chain. I implemented on our API a cache for this, so that it only finds the blocks for a given timestamp once and all subsequent block requests for that same proposal would be returned instantly from the cache. I think that if we could offer a similar cache service this would be viable but without it we would have to execute various api calls on every single vote and that breaks the stated rules.

src/strategies/multichain/README.md
```diff
@@ -4,9 +4,42 @@ If you want to calculate the balance from various chains like Ethereum, Binance
 
 In multichain strategy, the params should define sub strategies which would use different networks mentioned in the field to combine the voting power.
 
+In order to provide multichain functionality, this strategy requires a way for calculating which block number should be used on additional chains: If a snapshot was created on block 125 on mainnet, it needs to find the timestamp for than block and go find which block number corresponds to that same timestamp on every other wanted chain. This way it can accurately represent an address' voting power at a given point in time. In order to do this, it supports 2 different mechanisms:
```
mktcode, line 7:
```suggestion
In order to provide multichain functionality, this strategy requires a way for calculating which block number should be used on additional chains: If a snapshot was created on block 125 on mainnet, it needs to find the timestamp for that block and go find which block number corresponds to that same timestamp on every other wanted chain. This way it can accurately represent an address' voting power at a given point in time. In order to do this, it supports 2 different mechanisms:
```

src/strategies/multichain/README.md
```diff
@@ -4,9 +4,42 @@ If you want to calculate the balance from various chains like Ethereum, Binance
...
+
+- [DEFAULT] Querying a block subgraph. If a working block info subgraph is found for that chain, it can be passed into the strategy's options in the "graph" objects, this will allow the strategy to query that subgraph for the given chain and fetch the block number from there. An example of a graph object could be:
```
mktcode, line 9:
```suggestion
- [DEFAULT] Querying a block subgraph. If a working block info subgraph is found for that chain, it can be passed into the strategy's options in the "graphs" object. This will allow the strategy to query that subgraph for the given chain and fetch the block number from there. An example of a graphs object could be:
```

mktcode (approved):

ChaituVR:
@seguido I think even in your API, you are calling subgraph right?

https://github.com/beefyfinance/beefy-vote-api/blob/570c265e7ce6615014cb8384869dcbe22d631368/server/helpers/blocks.ts#L66

 I don't understand what is the exact need for this API 🤔As of now, there is no restriction on multichain strategy for the number of chains right? you can just pass these into graphs param, let me know if I am thinking in a wrong direction

bonustrack:
@ChaituVR The need of this change is for chains that doesn't have any subgraph support, like Cronos

bonustrack (approved):
Looks good!

PR #270, opened 2021-12-26
added erc721-exist strategy

dividebyzeroeth:
This is a modification of the ERC721 strategy, returning a `1` if a given ERC721 exists, and `0` if it does not. 

For example, an address with 100 `Hashmasks (HM)` tokens and an address with a single `Hashmasks (HM)` token will both return `1`.

*Warning: This strategy does not prevent a unique voter from using multiple addresses.*

### DIFF
--- a/src/strategies/erc721-exist/README.md
+++ b/src/strategies/erc721-exist/README.md
@@ -1,0 +1,14 @@
+# erc721-exists
+
+This strategy checks if a specific ERC721 NFT exists for a given address. 
+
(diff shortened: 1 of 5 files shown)


ChaituVR:
Hi, @dividebyzeroeth Thank you for the PR but there is already a strategy that does this. just name is different (we are looking for ways to rename it ) https://snapshot.org/#/strategy/erc20-with-balance , please feel free to reopen or create new PR if needed

PR #271, opened 2021-12-27
[contract-call-multi-outputs] Create contract-call-multi-outputs strategy

ChefSnoopy:
Create contract-call-multi-outputs strategy

Fixes # .

- 
- 
-

### DIFF
--- a/src/strategies/contract-call-multi-outputs/index.ts
+++ b/src/strategies/contract-call-multi-outputs/index.ts
@@ -1,0 +1,40 @@
+import { formatUnits } from '@ethersproject/units';
+import { multicall } from '../../utils';
+
+export const author = 'in19farkt';
+export const version = '0.1.0';
+
+function getArgs(options, address: string) {
+  const args: Array<string | number> = options.args || ['%{address}'];
+  return args.map((arg) =>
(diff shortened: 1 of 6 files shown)


src/strategies/contract-call-multi-outputs/index.ts
```diff
@@ -0,0 +1,40 @@
...
+      options.address,
+      options.methodABI.name,
+      getArgs(options, address)
+    ]),
+    { blockTag }
+  );
+  return Object.fromEntries(
+    response.map((value, i) => [
+      addresses[i],
+      parseFloat(formatUnits(value[options.output].toString(), options.decimals))
```
ChaituVR, line 37: I feel it is better to change `contract-call` strategy itself and make output param optional

```
      parseFloat(formatUnits(options.output ? value[options.output].toString() : value.toString(), options.decimals))
```

ChaituVR (changes requested):

src/strategies/contract-call-multi-outputs/index.ts
ChefSnoopy: so i can submit a new pull request which update the contract-call strategy with the code above ?

PR #279, opened 2022-01-04
[zorro] Zorro Proof of Personhood Strategy

RoboTeddy:
Basic 1 person 1 vote strategy

### DIFF
--- a/src/strategies/index.ts
+++ b/src/strategies/index.ts
@@ -228,6 +228,7 @@ import * as squidDao from './squid-dao';
 import * as pathBalanceStakedAndLocked from './path-balance-staked-and-locked';
 import * as bottoDao from './botto-dao';
 import * as genart from './genart';
+import * as zorro from './zorro';
 
 const strategies = {
   'nouns-rfp-power': nounsPower,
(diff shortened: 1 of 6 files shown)


bonustrack:
@RoboTeddy Looks great! Just some suggestions: 
- You could add a schema file on the strategy like this https://github.com/snapshot-labs/snapshot-strategies/blob/master/src/strategies/erc20-balance-of/schema.json this will generate a form on the UI for the strategy params and make it more handy for space owner to setup.
- What you think about adding an optional parameter to define the value to give if you are a verified person, instead of giving always 1 it could give 100 or a custom number, this would be useful if a space want to use multiple strategies like ERC20 + zorro

RoboTeddy:
> @RoboTeddy Looks great! Just some suggestions:
> 
> * You could add a schema file on the strategy like this https://github.com/snapshot-labs/snapshot-strategies/blob/master/src/strategies/erc20-balance-of/schema.json this will generate a form on the UI for the strategy params and make it more handy for space owner to setup.
> * What you think about adding an optional parameter to define the value to give if you are a verified person, instead of giving always 1 it could give 100 or a custom number, this would be useful if a space want to use multiple strategies like ERC20 + zorro

Done!

bonustrack (commented):
Looks great!

bonustrack (approved):

PR #282, opened 2022-01-05
[erc20-balance-of-indexed] Add erc20-balance-of-indexed strategy

### LINKED ISSUE #281: Create strategy for multiplying token balance by an index
### Expected behavior

Each voter's balance for the specified token at `address` should be scaled by an index value read from the `index` function on the specified `indexAddress` contract.

This is desirable for Olympus-style protocols like OlympusDAO and KlimaDAO which have a "wrapped staked" token whose voting power is tied to an ever-increasing index value.

0xAurelius:
Fixes #281 

- Add a strategy to multiply token balances by an ever-increasing index value.


0xAurelius:
I got all the tests passing on my local machine and the results are as expected
image

bonustrack (approved):

PR #285, opened 2022-01-05
[balancer-poolid] snapshot strategy + pagination

jo-chemla:
Snapshot Strategy for balancer
The main differences with the `balancer` strategy (from bonustrack on which this implementation is based) are:
 - the use of the `poolId` parameter, which makes the graphql query return only interesting payload.
 - and `pagination` using `skip: N`, which makes it better suited for balancer pools with more than 1.000 shares


src/strategies/index.ts
```diff
@@ -460,7 +461,8 @@ const strategies = {
   'path-balance-staked-and-locked': pathBalanceStakedAndLocked,
   'sumami-holders': sUmamiHolders,
   zorro,
-  'volt-voting-power': voltVotingPower
+  'volt-voting-power': voltVotingPower, 
+  'balancer-poolId': balancerPoolId
```
ChaituVR, line 465: Hi @jo-chemla Strategy name should be all lowercase maybe make this `balancer-poolid` or `balancer-pool-id`

ChaituVR (changes requested):

src/strategies/index.ts
jo-chemla: Thanks, just did change that

src/strategies/index.ts
```diff
@@ -230,6 +230,7 @@ import * as bottoDao from './botto-dao';
 import * as genart from './genart';
 import * as zorro from './zorro';
 import * as voltVotingPower from './volt-voting-power';
+import * as balancerPoolid from './balancer-poolid';
```
ChaituVR, line 233: Hi @jo-chemla Rename the folder name to lower case, like here `balancer-poolid`

ChaituVR (changes requested):

src/strategies/index.ts
jo-chemla: Ah I did that but it seems like github did not handle it

jo-chemla: Ok that's done thanks!

bonustrack (approved):
Looks good

ChaituVR (approved):

PR #286, opened 2022-01-06
Fix CALL_EXCEPTION Error

ChefSnoopy:
Fixes # .

- 
- 
-

### DIFF
--- a/src/strategies/contract-call/index.ts
+++ b/src/strategies/contract-call/index.ts
@@ -20,26 +20,37 @@ export async function strategy(
   snapshot
 ) {
   const blockTag = typeof snapshot === 'number' ? snapshot : 'latest';
-  const response = await multicall(
-    network,
-    provider,
-    [options.methodABI],
-    addresses.map((address: any) => [
-      options.address,
-      options.methodABI.name,
-      getArgs(options, address)
-    ]),
-    { blockTag }
-  );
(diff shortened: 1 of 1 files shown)


bonustrack (commented):
Hi, thanks for the PR but the error is on purpose, if the call fail we need to know that so we can report it on the frontend. Call could fail just because the node is offline, in this case we dont want the system to think voting power is 0.

ChefSnoopy:
> Hi, thanks for the PR but the error is on purpose, if the call fail we need to know that so we can report it on the frontend. Call could fail just because the node is offline, in this case we dont want the system to think voting power is 0.

can i add a new strategy with the default 0 ? Otherwise we need to customize the strategies for each proposal.
Our voting-power is a combination of multiple contracts ,we will always add new pools in voting-power.
now we have so many old voting proposal,When users view historical data pages, FE need to query the voting-power of each voter.
For example, I want to query two contracts ，C1:(deployed 10),C2:(deployed 100)
now i  have a proposal with snapshot block number 200, i can get voting-power with strategies({s1:contract-call C1, s2:contract-call C2}),
but if i have a proposal with snapshot block number 50 , i use the same strategies({s1:contract-call C1, s2:contract-call C2}) , your API will report error with call-error for C2 , if use my new pull code ,it will have no error ,just return 0 for C2

PR #287, opened 2022-01-06
[cyberkongz] removed liquidity pools from voting and supply

frosti-eth:
- blacklist LP addresses from voting and remove their balances from total supply


src/strategies/cyberkongz/index.ts
```diff
@@ -28,15 +43,25 @@ export async function strategy(
...
     blockTag
   });
 
+  response.forEach((value: any, i: number) => {
+    const address = calls[i][2][0];
+    if (Math.floor(i / addresses.length) == 2 && blacklistedAddresses.find((add) => add === address)) {
+      nanaSupply -= value;
+    }    
+  });
+
```
ChaituVR, line 58: Hello @frosti-eth we can't update this as it can affect existing space that are using this strategy

ChaituVR (changes requested):

frosti-eth:
Hello,

The only space that uses this strategy is our own, and our contract addresses are hardcoded in there. I think we’re good to merge this, but if you like we could submit a _v2 PR? Do you need a wallet signature somewhere?

frosti

frosti-eth:
Did we want to run the 500 address test again? That part of the code was not changed from the existing version

ChaituVR:
@frosti-eth I think it is better to create a -v2 strategy than changing the existing one. so we don't effect existing proposals :)

PR #289, opened 2022-01-06
[contract-call-default-zero] Add new strategy contract-call-default-zero

ChefSnoopy:
Fixes # .

- Return default 0 when call-exception
- 
-

### DIFF
--- a/src/strategies/contract-call-default-zero/index.ts
+++ b/src/strategies/contract-call-default-zero/index.ts
@@ -1,0 +1,56 @@
+import { formatUnits } from '@ethersproject/units';
+import { multicall } from '../../utils';
+
+export const author = 'snoopy';
+export const version = '0.1.0';
+
+function getArgs(options, address: string) {
+  const args: Array<string | number> = options.args || ['%{address}'];
+  return args.map((arg) =>
+    typeof arg === 'string' ? arg.replace(/%{address}/g, address) : arg
+  );
+}
+
+export async function strategy(
+  space,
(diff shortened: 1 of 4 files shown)


src/strategies/contract-call-default-zero/index.ts
```diff
@@ -0,0 +1,56 @@
+import { formatUnits } from '@ethersproject/units';
+import { multicall } from '../../utils';
+
+export const author = 'snoopy';
```
ChaituVR, line 4:
```suggestion
export const author = 'ChefSnoopy';
```

src/strategies/contract-call-default-zero/index.ts
```diff
@@ -0,0 +1,56 @@
...
+      ])
+    );
+  } catch (e) {
+    return Object.fromEntries(
+      addresses.map((address) => [
+        address,
+        parseFloat(formatUnits('0', options.decimals))
+      ])
+    );
+  }
```
ChaituVR, line 55: There is a chance that even if there is a network issue, this will return 0 voting power (can change voting power in a proposal in the future just in case there is a network error)

ChaituVR (changes requested):

src/strategies/contract-call-default-zero/index.ts
ChefSnoopy: Yes ,i know that , just do this on purpose , please merge this code ,thanks

PR #290, opened 2022-01-06
[aavegotchi-agip-17] new strategy for agip17

candoizo:
- updates aavegotchi-agip strategy to calculate realm voting power

### DIFF
--- a/src/strategies/aavegotchi-agip-17/examples.json
+++ b/src/strategies/aavegotchi-agip-17/examples.json
@@ -1,0 +1,15 @@
+[
+  {
+    "name": "AGIP 17: Voting power for GHST value of parcels",
+    "strategy": {
+      "name": "aavegotchi-agip-17",
+      "params": {
+        "tokenAddress": "0x86935f11c86623dec8a25696e1c19a8659cbf95d",
+        "symbol": "GOTCHI"
+      }
+    },
+    "network": "137",
+    "addresses": ["0x027Ffd3c119567e85998f4E6B9c3d83D5702660c"],
+    "snapshot": 12089223
+  }
+]
--- a/src/strategies/aavegotchi-agip-17/examples.json
+++ b/src/strategies/aavegotchi-agip-17/examples.json
@@ -4,12 +4,11 @@
     "strategy": {
       "name": "aavegotchi-agip-17",
       "params": {
-        "tokenAddress": "0x86935f11c86623dec8a25696e1c19a8659cbf95d",
-        "symbol": "GOTCHI"
+        "symbol": "REALM"
       }
     },
     "network": "137",
     "addresses": ["0x027Ffd3c119567e85998f4E6B9c3d83D5702660c"],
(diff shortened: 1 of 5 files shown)


src/strategies/aavegotchi-agip/index.ts
```diff
@@ -85,9 +88,26 @@ export async function strategy(
...
-    const { id, gotchisOwned } = addrInfo;
+  result.users.map(({ id, gotchisOwned, parcelsOwned }) => {
+    let realmVotingPowerValue = 0;
+    if (parcelsOwned.length > 0) {
+      parcelsOwned.map(({ size }) => {
+        let votePower = realmSizeVotePower[size];
+        if (isNaN(votePower)) votePower = 0;
+        realmVotingPowerValue += votePower;
+      });
+    }
```
ChaituVR, line 109: Hi @candoizo Existing proposals could be affected by this change, especially the proposal that is open 
https://snapshot.org/#/aavegotchi.eth/proposal/0x396844d504394f7f335f070d443c33455300ee21d90db6c4b089760a3e0469bf

It is better to create a `-v2` strategy for this

ChaituVR (changes requested):

src/strategies/aavegotchi-agip/index.ts
candoizo: alright ive made the changes requested

src/strategies/aavegotchi-agip-17/index.ts
```diff
@@ -0,0 +1,66 @@
...
+  137: 'https://api.thegraph.com/subgraphs/name/aavegotchi/aavegotchi-core-matic'
+};
+
+export async function strategy(
+  _space,
+  network,
+  provider,
+  addresses,
+  options,
+  snapshot
```
ChaituVR, line 16:
```
'snapshot' is defined but never used
```
Hi @candoizo is there some way to pass snapshot block number into the subgraph? without it, the voting power may change in the future or in the middle of voting a proposal etc. the VP should be based on snapshot block number to have consistent results

src/strategies/aavegotchi-agip-17/examples.json
```diff
@@ -0,0 +1,15 @@
+[
+  {
+    "name": "AGIP 17: Voting power for GHST value of parcels",
+    "strategy": {
+      "name": "aavegotchi-agip-17",
+      "params": {
+        "tokenAddress": "0x86935f11c86623dec8a25696e1c19a8659cbf95d",
```
ChaituVR, line 7: This is not used anymore?

ChaituVR (changes requested):

src/strategies/aavegotchi-agip-17/index.ts
candoizo: good point, fixed in recent commit thanks

bonustrack (approved):

ChaituVR (approved):

PR #294, opened 2022-01-09
[aavegotchi] newer staking pool power

candoizo:
- add logic to calculate new staking pools


ChaituVR:
Hi @candoizo Existing proposals could be affected by this change, especially the proposal that is open 
https://snapshot.org/#/aavegotchi.eth/proposal/0x396844d504394f7f335f070d443c33455300ee21d90db6c4b089760a3e0469bf

It is better to create a `-v2` strategy for this

bonustrack:
There should be a v2 of the strategy if you don't want past proposal to have different results

candoizo:
In this case after reviewing with team the preferred option is to apply the change on the existing strategy

src/strategies/aavegotchi/index.ts
```diff
@@ -1,7 +1,7 @@
 import { multicall } from '../../utils';
 
-export const author = 'vfatouros';
-export const version = '0.1.0';
+export const author = 'candoizo';
+export const version = '0.1.1';
 
 const tokenAbi = [
```
ChaituVR, line 6: Please replace ABI with this one 
```
[
  'function balanceOf(address account) view returns (uint256)',
  'function stakedInCurrentEpoch(address _account) view returns (tuple(address poolAddress, string poolName, string poolU
rl, uint256 rate, uint256 amount)[] _staked)',
  'function staked(address _account) view returns (uint256 ghst_, uint256 poolTokens_, uint256 ghstUsdcPoolToken_)'      
]
```

ChaituVR (changes requested):

src/strategies/aavegotchi/index.ts
candoizo: fixed!

bonustrack (approved):

ChaituVR (approved):

PR #296, opened 2022-01-10
[dog-dao] snapshot strategy

zencephalon:
Allows holders of $DOG who have deposited into Rari FUSE and SnoopDAO to vote using their deposited $DOG.


src/strategies/dog-dao/index.ts
```diff
@@ -0,0 +1,74 @@
...
+    snoopBalances,
+    sSnoopBalances
+  ] = await Promise.all([
+    dogMulti.execute(),
+    fdogMulti.execute(),
+    snoopMulti.execute(),
+    sSnoopMulti.execute()
+  ]);
+
+  console.log(dogBalances);
```
ChaituVR, line 61:
```suggestion
```

ChaituVR:
Hi @zencephalon Any reason why you are not using 4 `erc20-balance-of` strategies instead of this strategy? 🤔  let me know if there is any particular reason :)

zencephalon:
@ChaituVR Thanks for the review! Two major reasons:

1) Rari Fuse's fDOG-45 has a 5:1 redemption ratio to $DOG. Unfortunately it seems like the `math` strategy doesn't support division yet so I couldn't use that to implement this
2) We anticipate adding tiered NFTs with weighted voting power (e.g. 65,000 votes) in the future, so having a strategy setup at that point seems easiest to update

ChaituVR:
@zencephalon I think for 1. it is better to have a new method in `math` strategy for 2. we can look into it in future when you have it

I am not able to accept this PR because now there is a limit on the number of strategies a space can use and if this is merged people may point this strategy and request to add their own strategies that have multiple strategies in it, which can kill the point of having a limit 😅

zencephalon:
Yeah, fair, I think I can use the weighted strategies to do this for now!

PR #298, opened 2022-01-11
convert npm to yarn lock file and update readme

mktcode:
Fixes 23 on the TODO

- run `yarn import`
- removed `package-lock.json`
- updated readme
- updated workflow


ChaituVR:
Sorry for the delay @mktcode I thought of trying this change in local and Heroku before merging. i tried it now and it works great, we can merge this soon :+1:

mktcode:
Yeah for the strategies repo itself it should be fine. The only problem is the score repo I think. Because NPM runs the `prepare` script after installing from a GitHub URL (like we do with strategies repo). Yarn doesn't. That's why the build fails. Technically, changing `prepare` to `postinstall` could actually do the trick. I'll try that later.

mktcode:
I think best way to solve it, is to actually publish `@snapshot-labs/strategies` to npm and then use the `@latest` alias where we use it. This will then be similar to using `#master`. The version number can easily be upped on push to this repo. I can adjust [this workflow](https://github.com/snapshot-labs/snapshot-strategies/blob/master/.github/workflows/publish.yml).

Then NPM and Yarn will both install the same pre-built package.

bonustrack (approved):

PR #280, opened 2022-01-04
[frax-finance-with-delegations] Frax Finance Delegation

FortisFortuna:
Adds delegation capabilities

PR #299, opened 2022-01-11
Custom Frax Finance delegation strategy

FortisFortuna:
Closes https://github.com/snapshot-labs/snapshot-strategies/pull/280

Custom getDelegations strategy.

Assume two voters, A and B
A has 1000, B has 500.
A then delegates to B.

Right now, using the existing `delegation` strategy, B is showing 1000 instead of the sum, which should be 500 + 1000 = 1500. Final result should be:
A: 0
B: 1500

Example from test:
Screenshot from 2022-01-11 10-58-25


FortisFortuna:
Closing. Didn't

PR #300, opened 2022-01-13
[saddle-finance-v2] New Snapshot Strategy With Logic Fix

ug02fast:
- Change logic away from taking into consideration total vest amount to unclaimed and vested amount.
Closes https://github.com/saddle-finance/tickets/issues/244


ChaituVR:
Hi @ug02fast This can potentially change voting power in existing proposals (including the one that is still active) so better to create a new `-v2` strategy

ug02fast:
> Hi @ug02fast This can potentially change voting power in existing proposals (including the one that is still active) so better to create a new `-v2` strategy

Good idea. Thanks.

src/strategies/saddle-finance/index.ts
```diff
@@ -49,25 +50,24 @@ export async function strategy(
...
 
-  const beneficiaries = multicall(
+  const vestedAndUnclaimedAmountRes = multicall(
     network,
     provider,
     abi,
     vestingContractAddrs.map((vestingContractAddress: any) => [
       vestingContractAddress.toLowerCase(),
-      'beneficiary'
+      'vestedAmount'
```
hammeiam: The logic here would need to be to get both the `vestedAmount` as well as `released` from the same contract, and use `walletBalance + vestedAmount - released` as the voting amount

hammeiam: I was incorrect about this logic ^

bonustrack (approved):

PR #301, opened 2022-01-13
[erc721-with-tokenid-range-weights-simple] Strategy

gregegan:
This strategy is a modification of erc721-with-tokenid-range-weights. This strategy allows for multiple votes from a single wallet, and different token id's represent different weights. For example, a wallet containing two ERC721 TokenIDs from weight 1, and three ERC721 TokenIDS from weight 4, would receive fourteen votes. (2*1+3*4). This other difference, is this strategy does not require extension of OpenZeppelin's erc721Enumerable contract.


src/strategies/erc721-with-tokenid-weighted-types/index.ts
```diff
@@ -0,0 +1,57 @@
...
+  addresses,
+  options,
+  snapshot
+) {
+  const blockTag = typeof snapshot === 'number' ? snapshot : 'latest';
+  const responsePrimary = await multicall(
+    network,
+    provider,
+    abi,
+    options.tokenIdsPrimary.map((id: any) => [
```
ChaituVR, line 23: Hello @gregegan Instead of passing all tokenIDs into strategy params, would be great if we can do it like ranges like in `erc721-with-tokenid-range-weights` strategy 

Also, the name can be similar to that too, maybe like `erc721-with-tokenid-range-weights-v2` or `erc721-tokenid-range-weights-simple` or other

ChaituVR (changes requested):

ChaituVR:
From the discussion in discord:
- Documentation needs to be updated with new strategy params
- would be great if we can add some limits (a limit of 5 ranges that can send up to 7k tokens at once)

bonustrack (approved):

PR #302, opened 2022-01-14
[galaxy-nft-with-score] Add matic support and switch subgraphs

alberthaotan:
Feat: [galaxy-nft-with-score] Add matic support and switch subgraphs


ChaituVR:
@alberthaotan Can this affect the voting power of previous proposals? If so it would be better to create a new strategy with `-v2` as this strategy is being used by 3 different spaces

ChaituVR (changes requested):
Also, this strategy takes longer than 20 secs when we pass 500 addresses, you can test this in your local using:
```
npm test --strategy=galaxy-nft-with-score --more=500
```

ChaituVR (changes requested):
From the previous comment:
@alberthaotan Can this affect the voting power of previous proposals? If so it would be better to create a new strategy with -v2 as this strategy is being used by 3 different spaces

alberthaotan:
Hi, Chaitanya. Thanks for your review. This change is to switch to new subgraph and won't affect previous results. Also, 3 spaces that use this strategy are just testing , so should be safe to ignore them.

ChaituVR:
Thanks for confirming @alberthaotan  :)

bonustrack (approved):

ChaituVR (approved):

PR #306, opened 2022-01-17
[ocean-dao-brightid] Update `ocean-dao-brightid` strategy

trizin:
-  Getting the BrightId registry network & block and the snapshot network & block from options object.


src/strategies/ocean-dao-brightid/index.ts
```diff
@@ -15,6 +15,9 @@ export async function strategy(
   options,
   snapshot
 ) {
+  provider = getProvider(options.chainId);
+  snapshot = options.chainBlock;
```
bonustrack, line 19: Any reason to use a snapshot defined in the strategy param? This will not make it easy to use, it will force to update space settings with the proper snapshot block number everytime you want to make a new proposal.

trizin: Hey there! I'm planning to pass strategy params when creating the proposal using the API. I don't think it will be a problem.

ChaituVR: @trizin Problem is that users have to change the strategies settings everytime before a new proposal

trizin: Thanks, I removed this part.

src/strategies/ocean-dao-brightid/index.ts
```diff
@@ -27,16 +27,20 @@ export async function strategy(
     options.strategies.some((x) => !ALLOWED_STRATEGIES.includes(x.name))
   )
     return [];
+
+  const brightIdNetwork = options.brightIdNetwork || network;
+  const brightIdProvider = getProvider(brightIdNetwork);
+  const brightIdSnapshotBlock = options.brightIdSnapshotBlock || snapshot;
```
ChaituVR, line 33: Even `options.brightIdSnapshotBlock` has to be changed in space settings every time, now I feel it is better to make something like `multichain` strategy, where you can get block number of brightID based on snapshot block timestamp

https://github.com/snapshot-labs/snapshot-strategies/blob/master/src/strategies/multichain/index.ts#L38-L61

ChaituVR (changes requested):

src/strategies/ocean-dao-brightid/index.ts
trizin: Thanks, I've updated the strategy.

bonustrack:
@trizin The strategy fail with 500 test addresses due to node limitation, I think this come from the ERC20 on Polygon, can you try without? Here is the error I see:
```
missing revert data in call exception (error={"reason":"processing response error","code":"SERVER_ERROR","body":"{\"jsonrpc\":\"2.0\",\"id\":51,\"error\":{\"code\":-32000,\"message\":\"execution aborted (timeout = 5s)\"}}\n","error":{"code":-32000},"requestBody":"{\"method\":\"eth_call\",\"params\":[{\"to\":\"0xeefba1e63905ef1d7acba5a8513c70307c1ce441\",\"data\":\"0
```

src/strategies/ocean-dao-brightid/examples.json
```diff
@@ -4,86 +4,102 @@
...
-          {
-            "name": "uniswap",
-            "params": {
-              "symbol": "OCEAN",
-              "address": "0x967da4048cD07aB37855c090aAF366e4ce1b9F48",
-              "decimals": 18
+        "graphs": {
+          "137": "https://api.thegraph.com/subgraphs/name/sameepsi/maticblocks"
+        },
+        "chains": {
```
ChaituVR, line 15: I think it would be better to keep it same as before (strategies with limits) 

If users want to get voting power from other networks, they can pass the `multichain` strategy into this strategy, right?

trizin:
@bonustrack thanks, that worked!

src/strategies/ocean-dao-brightid/examples.json
ChaituVR: @trizin You had a look at my comment?

trizin: @bonustrack has told me that I can remove the limits. I was going to use the `multichain` strategy, but we only have one BrightId registry contract deployed and we'd like to only use that contract thus we are unable to use the `multichain` strategy directly. I've solved that problem using some code from the `multichain` strategy.

ChaituVR: Sorry for the confusion, I mean for this option `chains` , we can make it `strategies` like before right? I don't understand the need to have chains. users could pass a multichain strategy into this strategy and make it work for all chains right?


Also to get block with timestamp we now have a new API from snapshot, can you please make it the same as here https://github.com/snapshot-labs/snapshot-strategies/blob/7b60259ff6bb0eb79e8b2ec9b4cb5b29bb41d962/src/strategies/multichain/index.ts#L8-L27

Sorry for not mentioning it earlier

trizin: Thanks, the new API works great! I updated the code.

bonustrack (approved):
Looks good!

ChaituVR (approved):

PR #313, opened 2022-01-20
[total-axion-shares] Update total-axion-shares strategy

JTravaux:
# Fixes # 

- Change data_reader_address
- Add param to change it in the future

### DIFF
--- a/src/strategies/total-axion-shares/examples.json
+++ b/src/strategies/total-axion-shares/examples.json
@@ -4,7 +4,8 @@
     "strategy": {
       "name": "total-axion-shares",
       "params": {
-        "symbol": "TAS"
+        "symbol": "TAS",
+        "dataReader": "0x8458fe26CbF3B8295755654C02ABaDFB1d3CBCe6"
       }
     },
     "network": "1",
--- a/src/strategies/total-axion-shares/examples.json
+++ b/src/strategies/total-axion-shares/examples.json
@@ -8,12 +8,12 @@
         "dataReader": "0x8458fe26CbF3B8295755654C02ABaDFB1d3CBCe6"
       }
     },
-    "network": "1",
+    "network": "137",
     "addresses": [
       "0xe8b283b606a212d82036f74f88177375125440f6",
       "0x01E5fEafA5133bA6cdcc7b9afF7Aa5A3bE3f772F",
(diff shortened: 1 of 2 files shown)


src/strategies/total-axion-shares/index.ts
```diff
@@ -4,7 +4,7 @@ import { BigNumber } from '@ethersproject/bignumber';
 export const author = 'Axion Foundation';
 export const version = '0.2.0';
 
-const data_reader_address = '0x421456eFcEBf814975c8739CD613e5e7a954C474';
+const data_reader_address = '0x8458fe26CbF3B8295755654C02ABaDFB1d3CBCe6';
```
ChaituVR, line 7: https://etherscan.io/address/0x8458fe26CbF3B8295755654C02ABaDFB1d3CBCe6

Looks like the address is not a contract, maybe keep the older address and just make options.dataReader optional

ChaituVR (changes requested):

src/strategies/total-axion-shares/index.ts
JTravaux: Axion has migrated to Polygon, so we would switch the space to Polygon as well. This data reader exists on Polygon.

src/strategies/total-axion-shares/examples.json
```diff
@@ -4,7 +4,8 @@
     "strategy": {
       "name": "total-axion-shares",
       "params": {
-        "symbol": "TAS"
+        "symbol": "TAS",
+        "dataReader": "0x8458fe26CbF3B8295755654C02ABaDFB1d3CBCe6"
       }
     },
     "network": "1",
```
ChaituVR, line 11: In that case can you change the network here plz :)
also make sure to run 
```
yarn test --strategy=total-axion-shares
```
and make sure it pass all the test cases

ChaituVR (changes requested):

src/strategies/total-axion-shares/examples.json
ChaituVR: Or else better to not have any changes in examples.json, so it can pass based on previous values

JTravaux:
Updated to work with Polygon, and made sure tests pass.

image

bonustrack (approved):

ChaituVR (approved):

PR #314, opened 2022-01-20
[digitalax-deco-to-mona] new strategy

onigiri-x:
This is a new strategy for Digitalax

What it does is queries your balance of our DECO token

Converts to ETH value

Converts to MONA value

For our MONA based snapshot voting

Uses a certain subgraph for quickswap to get the pair conversion prices

Cheers

### DIFF
--- a/src/strategies/digitalax-deco-to-mona/index.ts
+++ b/src/strategies/digitalax-deco-to-mona/index.ts
@@ -1,11 +1,12 @@
-import { subgraphRequest } from '../../utils';
+import { multicall } from '../../utils';
 import { strategy as erc20BalanceOfStrategy } from '../erc20-balance-of';
 
 export const author = 'onigiri-x';
 export const version = '0.1.0';
 
-const QUICKSWAP_SUBGRAPH =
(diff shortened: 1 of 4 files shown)


src/strategies/digitalax-deco-to-mona/index.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  // Mona to eth
+  const holderParams2 = {
+    pair: {
+      __args: {
+        id: monaEthPair
+      },
+      token0Price: true
+    }
+  };
```
ChaituVR, line 40: Hi @onigiri-x maybe these prices could be changing in the future or during the voting period? Voting power of users could be changing with time? is there any way to query this with time/block? 

Also i think we can replace these with single query `pairs`
```
{
  pairs(where:{id_in:["0x7ecb3be21714114d912469810aedd34e6fc27736","0x3203bf44d434452b4605c7657c51bfeaf2a0847c"] 

  }){
    token1Price
    token0Price
  }
}
```

ChaituVR (changes requested):

bonustrack (approved):

ChaituVR (approved):

PR #317, opened 2022-01-21
[giveth-xdai-balance] added two new strategies for GIveth

pkretzschmar:
Added two new streategies:

- giveth-xdai-balance
- giveth-balancer-balance


src/strategies/giveth-xdai-balance/index.ts
```diff
@@ -0,0 +1,106 @@
...
+  return amountLP.mul(givBalance).div(totalLP);
+};
+
+export async function strategy(
+  space,
+  network,
+  provider,
+  addresses,
+  options,
+  snapshot
```
ChaituVR, line 63: Hi @pkretzschmar you may have to request data from subgraph based on snapshot or its timestamp, else the voting power will keep changing in future

ChaituVR (changes requested):

src/strategies/giveth-xdai-balance/index.ts
pkretzschmar: Done!

src/strategies/giveth-balancer-balance/index.ts
```diff
@@ -0,0 +1,101 @@
...
+  }
+  const balData = await subgraphRequest(BALANCER_SUBGRAPH_API, poolParams);
+  const balFormatedData = formatReserveBalance(balData, options.decimals);
+
+  const data = await Promise.all(
+    addresses.map((address) => {
+      params.balance.__args.id = address.toLowerCase();
+      return subgraphRequest(GIVETH_SUBGRAPH_API, params);
+    })
+  );
```
ChaituVR, line 76: Hi @pkretzschmar Calling the subgraph per one address won't scale for proposals having large voters, Is there a way to request data for all addresses at once?

ChaituVR (changes requested):

src/strategies/giveth-balancer-balance/index.ts
pkretzschmar: I found a way, just pushed it!

src/strategies/giveth-balancer-balance/index.ts
```diff
@@ -0,0 +1,103 @@
...
+      balancerLpStaked
+    } = addressBalance;
+    const totalGIV = BigNumber.from(balance).add(givStaked);
+
+    const balGIV = calcGivAmount(
+      BigNumber.from(balancerLp).add(balancerLpStaked),
+      balFormatedData.totalShares,
+      balFormatedData.balance
+    );
+    score[id] = parseFloat(formatUnits(totalGIV.add(balGIV), options.decimals));
```
ChaituVR, line 99: Looks good @pkretzschmar Just one small fix needed, looks like all the addresses returned from subgraph are lowercase, we need to return checksum addresses, you can use `getAdress` for that, example
https://github.com/snapshot-labs/snapshot-strategies/blob/master/src/strategies/synthetix-quadratic/index.ts#L124

We need to do same thing for `giveth-xdai-balance` strategy too

ChaituVR (changes requested):

src/strategies/giveth-balancer-balance/index.ts
pkretzschmar: Done! Thanks for your feedbacks!

src/strategies/giveth-xdai-balance/index.ts
```diff
@@ -0,0 +1,141 @@
...
+
+  params.balances.__args.where.id_in = addresses.map((address) =>
+    address.toLowerCase()
+  );
+
+  const data = await subgraphRequest(GIVETH_SUBGRAPH_API, params);
+  const dataBalances = data.balances;
+
+  const score = {};
+  console.log(dataBalances);
```
ChaituVR, line 114:
```suggestion
```

ChaituVR:
image

Fail when passing latest snapshot block number

ChaituVR (approved):

PR #323, opened 2022-01-24
Add immutable-x strategy [immutable-x]

sergerad:
Fixes # .

- 
- 
-


src/strategies/immutable-x/index.ts
```diff
@@ -0,0 +1,127 @@
...
+    network: string,
+    provider: StaticJsonRpcProvider,
+    options: Options,
+    addresses: string[],
+    block: number | string
+): Promise<Record<string, number>> {
+    const multi = new Multicaller(network, provider, abi, { blockTag: block });
+    addresses.forEach((address: string) => multi.call(address, options.address, 'balanceOf', [address]));
+    const result: Record<string, BigNumberish> = await multi.execute();
+    return mapL1Response(result, options);
```
ChaituVR, line 64: @sergerad Hello! You can use `erc20-balance-of` strategy for this

ChaituVR (changes requested):
Looks like this strategy is just to get balance of from mainnet and ropston ??? i think you can try using multichain strategy for that 🤔 Let me know if this is something else

PR #328, opened 2022-01-25
[immutable-x] Add Immutable X strategy

sergerad:
- New strategy under `immutable-x`

### DIFF
--- a/src/strategies/immutable-x/index.ts
+++ b/src/strategies/immutable-x/index.ts
@@ -1,0 +1,106 @@
+import { BigNumber, BigNumberish } from '@ethersproject/bignumber';
+
+import { StaticJsonRpcProvider } from '@ethersproject/providers';
+import fetch from 'cross-fetch';
+import { formatUnits } from '@ethersproject/units';
+import { strategy as getL1Balances } from '../erc20-balance-of'
+
+export const author = 'immutable';
+export const version = '1.0.0';
+
+export const name = 'immutable-x';
+
+const snapshotPath = '/v1/snapshots/balances';
+
+const networkMapping = {
+    1: 'https://api.x.immutable.com',
+    3: 'https://api.ropsten.x.immutable.com'
+};
+
+const defaultPageSize = 100;
+
+interface Response {
+    records: Score[];
+    cursor: string;
+}
+
+interface Score {
+    ether_key: string;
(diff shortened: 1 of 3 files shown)


src/strategies/immutable-x/index.ts
```diff
@@ -0,0 +1,127 @@
...
+    addresses: string[],
+    block: number | string
+): Promise<Record<string, number>> {
+    const multi = new Multicaller(network, provider, abi, { blockTag: block });
+    addresses.forEach((address: string) => multi.call(address, options.address, 'balanceOf', [address]));
+    const result: Record<string, BigNumberish> = await multi.execute();
+    return mapL1Response(result, options);
+}
+
+async function getL2Balances(network: string, options: Options, addresses: string[], block: number | string): Promise<Record<string, number>> {
```
sergerad, line 67: @ChaituVR This strategy adds L2 and L1 balances. Did you mean I should leverage another strategy within this one for L1 only?

ChaituVR: Hi @sergerad, No, I mean in this case you can use multichain strategy and pass two `erc20-balance-of` strategies into it 

More details here: https://snapshot.org/#/strategy/multichain
Play with it more and let me know if it multichain strategy doesn't work for you

ChaituVR (changes requested):

src/strategies/immutable-x/index.ts
bonustrack: The voting strategy "multichain" would only work if the chain is supported. If you want to get voting power from StarkEx for example I would recommend keep this approach.

sergerad: I have used `erc20-balance-of` strategy for L1 in this PR 👍

sergerad: @bonustrack PR is labelled with `changes requested` but my interpretation of your comment was to keep the current impl

ChaituVR: @sergerad Can you revert changes from `yarn.lock` file please, everything else looks good

sergerad: @ChaituVR done, thanks

ChaituVR (approved):

PR #324, opened 2022-01-24
[sandman-dao] Adds SandmanDAO strategy

sandmanfinance:
Adds SandmanDAO NFTs Strategy

Changes proposed in this pull request:
- It reads all "Endless NFTs" of a wallet
- It reads all "Experience" Skill of each Endless NFT.
- Create a weighted voting value according to that skill value.

PR #335, opened 2022-01-27
Adds Banksy Farm DAO strategy [banksy-dao]

BanksyFarm:
- Adds support for reading a value in NFT and adding it. (Same as Sandman Finance)

### DIFF
--- a/src/strategies/banksy-dao/index.ts
+++ b/src/strategies/banksy-dao/index.ts
@@ -75,7 +75,6 @@ export async function strategy(
   const results = {} as Record<string, number>;
   for (const [walletID, values] of Object.entries(walletIDToSkills)) {
     const address = walletID.split('-')[0];
-    console.log(values[1]);
     const currentExperience = values[1] / 1e18;
 
     let extraBoosted = 1;
--- a/src/strategies/banksy-dao/index.ts
+++ b/src/strategies/banksy-dao/index.ts
@@ -1,7 +1,7 @@
 import { Multicaller } from '../../utils';
 import { BigNumber } from '@ethersproject/bignumber';
 
-export const author = 'myst4';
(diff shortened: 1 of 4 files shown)


src/strategies/banksy-dao/index.ts
```diff
@@ -0,0 +1,95 @@
+import { Multicaller } from '../../utils';
+import { BigNumber } from '@ethersproject/bignumber';
+
+export const author = 'myst4';
```
ChaituVR, line 4:
```suggestion
export const author = 'BanksyFarm';
```

ChaituVR (changes requested):
Hi @jordanmessina Looks like this is same as https://github.com/snapshot-labs/snapshot-strategies/pull/324/files why not reuse it?

sandmanfinance:
Although very similar, the strategies are for two different NFT contracts.
If you check, the ABIs are different.
So they must be different strategies.

Thank you very much.

src/strategies/banksy-dao/index.ts
```diff
@@ -0,0 +1,95 @@
...
+  }
+  const walletIDToSkills: Record<
+    string,
+    any
+  > = await callTokenToSkill.execute();
+  
+  const results = {} as Record<string, number>;
+  for (const [walletID, values] of Object.entries(walletIDToSkills)) {
+    const address = walletID.split('-')[0];
+    console.log(values[1]);
```
ChaituVR, line 78:
```suggestion
```

ChaituVR (changes requested):

ChaituVR:
Yea , looks very similar except `getArtWorkOverView`

ChaituVR (approved):

PR #340, opened 2022-01-28
[stakers-and-holders] Add new strategy

tony31913:
- Add my new strategy - herobabies-club-stakers-and-owners


ChaituVR (changes requested):
Hi @tony31913 I see this is equal to `mutant-cats-stakers-and-holders` strategy,
Is it not possible to reuse it? 

Maybe this is because of the name? then I would suggest renaming this new strategy to something like `stakers-and-holders` so others can reuse it too.

let me know :)

tony31913:
Oh yes, it is because of the name. I've updated the name. Are you able to merge?

ChaituVR (approved):

PR #342, opened 2022-01-29
[ethercats-founder-series] added ethercats nft strategy

naomsa:


ChaituVR:
@naomsa#5548 Also there is a heap memory error on GitHub action and test cases fail

image

ChaituVR:
I think we should try the solution with subgraph  https://thegraph.com/hosted-service/subgraph/alexvorobiov/eip1155subgraph

woodydeck:
> I think we should try the solution with subgraph https://thegraph.com/hosted-service/subgraph/alexvorobiov/eip1155subgraph

This subgraph reports ownership incorrectly for the collection. This subgraph probably shouldn't be used for anything without a patch and full audit. It's not really an edge case either, it will affect all collections using Chainlink VRF I believe to start with.

Is there another subgraph we can validate against?

src/strategies/ethercats-founder-series/index.ts
```diff
@@ -0,0 +1,61 @@
+import { subgraphRequest } from '../../utils';
+
+export const author = 'naomsa';
```
ChaituVR, line 3:
```suggestion
export const author = 'woodydeck';
```

src/strategies/ethercats-founder-series/examples.json
```diff
@@ -0,0 +1,14 @@
+[
+  {
+    "name": "Example query",
+    "strategy": {
+      "name": "ethercats-founder-series",
+      "params": {
+        "symbol": "Ether Cat"
```
ChaituVR, line 7:
```suggestion
        "symbol": "EtherCat"
```

ChaituVR (approved):

PR #344, opened 2022-01-30
[multichain] Use blockfinder to get block nums

bonustrack:


ChaituVR:
Looks good, I will update readme before merging

bonustrack:
@ChaituVR Merging this, I'll update le README later after test in prod

PR #345, opened 2022-01-31
[landdao-token-tiers] weighted tier based strategy for nfts

ethantddavis:
Fixes # .

- add a strategy for the land dao which checks a token tier list for voting weights.


src/strategies/landdao-token-tiers/examples.json
```diff
@@ -0,0 +1,8914 @@
...
+          "8879": 2,
+          "8880": 2,
+          "8881": 2,
+          "8882": 3,
+          "8883": 3,
+          "8884": 1,
+          "8885": 1,
+          "8886": 1,
+          "8887": 1,
+          "8888": 1
```
ChaituVR: Hi @ethantddavis Adding tokens like this can potentially create scalability issues on the snapshot server, is it possible to fetch this list from somewhere? maybe from an IPFS file?

ChaituVR (changes requested):
Also this strategy is almost similar to `erc721-with-tokenid-range-weights` strategy , can you separate the common code and reuse it?

src/strategies/landdao-token-tiers/examples.json
ethantddavis: Thanks for the feedback @ChaituVR, I will set up an IPFS file to store the data. I see the similarity to tokenId-range-weights, but many token weights for our use case are not within a range, instead were assigned randomly. I am not sure how I would separate common code in this scenario.

PR #352, opened 2022-02-01
[razor-network-voting] Razor Network Voting Strategy

adi44:
A custom strategy for calculating the voting power of stakers and delegators of Razor Network. With this strategy, we will be calculating the total number of equivalent razors that they own for sRAZORs amount. The amount of razor is calculated using the specific formal which is (total stake of staker * Amount of sRAZORs) / total supply.


adi44:
@ChaituVR Could you please review our strategy and suggest if any changes needed

src/strategies/razor-network-voting/index.ts
```diff
@@ -0,0 +1,125 @@
...
+  if (snapshot !== 'latest') {
+    // @ts-ignore
+    params.delegators.__args.block = { number: snapshot };
+  }
+
+  const score = {};
+
+  // subgraph request 1 : it fetches all the details of the stakers and delegators.
+  const result = await subgraphRequest(RAZOR_NETWORK_SUBGRAPH_URL, params);
+  console.log(result);
```
ChaituVR, line 72:
```suggestion
```

src/strategies/razor-network-voting/index.ts
```diff
@@ -0,0 +1,125 @@
...
+      }) => {
+        const razor_amount = sRZR_to_RZR(
+          BigNumber.from(Staker.sAmount),
+          BigNumber.from(Staker.totalSupply),
+          BigNumber.from(Staker.stake)
+        );
+        //score will be based on the current stake in the block Number
+        if (!score[Staker.staker]) {
+          //if score[delegator] has no score setup already we will put it as intial amount
+          score[Staker.staker] = wei_to_ether(Number(razor_amount));
```
ChaituVR, line 114: The addresses from the subgraph are lowecased,  we need to return checksum addresses, you can use getAdress for that, example
https://github.com/snapshot-labs/snapshot-strategies/blob/master/src/strategies/synthetix-quadratic/index.ts#L124

ChaituVR (changes requested):

adi44:
@ChaituVR I have made changes as per comments. 
cc : @SkandaBhat

ChaituVR (approved):

PR #354, opened 2022-02-02
[revest] Created Revest Staking System strategy 

RobAnon:
A custom strategy proposed to offer support for snapshots taken through the Revest staking system. Calculates who has what representational power based on the size of their staked position into Revest and the length it is locked up for, all off of what FNFTs they possess and the voting power each FNFT represents on its own


src/strategies/revest/index.ts
```diff
@@ -0,0 +1,158 @@
...
+    );
+    rewards = rewards[0][0];
+    let allocLP = await multicall(
+      network,
+      _provider,
+      abi3,
+      [""].map((id: any) => [rewards, 'totalLPAllocPoint', []]),
+      {blockTag}
+    );
+    let allocToken = await multicall(
```
ChaituVR, line 126: Hi @RobAnon There is a limit on the number of requests being sent (multicall/subgraph/fetch), It is limited to 5 calls 

Can you try to find if there is some alternate way to get this in fewer calls? maybe try sending multiple calls into multicall

ChaituVR (changes requested):

src/strategies/revest/index.ts
RobAnon: Can I call multiple functions via multicall?

RobAnon: @ChaituVR I've reduced the number to 5 calls through the use of multicall

src/strategies/revest/index.ts
```diff
@@ -0,0 +1,153 @@
...
+    });
+    let returnVals = {};
+    Object.keys(finalResult).forEach(element => {
+      returnVals[element] = parseInt(finalResult[element].toString(), 10); 
+    });
+    console.log(returnVals);
+    return returnVals;
+  } catch (err) {
+    return {};
+  }
```
ChaituVR, line 152: Hi, It should still throw an error, `{}` may mean there is 0 voting power for all addresses, maybe remove the try-catch, should be good to go :)

ChaituVR (changes requested):

RobAnon:
@ChaituVR Ready to merge

src/strategies/revest/index.ts
```diff
@@ -0,0 +1,150 @@
...
+          finalResult[getAddress(account.id)] = score;
+        }
+      }
+    });
+  });
+  let returnVals = {};
+  Object.keys(finalResult).forEach(element => {
+    returnVals[element] = parseInt(finalResult[element].toString(), 10); 
+  });
+  console.log(returnVals);
```
ChaituVR, line 147:
```suggestion
```

ChaituVR (approved):

PR #357, opened 2022-02-02
[chubbykaijudao] ChubbyKaijuDAO staker-and-holder strategy

ekwelfweer:
A custom strategy for both granting staker-and-holder to use snapshot service. The most of codes are borrowed from "mutant-cats-stakers-and-holders" strategy. As this strategy is not the first of a kind, so it already proved its completeness. We revised a token/staking contract address and function name to be consistent with our contract.  

We omitted the example.json because we could not understand what do "address", "network", "snapshot" parameters mean.


ChaituVR:
Hi @ekwelfweer Answered in the issue, make sure to remove comments and try running this in your local
```
npm test --strategy=chubbykaijudao --more=500
```

src/strategies/chubbykaijudao/index.ts
```diff
@@ -0,0 +1,49 @@
...
+  const stakingPool = new Multicaller(network, provider, stakingAbi, {
+    blockTag
+  });
+  const tokenPool = new Multicaller(network, provider, tokenAbi, {
+    blockTag
+  });
+
+  addresses.forEach((address) => {
+    stakingPool.call(address, options.staking, 'GEN1depositsOf', [address]);
+    tokenPool.call(address, options.token, 'balanceOf', [address]);
```
ChaituVR, line 34: @ekwelfweer For `balanceOf` you can use `erc20-balance-of` strategy, maybe just keep `GEN1depositsOf` in this . so the voting power will be sum of these two strategies

ekwelfweer: hello @ChaituVR,

both tokens are erc-721. that is none of them are erc-20.

When people "stake" the ERC-721 NFT, the ownership of their nfts is transferred to the contract, and that's why we call GEN1depositsOf or balanceOf in both sides.

ekwelfweer:
hello @ChaituVR,

I removed the comments in readme.md, but I have trouble with testing.

I tried at the origin path "./snapshot-strategies" and this path "snapshot-strategies/src/strategies/chubbykaijudao" but both do not work well.

Below is the error I got. should I insert index.ts and example.json to the "./snapshot-strategies/test" folder? 

image

sorry for disturbing you, I'm a novice to this.

src/strategies/chubbykaijudao/index.ts
ChaituVR: oh you can use `erc721` strategy for that :) ( will call `balanceOf` function ) so i think would be great if this strategy just have `GEN1depositsOf`

ChaituVR:
@ekwelfweer You can run `yarn install` and then `yarn test --strategy=chubbykaijudao --more=500`

src/strategies/chubbykaijudao/index.ts
ChaituVR: I think we can leave it like this for now :) ignore my comment

src/strategies/chubbykaijudao/examples.json
```diff
@@ -0,0 +1,20 @@
+[
+  {
+    "name": "ChubbyKaijuDAO Stakers and Holders",
+    "strategy": {
+      "name": "chubbykaijudao-stakers-and-holders",
```
ChaituVR, line 5: Strategy name should be same as folder name @ekwelfweer Change the name here or change the folder name

ChaituVR (changes requested):

ChaituVR (changes requested):
Also make sure to add your strategy to https://github.com/curelycue/snapshot-strategies/blob/master/src/strategies/index.ts like other strategies

src/strategies/chubbykaijudao/examples.json
ekwelfweer: yep. i changed it all to "chubbykaijudao"

ekwelfweer:
image

passed. i'll create pr.

ekwelfweer:
there seems a conflict here. 
image

should i pull your repo and create pr again

ChaituVR (approved):

PR #358, opened 2022-02-02
Use typeof for latest block query param in snapshot strategy [immutable-x]

sergerad:
- Replace Immutable X strategy logic for adding query param `block` because test environment shows error when a string is used


ChaituVR:
@sergerad `block` can be `'latest'` too, there is no way to handle it in your API?

ChaituVR (approved):

PR #378, opened 2022-02-11
[razor-network-voting] Razor Network Voting

adi44:
Bug Fix for razor-network-voting strategy. There was a key mismatch in the scores object, if the voter was delegator and staker that was causing overriding of the score due to which for a few addresses the score was zero


adi44:
@ChaituVR there was some bug in our strategy that we have fixed now and created this PR. Could you Please review and merge it.

ChaituVR (approved):

PR #379, opened 2022-02-11
[govidao] Govidao

amibenson:
Fixes # .

- Adding govidao strategy
- 
-


ChaituVR:
Hi @amibenson I see this strategy is the same as `agive` strategy, why not reuse it? for erc20 balance, you can use `erc20-balance-of` with combination of `agive` strategy, which will result in same

amibenson:
> Hi @amibenson I see this strategy is the same as `agive` strategy, why not reuse it? for erc20 balance, you can use `erc20-balance-of` with combination of `agive` strategy, which will result in same

Thanks I see (`agave`)

PR #381, opened 2022-02-13
Update index.ts

illLoomiNate:
Would it be possible to implement a change similar to this idea? (perhaps as a separate strategy?)
Where instead of only having the option of having a range, it will have the tokenId and weight? 
Perhaps a range or tokenID built-in as an option? { length: (end - start) || tokenId / step + 1) an so on?

Question # .

- 
- 
-

### DIFF
--- a/src/strategies/erc721-with-tokenid-range-weights-simple/index.ts
+++ b/src/strategies/erc721-with-tokenid-range-weights-simple/index.ts
@@ -7,8 +7,8 @@ const abi = [
   'function ownerOf(uint256 tokenId) public view returns (address owner)'
 ];
 
-const range = (start, end, step) =>
-  Array.from({ length: (end - start) / step + 1 }, (_, i) => start + i * step);
+const range = ((uint256 tokenId), step) =>
+  Array.from( _tokenId / step + 1 }, (_, i) => start + i * step);
 
 export async function strategy(
   space,
@@ -27,7 +27,7 @@ export async function strategy(
 
   const responses: Array<any> = await Promise.all(
     tokenIdWeightRanges.map(async (tokenIdWeightRange) => {
-      const { start, end, weight } = tokenIdWeightRange;
+      const { tokenID, weight } = tokenIdWeightRange;
       if (end - start > 7000) {
         throw new Error('Strategy range too large');
       }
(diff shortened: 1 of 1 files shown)


ChaituVR (changes requested):
Hello @illLoomiNate you can use https://snapshot.org/#/strategy/erc721-with-tokenid-weighted for that :) let me know if it doesn't work for you

illLoomiNate:
Hello,
Thank you for the response!
Unfortunately, this does not provide the Strategy that I am looking to
utilize.

I need to be able to list individual Token ID's and each ID have a specific
weight instead of the strategy that you sent.

If you look at what I suggested in the request - would this be possible to
add?

Thank you once again in advance,
Nathan

ChaituVR:
@illLoomiNate Oh I see what you are referring to, I think it better to pass an optional param (`weights`) to this strategy https://snapshot.org/#/strategy/erc20-balance-of-weighted and map weights to tokens, is it possible to create a new PR?

illLoomiNate:
@ChaituVR This would work for the erc721 tokenid?

I would like to set ranges and then have the weight set to those ranges
Additionally weight that can be assigned to individual lists of erc721TokenIds as well 

Just as I mentioned in the first comment and in the changes suggested here as well. 

I think the changes I proposed should give a direct understanding of the main concept of the strategy
Thank you once again in advance!

illLoomiNate:
> @illLoomiNate Oh I see what you are referring to, I think it better to pass an optional param (`weights`) to this strategy https://snapshot.org/#/strategy/erc20-balance-of-weighted and map weights to tokens, is it possible to create a new PR?

Hello, Did you have an opportunity to check my latest message?

ChaituVR:
@illLoomiNate i didn't quite get what you are trying, you mean tokenIDs from 1 to tokenID ?

illLoomiNate:
image

@ChaituVR 

I just want to replace the range with tokenID as shown here - so we can have different tokenIDs that count as more votes

ChaituVR:
I don't quite understand it, Can you share an example with your contract and tokenIDs,, here for example your tokenID is 100, you want voting power from 1,100 ??

ChaituVR:
Or create a new PR with new strategy, with your contract and tokenIDs in examples.json. this strategy is being used by other spaces so can break that

ChaituVR:
Closing this as stale, feel free to reopen if needed

PR #384, opened 2022-02-14
Pr/342

woodydeck:
Updated the strategy to batch requests. Changed the strategy name to the correct spelling ethercats-founders-series.


ChaituVR:
So many changes in src/strategies/index.ts, please make sure changes are related to only this strategy
Also, leave the old strategy the same, add only new strategy

PR #403, opened 2022-02-17
Return score only for requested addresses

ChaituVR:


bonustrack:
What the issue we trying to fix here?

bonustrack (approved):

PR #415, opened 2022-02-23
[ethtrader-gov-weight] strategy for r/ethtrader dao governance

kohrts:
strategy fetches governance weights from a json that we manually update every 4 weeks during donut & contrib distributions to contributing DAO members on the r/ethtrader subreddit

### DIFF
--- a/src/strategies/ethtrader-gov-weight/index.ts
+++ b/src/strategies/ethtrader-gov-weight/index.ts
@@ -1,0 +1,25 @@
+import fetch from 'cross-fetch';
+
+export const author = 'kohrts';
+export const version = '0.1.0';
+
+export async function strategy(space, network, provider, addresses, options, snapshot) {
+  const ETHTRADER_USERS_URL = 'https://raw.githubusercontent.com/EthTrader/donut.distribution/main/docs/users.json'
+  
+  const response = await fetch(ETHTRADER_USERS_URL, {
+    method: 'GET',
+      headers: {
+      Accept: 'application/json',
+      'Content-Type': 'application/json'
+      }
+  });
+
+  const data = await response.json();
+
+  return Object.fromEntries(
+    data.map((value) => [
+      value.address,
+      parseFloat(value.weight.toString())
+    ])
+  );
+}
--- a/src/strategies/index.ts
+++ b/src/strategies/index.ts
@@ -529,7 +529,7 @@ const strategies = {
   starsharks,
   'printer-financial': printerFinancial,
   'ethercats-founders-series': ethercatsFoundersSeries,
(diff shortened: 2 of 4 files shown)


src/strategies/index.ts
```diff
@@ -528,6 +529,7 @@ const strategies = {
   starsharks,
   'printer-financial': printerFinancial,
   'ethercats-founders-series': ethercatsFoundersSeries,
+  'ethtrader-gov-weight': ethtraderGovWeight
```
ChaituVR, line 532:
```suggestion
  'ethtrader-gov-weight': ethtraderGovWeight,
```

src/strategies/ethtrader-gov-weight/index.ts
```diff
@@ -0,0 +1,25 @@
+import fetch from 'cross-fetch';
+
+export const author = 'kohrts';
+export const version = '0.1.0';
+
+export async function strategy(space, network, provider, addresses, options, snapshot) {
+  const ETHTRADER_USERS_URL = 'https://raw.githubusercontent.com/EthTrader/donut.distribution/main/docs/users.json'
```
ChaituVR, line 7: Hi, @kohrts  like this owner of this repo can change the data any time and won't give us consistent results, We need this data from on-chain / IPFS so it won't change in the future, Can you think of any alternate way to get data based on block and network?

ChaituVR (changes requested):

src/strategies/ethtrader-gov-weight/index.ts
kohrts: Yes, all this data does originate on-chain and IPFS, but it requires more than 5 calls to compile it all.

bonustrack: Is it 5 calls in total or 5 calls per address? If it's in total that should work within the strategy. If you use external service to get voting power you need to take the "snapshot" in consideration and store voting power history in a way that it always answer with same power based on the snapshot

src/strategies/ethtrader-gov-weight/index.ts
```diff
@@ -0,0 +1,25 @@
...
+  
+  const response = await fetch(ETHTRADER_USERS_URL, {
```
bonustrack, line 9: "snapshot" should be used here, you don't just need voting power, you need voting power at a specific snapshot / time

src/strategies/ethtrader-gov-weight/index.ts
kohrts: thanks, i get it! that makes a lot of sense. i'll make those updates

PR #431, opened 2022-03-03
Strategy symbol optional

ChaituVR:
Fixes https://github.com/snapshot-labs/snapshot/issues/1020


src/strategies/erc721/schema.json
```diff
@@ -10,7 +10,6 @@
           "type": "string",
           "title": "Symbol",
           "examples": ["e.g. DOODLE"],
-          "minLength": 1,
```
samuveth, line 13: Shouldn't symbol also be removed inside required?
image

samuveth:
@bonustrack should we make symbol optional on all strategies?

src/strategies/erc721/schema.json
ChaituVR: Removed 👍

samuveth:
@ChaituVR I don't think you need to remove the minLength if you remove the required.

ChaituVR:
@samuveth If we leave minLength and If we pass an empty string it throws an error,
image


But why do we need minLength now anyway? any problem by removing it? 😅

bonustrack:
> @bonustrack should we make symbol optional on all strategies?

@samuveth On all strategies yes

bonustrack:
@ChaituVR Was with with removing symbol in "required"?

samuveth:
> @samuveth If we leave minLength and If we pass an empty string it throws an error, image
> 
> But why do we need minLength now anyway? any problem by removing it? 😅

Not sure how we are handling this one the backend but if we don't have a minlength the symbol could be an empty string.

samuveth:
> > But why do we need minLength now anyway? any problem by removing it? 😅
> 
> Not sure how we are handling this one the backend but if we don't have a minlength the symbol could be an empty string.

samuveth:
> @samuveth On all strategies yes

So yeah @ChaituVR all strategies need this change.

ChaituVR:
> On all strategies yes

@bonustrack Does it also mean in the future, anyone wants to add a strategy with "symbol" required, we should change it to optional? I mean is it always optional regardless of strategy/ its owner?

bonustrack:
@ChaituVR Yes, ideally we would move at some point the symbol param into the parent object like network example:
```json
{
  "name": "ticket",
  "symbol": "ABC",
  "network": "1",
  "params": {}
}
```
This way we wont need to force ourself some schema for strategy params

ChaituVR:
Awesome @bonustrack we should do that soon 👍 made it optional in all strategies and added a test case so users will not make it required for future strategies

samuveth:
> Awesome @bonustrack we should do that soon 👍 made it optional in all strategies and added a test case so users will not make it required for future strategies

Would be good to maybe create an issue for it

samuveth (approved):

PR #432, opened 2022-03-03
[ocean-marketplace] Add other networks for ocean-marketplace strategy

idiom-bytes:
Updating ocean subgraph routes for BSC & Poly network.
- Enables us to expand marketplace strategy support for BSC & Poly


ChaituVR:
Hi @idiom-bytes Can you remove these functions in the next PR https://github.com/snapshot-labs/snapshot-strategies/blob/master/src/strategies/ocean-marketplace/oceanUtils.ts There are just for logs right?

ChaituVR (approved):

PR #448, opened 2022-03-08
Adding ERC-20 votes with override strategy.

serenae-fansubs:
This strategy is similar to [ERC-20 Votes](https://github.com/snapshot-labs/snapshot-strategies/tree/master/src/strategies/erc20-votes), except that it also allows individual delegators to **override** their vote on a particular proposal if they wish. This is most useful for social (off-chain only) proposals.

If an address has any delegated votes, then that amount will be used, minus the balances of any individual delegators who have also voted.

If an address has 0 delegated votes, then the current token balance will be used. The address must be delegated to another valid address, otherwise 0 will be returned.

### DIFF
--- a/src/strategies/erc20-votes-with-override/index.ts
+++ b/src/strategies/erc20-votes-with-override/index.ts
@@ -1,0 +1,140 @@
+import { formatUnits } from '@ethersproject/units';
+import { isAddress } from '@ethersproject/address';
+import { multicall } from '../../utils';
+
+export const author = 'serenae-fansubs';
+export const version = '0.1.0';
+
+const getVotesName = "getVotes";
+const getVotesABI = ["function getVotes(address account) view returns (uint256)"];
+const balanceOfName = "balanceOf";
+const balanceOfABI = ["function balanceOf(address account) external view returns (uint256)"];
+const delegatesName = "delegates";
+const delegatesABI = ["function delegates(address account) external view returns (address)"];
+
+/*
+  Counts votes from delegates, and also from individual delegators who wish
+  to override the vote of their delegate.
+
+  Makes three multicalls for votes, delegates, and balances.
+
+  If getVotes returns a nonzero value, then that is used. Then any delegators
+  that have individually voted will have their balances subtracted from the
+  result.
+
+  If getVotes returns zero, then it will return the token balance, provided
+  that the address is delegated to a valid address.
+
+  Otherwise, returns zero.
+
+  The function names/ABI can be overridden in the options.
+*/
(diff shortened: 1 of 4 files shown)


ChaituVR:
```js

/home/runner/work/snapshot-strategies/snapshot-strategies/src/strategies/erc20-votes-with-override/index.ts
  132:15  error  Don't use `Object` as a type. The `Object` type actually means "any non-nullish value", so it is marginally better than `unknown`.
- If you want a type meaning "any object", you probably want `Record<string, unknown>` instead.
- If you want a type meaning "any value", you probably want `unknown` instead  @typescript-eslint/ban-types
  133:14  error  Don't use `Object` as a type. The `Object` type actually means "any non-nullish value", so it is marginally better than `unknown`.
- If you want a type meaning "any object", you probably want `Record<string, unknown>` instead.
- If you want a type meaning "any value", you probably want `unknown` instead  @typescript-eslint/ban-types
  134:13  error  Don't use `Object` as a type. The `Object` type actually means "any non-nullish value", so it is marginally better than `unknown`.
- If you want a type meaning "any object", you probably want `Record<string, unknown>` instead.
- If you want a type meaning "any value", you probably want `unknown` instead  @typescript-eslint/ban-types
```

serenae-fansubs:
@ChaituVR Thank you! Addressed those comments about Object/Record, all checks are passing now :)

src/strategies/erc20-votes-with-override/index.ts
```diff
@@ -0,0 +1,144 @@
...
+/*
+  Counts votes from delegates, and also from individual delegators who wish
+  to override the vote of their delegate.
+
+  Makes three multicalls for votes, delegates, and balances.
+
+  If getVotes returns a nonzero value, then that is used. Then any delegators
+  that have individually voted will have their balances subtracted from the
+  result.
+
```
ChaituVR, line 24: How do the `getVotes` know that users have already voted? Sorry @serenae-fansubs Trying to get my brain around it :)

serenae-fansubs: @ChaituVR it's the same method used in the regular erc20-votes strategy: https://github.com/snapshot-labs/snapshot-strategies/blob/master/src/strategies/erc20-votes/index.ts#L24

serenae-fansubs: I also updated the comments there to make it more clear that it's talking about _voting power_, and refactored to make it a bit easier to follow.

ChaituVR (approved):

PR #456, opened 2022-03-10
[ve-ribbon-voting-power] verbn voting

chuddster:
Fixes # .

- return effective voting balance of veRBN lockers
- 
-


src/strategies/ve-ribbon-voting-power/index.ts
```diff
@@ -0,0 +1,33 @@
+import { BigNumberish } from '@ethersproject/bignumber';
+import { formatUnits } from '@ethersproject/units';
+import { Multicaller, getBlockNumber } from '../../utils';
+
+export const author = 'chuddster';
+export const version = '0.1.1';
```
ChaituVR, line 6:
```suggestion
export const version = '0.1.0';
```

ChaituVR (approved):

PR #459, opened 2022-03-14
[conv-finance] Adding new strategy for LP token, staking pool and reward calculation

eric-convergence:
Adding new strategy for Convergence Finance to calculating voting power based on LP token worth of platform token, Staking Pool balance and rewards


src/strategies/index.ts
```diff
@@ -544,7 +545,11 @@ const strategies = {
   potion,
   'safety-module-bpt-power': safetyModuleBptPower,
   'minotaur-money': MinotaurMoney,
+<<<<<<< HEAD
+  'conv-finance': convFinance
+=======
   'sd-boost': sdBoost
+>>>>>>> bae574f0d0e4f06b513f7cf251c9b1531d4ed4f6
```
ChaituVR, line 552: Hi @eric-convergence please resolve this

ChaituVR (changes requested):

src/strategies/index.ts
eric-convergence: Didn't notice this conflict, thanks.

src/strategies/conv-finance/index.ts
```diff
@@ -0,0 +1,343 @@
...
+        );
+      });
+    }
+  });
+
+  let res = await Promise.all([
+    ...lpMultiCalls,
+    ...stakingPoolsMultiCalls,
+    ...stakingPoolRewarderMultiCalls
+  ]);
```
ChaituVR, line 209: Hi @eric-convergence Can you add a limit for `options.rewarder.poolIds` and `options.stakingPools` like this it may send many requests if the length of these is more , there is a limit of 5 requests from a strategy https://github.com/snapshot-labs/snapshot-strategies#checklist-for-adding-a-new-strategy

ChaituVR (changes requested):

src/strategies/conv-finance/index.ts
eric-convergence: Tried my best to add limit in the new commit

src/strategies/conv-finance/index.ts
```diff
@@ -0,0 +1,347 @@
...
+        });
+      }
+    }
+  });
+
+  let res = await Promise.all([
+    ...lpMultiCalls,
+    ...stakingPoolsMultiCalls,
+    ...stakingPoolRewarderMultiCalls
+  ]);
```
ChaituVR, line 213: Hello @eric-convergence This definitely crossing number of requests sent from a strategy, there is a limit of 5 requests you can send from a strategy, https://github.com/snapshot-labs/snapshot-strategies#checklist-for-adding-a-new-strategy


Can you try if there is any alternate way for this, thanks :)

ChaituVR (changes requested):

src/strategies/conv-finance/index.ts
eric-convergence: I have merge all contract calls into 3 multi-call request, and it should be independent of the pool and rewarder size.

ChaituVR (approved):

Issue #449, opened 2022-03-08
Create strategy for OpenZeppelin ERC20Votes with individual override option


serenae-fansubs:
This strategy is similar to [ERC-20 Votes](https://github.com/snapshot-labs/snapshot-strategies/tree/master/src/strategies/erc20-votes), except that it also allows individual delegators to override their vote on a particular proposal if they wish. This is most useful for social (off-chain only) proposals.

If an account has any delegated voting power returned from getVotes, adds that value, minus the balances from any delegators that have also individually voted.

If an account is delegating to itself, then its own token balance will already be included in the getVotes return value.

If an account is delegating to a different valid address, adds the local token balance. The account must be delegated to another valid address, otherwise the local token balance will not be added.

Primarily for token contracts using OpenZeppelin's ERC20Votes: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/extensions/ERC20Votes.sol

However the getVotes method can also be overridden in the options too if need be.

PR #463, opened 2022-03-15
[erc20-votes-with-override] Adding compatibility for Snapshot delegations.

serenae-fansubs:
The standard "delegation" strategy is not compatible with the overridden voting power calculation, because it can lead to double-counting.

So instead, this strategy adds an "includeSnapshotDelegations" option. When enabled, one additional request is made to the subgraph to get Snapshot delegations. Then the unique delegators are added to the list of addresses and the normal overridden voting power calculation is used as before.

The "isSnapshotDelegatedScore" option determines whether the delegated or non-delegated score is used. So two strategies can be used, one with this option enabled and one with this option disabled.

#449

### DIFF
--- a/src/strategies/erc20-votes-with-override/README.md
+++ b/src/strategies/erc20-votes-with-override/README.md
@@ -8,11 +8,42 @@ If an account is delegating to itself, then its own token balance will already b
 
 If an account is delegating to a different valid address, adds the local token balance. The account must be delegated to another valid address, otherwise the local token balance will not be added.
 
+If the `includeSnapshotDelegations` option is enabled, then Snapshot delegations will also be taken into account. In this case, the `isSnapshotDelegatedScore` option will determine whether the delegated or non-delegated scores will be returned. This is done because the overridden voting power calculation is not compatible with the standard [delegation](../delegation) strategy. So instead, space admins can use this strategy **twice**, each with `includeSnapshotDelegations` enabled and the `isSnapshotDelegatedScore` enabled or disabled.
+
+Here is an example of enabling Snapshot delegations using this strategy twice:
+
+Delegated strategy:
+
+```
+{
+  "symbol": "ENS (delegated)",
+  "address": "0xC18360217D8F7Ab5e7c516566761Ea12Ce7F9D72",
+  "decimals": 18,
+  "includeSnapshotDelegations": true,
+  "isSnapshotDelegatedScore": true
+}
+```
+
+Non-delegated strategy:
+
+```
+{
+  "symbol": "ENS",
+  "address": "0xC18360217D8F7Ab5e7c516566761Ea12Ce7F9D72",
+  "decimals": 18,
+  "includeSnapshotDelegations": true,
+  "isSnapshotDelegatedScore": false
+}
+```
+
 ## Options
 
 - **address:** The address of the ERC-20 token contract.
 - **symbol:** The display symbol for the token, e.g. "ENS".
 - **decimals:** Used to display the correct base units for the token.
+- **includeSnapshotDelegations:** Optional. If enabled, Snapshot delegations are taken into account. See description above.
+- **isSnapshotDelegatedScore:** Optional. Only used if `includeSnapshotDelegations` is enabled. If true, the delegated score will be returned, otherwise the non-delegated score will be returned. See description above.
+- **delegationSpace:** Optional. Only used if `includeSnapshotDelegations` is enabled. Determines what specific Snapshot space to retrieve delegations for.
 - **getVotesName:** Optional. The function name of the [getVotes](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/extensions/ERC20Votes.sol#L64) function, e.g. "getVotes".
 - **getVotesABI:** Optional. The ABI specification for the getVotes function.
 - **balanceOfName:** Optional. The function name of the [balanceOf](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/IERC20.sol#L18) function, e.g. "balanceOf".
--- a/src/strategies/erc20-votes-with-override/README.md
+++ b/src/strategies/erc20-votes-with-override/README.md
@@ -1,5 +1,13 @@
 # ERC-20 Votes with Override
 
+- [Overview](#overview)
+- [Example](#example)
+- [Snapshot Delegations](#snapshot-delegations)
+- [Example With Snapshot Delegations](#example-with-snapshot-delegations)
+- [Options](#options)
+
+### Overview
+
 This strategy is similar to [ERC-20 Votes](../erc20-votes), except that it also allows individual delegators to **override** their vote on a particular proposal if they wish. This is most useful for social (off-chain only) proposals.
 
 If an account has any delegated voting power returned from getVotes, adds that value, minus the balances from any delegators that have also individually voted.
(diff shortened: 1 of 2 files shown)


serenae-fansubs:
@ChaituVR I was testing out our new strategy and noticed that it wasn't compatible with the [delegation](https://github.com/snapshot-labs/snapshot-strategies/tree/master/src/strategies/delegation) strategy because of how the overridden voting power is calculated. Some double-counting can happen in some scenarios.

So I incorporated the Snapshot delegations into this strategy, meaning that (if that option is enabled) 1 additional request is made, to get delegations from the Snapshot subgraph.

Happy to answer any questions!

ChaituVR:
> I was testing out our new strategy and noticed that it wasn't compatible with the [delegation](https://github.com/snapshot-labs/snapshot-strategies/tree/master/src/strategies/delegation) strategy because of how the overridden voting power is calculated. Some double-counting can happen in some scenarios.

Hey @serenae-fansubs Did you try delegate strategy on snapshot? It should specifically designed to avoud double counting :) Should work in your case, let me know if you see some double calculation, this is how strategy works:

> This is how delegation works:
Delegation strategy is used get delegated voting power of a address (we can use any strategies to calculate that voting power), but if delegator votes on the proposal, the other person won't get his voting power
For example A have 10 tokens, B got 10
A delegates their voting power to B from https://snapshot.org/#/delegate, there are two cases to calculate Voting power:

> If Space is using delegation strategy (with erc20-balance-of passed into params)
B's voting power will be 10 (A delegated voting power + if any other address's delegated voting power)
A can't vote in this case because no one delegate to him
If Space is using delegation strategy (with erc20-balance-of passed into params) + erc20-balance-of strategy
B's voting power will be 20 (His own tokens + A delegated voting power + if any other address's delegated voting power)
A can vote now because they got 10 tokens and If they vote, B's voting power will be reduced to 10 (His own tokens + if any other address's delegated voting power) - He won't have access to A's voting power.

serenae-fansubs:
@ChaituVR Yep I did try, and unfortunately the delegation strategy is incompatible with this strategy.

The reason why is that the delegation strategy retrieves all delegator accounts that weren't in the original address list, and sends **only that list of delegators** to the underlying strategy to get direct scores.

However, in this case the underlying strategy will only work properly if the full original list of addresses are also included. This is because it needs to know which other accounts have also voted so that it can correctly subtract the balances.

Example:

- A and B both have 100 tokens
- A delegates to B on-chain
- A also delegates to B via Snapshot

If **only B votes**, then the underlying strategy will correctly calculate 200 for B. However when the delegation strategy is also used, then it will pass A into the underlying strategy, and **because it allows overrides for delegators**, the underlying strategy will return 100 for A. Now the total voting power B has is 300 (200 + 100 delegated) which is incorrect.

serenae-fansubs:
I have a test proposal up on my space to demonstrate the double counting: https://snapshot.org/#/serenae.eth/proposal/0x66eec9115308230110b07a74775628af0bf557a5c137f6b98b2338f1d27ed351

Account A (70DC) has ~17 tokens and is delegating on-chain to Account C (DE70)
Account B (F013) has ~202 tokens and is delegation on-chain to Account C (DE70)

Account A (70DC) is delegating on Snapshot to Account C (DE70)

serenae-fansubs:
@ChaituVR I have a test proposal up on my space to demonstrate the double counting: https://snapshot.org/#/serenae.eth/proposal/0x66eec9115308230110b07a74775628af0bf557a5c137f6b98b2338f1d27ed351

Account A (70DC) has ~17 tokens and is delegating on-chain to Account C (DE70)
Account B (F013) has ~202 tokens and is delegation on-chain to Account C (DE70)

Account A (70DC) is delegating on Snapshot to Account C (DE70)
Account B (F013) is delegating on Snapshot to Account C (DE70)

When accounts A and C both vote, the 17 tokens from A are double-counted.
Account A has a score of 17 (+0 delegated)
Account C has a score of 2.9k (+219 delegated)
Account C _should_ only have 202 delegated, but it actually has 219 delegated. The 17 tokens from account A are double-counted in this case.

serenae-fansubs:
@ChaituVR Maybe this visual diagram will help here:

image

Here's the on-chain voting power with these delegations (again, using the standard [OpenZeppelin ERC20Votes](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/extensions/ERC20Votes.sol) system)
| A | B | C | 
| ------------- | ------------- | ------------- |
| 0 | 100 | 500 | 

With the regular [erc20-votes strategy](https://github.com/snapshot-labs/snapshot-strategies/tree/master/src/strategies/erc20-votes), each account would only have access to its own on-chain delegated voting power, and delegators without voting power will not be able to vote on Snapshot proposals.

However, with this new [erc20-votes-with-override strategy](https://github.com/snapshot-labs/snapshot-strategies/tree/master/src/strategies/erc20-votes-with-override), individual delegators will be able to vote on Snapshot proposals, without needing to perform any additional on-chain transactions. Here are the scores that would be given using this override strategy, depending on who votes:
| Voters  | Score A | Score B | Score C | 
| ------------- | ------------- | ------------- | ------------- |
| A | 100 | | |
| B | | 300 | |
| C | | | 500 |
| A,B | 100 | 200 | |
| A,C | 100 | | 500 |
| B,C | | 300 | 300 |
| A,B,C | 100 | 200 | 300 |

The override strategy takes the voting power into account, subtracts any balances from delegators that are also voting, and also adds the local token balance if the account is delegating to another account. That's how it ensures that delegators can "override" their delegates but still no double-counting happens.

Now, what about if the delegation strategy is used? Here are the Snapshot delegated scores that occur:
| Voters  | Score A | Score B | Score C | 
| ------------- | ------------- | ------------- | ------------- |
| A | 0 | | |
| B | | 0 | |
| C | | | 300 |
| A,B | 0 | 0 | |
| A,C | 0 | | 300 |
| B,C | | 0 | 100 |
| A,B,C | 0 | 0 | 0 |

So the total scores, if both the regular and delegated strategies are used together, are:
| Voters  | Score A | Score B | Score C | 
| ------------- | ------------- | ------------- | ------------- |
| A | 100 | | |
| B | | 300 | |
| C | | | 800 |
| A,B | 100 | 200 | |
| A,C | 100 | | 800 |
| B,C | | 300 | 400 |
| A,B,C | 100 | 200 | 300 |

As you can see, in multiple scenarios, double-counting occurs when using the regular strategy and delegation strategy together.

Take the example where just C votes. The regular strategy just returns the on-chain delegated value of 500, and nothing else, because nobody else voted.

However, the delegation strategy retrieves the delegators [A,B], and passes them into the underlying strategy. From the underlying strategy's point of view, it's as if [A,B] is voting and nobody else. So it returns the values [100,200], and the delegation strategy reduces that and calculates the delegated voting power as 300. Now the 200 is double-counted, and the total scores add up to 800 which is incorrect.

There is no way for the current delegation strategy to work with this strategy, because the delegation strategy only ever passes the list of delegators that were not in the original address list, so the underlying strategy has no idea that other addresses voted as well.

My changes in this PR address this by retrieving the Snapshot delegations directly, and then calculating the scores as if all those delegators were also voting. So with my changes here, if only C votes, then:
- My strategy retrieves Snapshot delegators [A,B]
- [A,B] are added to the total address list, so it becomes [A,B,C]
- The regular calculations are done, so: {A: 100,  B: 200,  C: 300}
- If `isSnapshotDelegatedScore` is false, only the score for C is returned: {C: 300}
- If `isSnapshotDelegatedScore` is true, the scores for all delegators of C are summed up, so {A:100,B:200} is summed and the strategy returns {C: 300}
- The final score is then C: 300 + 300 delegated, which is correct

ChaituVR:
Hi @serenae-fansubs Can you add this info to Readme too, will be helpfull to everyone who want to use this strategy in future

ChaituVR:
Thanks @serenae-fansubs :)

ChaituVR (approved):

PR #474, opened 2022-03-18
[saffron-finance-v2] Use correct SFI staked amount from each staking pool

turpintinz:
Fixes error with counting SFI staked amounts within Saffron Finance V2 staking pools

* Add SFI total to a user's totlal correctly
* Update author field
* Update version
* Use singleAssets where appropriate
* Use conditional to add userInfo.amount to account's voting power

7e7ccd


ChaituVR:
Hi @turpintinz There is an active proposal here https://snapshot.org/#/saffronfinance.eth and these changes can affect it, maybe it is better to create a `-v3` strategy instead so that it will not effect anything else

turpintinz:
Thank you, @ChaituVR . Actually, that's okay if this PR impacts the proposal that closed today. The change fixes a counting bug that was discovered with that proposal. We would like the fix to go into saffron-finance-v2 'cuz the "V2" refers to the protocol's version number rather than the voting strategy. If this meets all of your check, it would be helpful to have it merged in.

ChaituVR (approved):

PR #485, opened 2022-03-22
Double delegation limit to 10000

oo-00:
(Temporarily) fixes delegation limit, which is imposed by a max skip of 5000 on graphql.

After getting 6 pages in asc order, retrieves up to 6 pages in desc order and breaks at a collision.


ChaituVR:
Refer https://github.com/snapshot-labs/snapshot-strategies/issues/494

PR #486, opened 2022-03-23
[math] Add a new math method: multiply.

Skyge:


src/strategies/math/options.ts
```diff
@@ -3,6 +3,7 @@ export type Operand = StrategyOperand | ConstantOperand;
 export interface Options {
   operands: Operand[];
   operation: Operation;
+  multiplier: number;
```
ChaituVR, line 6: Hi @Skyge It should be like other types, you can pass multiplier into operands like other operations

ChaituVR (changes requested):
Also would be great if you can change Readme as well

src/strategies/math/options.ts
Skyge: Okay, let me update it.

Skyge: So do you mean like this:
```
"operands": [
          {
            "type": "strategy",
            "strategy": {
              "network": "1",
              "name": "erc20-balance-of",
              "params": {
                "address": "0x6b175474e89094c44da98b954eedeac495271d0f",
                "symbol": "DAI",
                "decimals": 18
              }
            },
            "multiplier": 6
          }
        ],
```

Skyge:
It seems like there is something wrong.

ChaituVR (approved):

PR #480, opened 2022-03-21
Add cake pools

chef-jojo:
Changes proposed in this pull request:

Add `cake-pools` strategy for smart chef pools. 
It uses for displaying total Syrup Pools scores and can be consumed by `cake` strategy later on.
The params `chefAddresses` should be reflect to active syrup pools base on snapshot number

PR #489, opened 2022-03-23
[cake] fix cake

chef-jojo:
-  while trying to move vp to subgraph, here to fix at least make cake works


ChaituVR:
Hi @chef-jojo these changes can affect the voting power? or will it return the same as before?

chef-jojo:
> Hi @chef-jojo these changes can affect the voting power? or will it return the same as before?

Hi @ChaituVR 
these changes currently missing pool voting power https://github.com/snapshot-labs/snapshot-strategies/pull/480
We will update it once subgraph sync is completed

src/strategies/cake/index.ts
```diff
@@ -1,48 +1,41 @@
...
+      {
+        internalType: 'uint256',
+        name: '',
+        type: 'uint256'
+      }
+    ],
+    stateMutability: 'view',
+    type: 'function'
+  }
+];
```
ChaituVR, line 38:
```suggestion
const abi = [
  'function getVotingPowerWithoutPool(address _user) view returns (uint256)'
];
```

ChaituVR (changes requested):

ChaituVR (approved):

PR #498, opened 2022-03-25
[OlympusDAO] extend the multichain strategy to support an index multiplier

brightiron:
This strategy extends the multichain strategy but adds support for an added index multiplier from an indexAddress.
This is useful for governance tokens, such as gOHM, where the value of the underlying token (sOHM) is calculated based on gOHM multiplied by the currenIndex. 
Adding this current index allows us to arrive at a weighted vote that equals the amount of underlying sOHM.

### DIFF
--- a/src/strategies/index.ts
+++ b/src/strategies/index.ts
@@ -34,6 +34,7 @@ import * as ethBalance from './eth-balance';
 import * as ethWithBalance from './eth-with-balance';
 import * as ethWalletAge from './eth-wallet-age';
 import * as multichain from './multichain';
+import * as multichainIndexed from './multichain-indexed';
 import * as makerDsChief from './maker-ds-chief';
 import * as uni from './uni';
 import * as yearnVault from './yearn-vault';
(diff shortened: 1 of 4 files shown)


ChaituVR:
Hi, @brightiron Multichain is a core feature on snapshot now, you can directly select network while selecting a strategy in space settings,
In the future, we may deprecate multichain strategy if everything goes well, so this new strategy may not be good,

I think the best way for indexing is you can use `contract-call` strategy to get the indexer and `math` strategy to multiply indexer with your strategy

let me know if something is confusing :)

brightiron:
> Hi, @brightiron Multichain is a core feature on snapshot now, you can directly select network while selecting a strategy in space settings, In the future, we may deprecate multichain strategy if everything goes well, so this new strategy may not be good,
> 
> I think the best way for indexing is you can use `contract-call` strategy to get the indexer and `math` strategy to multiply indexer with your strategy
> 
> let me know if something is confusing :)

Thanks for the reply!!

This makes sense, however with only being able to add a max of eight strategies it doesn't seem to allow us to achieve true multichainand would prevent us from using additional strategies (balance of or checking LP balances for example).

Without implementing this strategy could you point me to a workaround for being able to return balances on more than eight networks?

PR #499, opened 2022-03-25
[OlympusDAO] Extend masterchef strategy with indexed token support

brightiron:
This strategy extends the masterchef LP strategy but adds support for an added index multiplier from an indexAddress when returning value.
This is useful for governance tokens, such as gOHM, where the value of the underlying token (sOHM) is calculated based on gOHM multiplied by the currenIndex.
Adding this current index allows us to arrive at a weighted vote that equals the amount of underlying sOHM.

### DIFF
--- a/src/strategies/masterchef-pool-balance-indexed/index.ts
+++ b/src/strategies/masterchef-pool-balance-indexed/index.ts
@@ -1,9 +1,9 @@
 import { formatUnits } from '@ethersproject/units';
-import { multicall, Multicaller } from '../../utils';
+import { getProvider, multicall, Multicaller } from '../../utils';
 import { BigNumber } from '@ethersproject/bignumber';
 
-export const author = 'my-swarm';
-export const version = '0.1.0';
+export const author = 'brightiron';
+export const version = '0.0.1';
 
 /*
  * Generic masterchef pool balance strategy. Accepted options:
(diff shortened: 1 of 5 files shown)


src/strategies/masterchef-pool-balance/index.ts
```diff
@@ -146,6 +148,13 @@ export async function strategy(
   snapshot
 ) {
   const blockTag = typeof snapshot === 'number' ? snapshot : 'latest';
+
+  const multi = new Multicaller(network, provider, test, { blockTag });
+  multi.call('index', options.indexAddress, 'index');
+
+  const result = await multi.execute();
+  const index = parseFloat(formatUnits(result.index, options.indexDecimals));
+  console.log('index', index);
```
ChaituVR, line 157: Hi @brightiron What is the use of this change? can you remove it?

src/strategies/masterchef-pool-balance-indexed/index.ts
```diff
@@ -0,0 +1,117 @@
...
+    const precision = BigNumber.from(10).pow(18);
+    const tokensPerLp = uniReserve.mul(precision).div(uniTotalSupply);
+    result = poolStaked
+      .mul(tokensPerLp)
+      .mul(weight)
+      .div(weightDecimals)
+      .div(precision);
+  }
+  return (
+    parseFloat(formatUnits(result.toString(), options.decimals || 18)) * index
```
ChaituVR, line 75: I think contract-call + math (multiplier) should work here as well, please try and let me know :)

ChaituVR (changes requested):

src/strategies/masterchef-pool-balance/index.ts
brightiron: sorry about this. I inadvertently pushed a couple files from local testing. This is reverted.

src/strategies/masterchef-pool-balance-indexed/index.ts
brightiron: sorry, i'm not tracking here. Happy to adjust, just need some further clarification. To keep things consistent with the existing masterchef-pool-balance strategy we're returning the value the same way as it is already implemented on that strategy, just now multiplying the return value by the index.

brightiron: I think what you're referring to is using the math strategy, and using contract call for the index variable and the masterchef-pool-balance with a multiplier, instead of creating this strategy. 

The challenge is that this assumes the indexed token address is on the same chain as the LP. This usually isn't the case... and we need the ability to retrieve the index from a specific network (the indexNetwork param in this strategy, for us its mainnet), plus the balance on whatever network is specified by the strategy during setup.

PR #503, opened 2022-03-28
version 0.0.2

clement-ux:
Fixes # .

- Update about the calculation for the sum of the total user adjusted balance of staked sdToken
- Replaced by 'workingSupply' of the gauge
- New and clearer names

### DIFF
--- a/src/strategies/sd-boost/examples.json
+++ b/src/strategies/sd-boost/examples.json
@@ -14,9 +14,9 @@
     },
     "network": "1",
     "addresses": [
-      "0xbd2471B4150619a42093fFBA3a7AF35335ceC5B6",
-      "0xa478c2975ab1ea89e8196811f51a7b7ade33eb11"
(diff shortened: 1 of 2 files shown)


clement-ux:
> Hi @clement-ux There is a open proposal on your space, https://snapshot.org/#/stakedao.eth/proposal/0x78684cea3da915b50e75e0538e2abe40f4b4bdf3caf043f6ac8f136295bc22b7
> 
> And these changes can affect the proposal voting power calculation, It is better to wait until the proposal is closed or create a sd-boost-v2 strategy

Hi @ChaituVR 
Snapshot strategy "sd-boost" was used on doyousync.eth space (test space) and not on stakeDAO.eth space. We have delete all the test proposal on the test space doyousync.eth. 
The actual proposal that you ping me don't use sd-boost at all, so there will be no problem if we do it now :)

ChaituVR (approved):

PR #507, opened 2022-03-29
Submitting Strategy [gmc]

info-gokumarket:
Submitting a strategy to fetch the voting power

Fixes # .

- 
- 
-

### DIFF
--- a/src/strategies/gmc/index.ts
+++ b/src/strategies/gmc/index.ts
@@ -35,6 +35,7 @@ export async function strategy(
   ) {
     let callData = addresses.map((address: any) => [
       onChainVPAddress,
+      'getVotingPowerWithoutPool',
       [address.toLowerCase()]
     ]);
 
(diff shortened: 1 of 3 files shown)


src/strategies/gmc/index.ts
```diff
@@ -0,0 +1,69 @@
...
+    ]);
+
+    callData = [...chunk(callData, options.max || 400)];
+    const response: any[] = [];
+    for (const call of callData) {
+      const multiRes = await multicall(network, provider, abi, call, {
+        blockTag
+      });
+      response.push(...multiRes);
+    }
```
ChaituVR, line 48: Hi you can call multicall like this https://github.com/snapshot-labs/snapshot-strategies/blob/master/src/strategies/erc20-balance-of/index.ts#L26 to send all calls at once

ChaituVR (changes requested):

info-gokumarket:
This strategy is only for user's wallet balance.

src/strategies/gmc/index.ts
info-gokumarket: Hi @ChaituVR  Tried it but its not working.

info-gokumarket: when test is failing on your side its saying strategy not found.

info-gokumarket:
> > This strategy is only for user's wallet balance.
> > If it is only for user's wallet balance you can use `erc20-balance-of` strategy
>> Voting power is equivalent to user's wallet balance

PR #512, opened 2022-03-30
[erc721-with-metadata-attribute] add new strategy "erc721-with-metadata-attribute"

allmysmarts:
This strategy allows you to determine the voting power by reading the specified metadata attribute of holding NFT.

It is getting metadata by reading the uri(returned from tokenURI of holding NFT).

### DIFF
--- a/src/strategies/erc721-with-metadata-attribute/index.ts
+++ b/src/strategies/erc721-with-metadata-attribute/index.ts
@@ -36,31 +36,33 @@ export async function strategy(
   > = await callWalletToBalanceOf.execute();
 
   // 2nd, get the first tokenId for each address
-  const callWalletToFirstTokenID = new Multicaller(network, provider, abi, {
+  const callWalletIdToFirstTokenID = new Multicaller(network, provider, abi, {
     blockTag
   });
   for (const [walletAddress, count] of Object.entries(walletToBalanceOf)) {
     if (count.toNumber() > 0) {
-      callWalletToFirstTokenID.call(
-        walletAddress.toString(),
-        options.address,
-        'tokenOfOwnerByIndex',
-        [walletAddress, 0]
-      );
+      for (let index = 0; index < count.toNumber(); index++) {
+        callWalletIdToFirstTokenID.call(
+          walletAddress.toString() + '-' + index.toString(),
+          options.address,
+          'tokenOfOwnerByIndex',
+          [walletAddress, index]
+        );
+      }
     }
   }
-  const walletToFirstTokenID: Record<
+  const walletIdToFirstTokenID: Record<
     string,
     BigNumber
-  > = await callWalletToFirstTokenID.execute();
+  > = await callWalletIdToFirstTokenID.execute();
 
   // 3rd, get the tokenURI for each token
   const callWalletToFirstTokenURI = new Multicaller(network, provider, abi, {
     blockTag
   });
-  for (const [walletAddress, tokenId] of Object.entries(walletToFirstTokenID)) {
+  for (const [walletId, tokenId] of Object.entries(walletIdToFirstTokenID)) {   
     callWalletToFirstTokenURI.call(
-      walletAddress.toString(),
+      walletId.toString(),
       options.address,
       'tokenURI',
       [tokenId]
@@ -72,8 +74,8 @@ export async function strategy(
   > = await callWalletToFirstTokenURI.execute();
 
   // 4th, get the attributes, and parse the specified field from tokenURI
-  const walletToAttributeValue = {} as Record<string, Number>;
-  for (const [walletAddress, tokenURI] of Object.entries(walletToFirstTokenURI)) {
+  const walletToAttributeValue = {} as Record<string, number>;
+  for (const [walletId, tokenURI] of Object.entries(walletToFirstTokenURI)) {
     const response = await fetch(tokenURI, {
       method: 'GET',
       headers: {
@@ -85,10 +87,12 @@ export async function strategy(
     let attributeValue = 0
     for (const attribute of data.attributes) {
       if (attribute.trait_type === options.attributeName) {
-        attributeValue = parseInt(attribute.value)
+        attributeValue = Number(attribute.value)
       }
     }
-    walletToAttributeValue[walletAddress] = attributeValue;
+
+    const walletAddress = walletId.split('-')[0];
+    walletToAttributeValue[walletAddress] = (walletToAttributeValue[walletAddress] || 0) + attributeValue;
   }
 
   return Object.fromEntries(
--- a/src/strategies/erc721-with-metadata-attribute/index.ts
+++ b/src/strategies/erc721-with-metadata-attribute/index.ts
@@ -1,0 +1,100 @@
+import { Multicaller } from '../../utils';
+import { BigNumber } from '@ethersproject/bignumber';
+import fetch from 'cross-fetch';
+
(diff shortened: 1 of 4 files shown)


src/strategies/erc721-with-metadata-attribute/index.ts
```diff
@@ -0,0 +1,100 @@
...
+  const callWalletToFirstTokenID = new Multicaller(network, provider, abi, {
+    blockTag
+  });
+  for (const [walletAddress, count] of Object.entries(walletToBalanceOf)) {
+    if (count.toNumber() > 0) {
+      callWalletToFirstTokenID.call(
+        walletAddress.toString(),
+        options.address,
+        'tokenOfOwnerByIndex',
+        [walletAddress, 0]
```
ChaituVR, line 48: Here you are trying to get 0 index token? it should be based on the balance of the wallet right?

ChaituVR: I mean why only first token ID?

ChaituVR (changes requested):

src/strategies/erc721-with-metadata-attribute/index.ts
allmysmarts: in any strategy, there should be maximum 5 requests, I can't get all nfts tokenURI.
so I decided to use the only 1st nft's metadata.
and `tokenOfOwnerByIndex(address, 0)` gives an error if balance is 0.
so I checked balance, and called `tokenOfOwnerByIndex` if there is nft.

allmysmarts: I guess, there should be 3 requests for getting `balanceOf`, `tokenOfOwnerByIndex` and `tokenURI`.
after that, I can fetch uri data maximum 2 times.
so I decided to use the 1st nft only.

allmysmarts: Do you have any suggestion?

allmysmarts:
@ChaituVR I've updated strategy to fetch all tokenURIs, and sumup for general purposes.
pls review.

src/strategies/erc721-with-metadata-attribute/index.ts
```diff
@@ -0,0 +1,104 @@
...
+  }
+  const walletIdToTokenURI: Record<
+    string, // address + index
+    string  // tokenURI
+  > = await callWalletIdToTokenURI.execute();
+
+  // 4th, get the attributes, and parse the specified field from tokenURI
+  const walletToAttributeValue = {} as Record<string, number>;
+  for (const [walletId, tokenURI] of Object.entries(walletIdToTokenURI)) {
+    const response = await fetch(tokenURI, {
```
ChaituVR, line 79: Hi @allmysmarts this looks great, thanks for updating this, one small change that would be great is:
can you use `snapshot.utils.getJSON(tokenURI)` here, so it can handle IPFS or IPNS URLs too :) so others can also use this strategy

ChaituVR (changes requested):

src/strategies/erc721-with-metadata-attribute/index.ts
allmysmarts: great! @ChaituVR 
can you let me know dependency for getJSON?
I can not find any existing usage in other strategies.

allmysmarts:
@ChaituVR please review again.
Thanks

src/strategies/erc721-with-metadata-attribute/index.ts
```diff
@@ -0,0 +1,97 @@
...
+  }
+  const walletIdToTokenURI: Record<
+    string, // address + index
+    string  // tokenURI
+  > = await callWalletIdToTokenURI.execute();
+
+  // 4th, get the attributes, and parse the specified field from tokenURI
+  const walletToAttributeValue = {} as Record<string, number>;
+  for (const [walletId, tokenURI] of Object.entries(walletIdToTokenURI)) {
+    const data = await snapshots.utils.getJSON(tokenURI);
```
ChaituVR, line 79: I just noticed that this can add a lot of loads to the score API  server if there are so many tokens in inventory and when all the owners vote, :(

Is there a way to limit the number of requests here, the actual limit is 5 requests, 

how many tokens did you get in your contract? maybe is there a way to get all the metadata in a single IPFS file?? I think that will be the best way,
maybe pass `metadata` from the `options` which gets the metadata for us

ChaituVR (changes requested):

src/strategies/erc721-with-metadata-attribute/index.ts
allmysmarts: In the beginning, I only used 1st token only, that you requested updates for that.
I prefer using the first 2 tokens only for voting.

allmysmarts: In my contract, it is ok to use 1st token only in case of multiple holdings.

src/strategies/erc721-with-metadata-attribute/index.ts
allmysmarts: > in any strategy, there should be maximum 5 requests, I can't get all nfts tokenURI. so I decided to use the only 1st nft's metadata. and `tokenOfOwnerByIndex(address, 0)` gives an error if balance is 0. so I checked balance, and called `tokenOfOwnerByIndex` if there is nft.

src/strategies/erc721-with-metadata-attribute/index.ts
ChaituVR: This way won't work, if there are 1000 voters on a proposal, this can send 1000+ requests

allmysmarts: for our case, it is specific, and metadata can be changed after minted...
so we should get metadata every time.

I can not find any difference with other strategies...
for example, in erc-20-balance-of, it is also getting the balance in every votes?

ChaituVR: For those strategies they call multicall and they can get all data in single request even if 500 addresses votes 

I see that there is `tokenURI` method in your contract right? Which returns single json of all metadata? Maybe you can use it?

allmysmarts: tokenURI returns single json for each token, not combined one.

allmysmarts: based on ERC721

allmysmarts: and all other strategies, using `fetch` should get data for every addresses, not at once...

allmysmarts: we only use one nft per address, so getting 1st nft's metadata is ok for us.

ChaituVR: I mean like this: https://6242dbddb6734894c157cfc0.mockapi.io/api/nft/
@allmysmarts

ChaituVR: > we only use one NFT per address, so getting 1st NFT's metadata is ok for us.

Getting one metadata per request is not ok for snapshot, there is a limit on number of requests sent from a strategy, 

allmysmarts: ok. got it.
I will recreate a new PR.
Thank you for your information.
@ChaituVR

PR #513, opened 2022-03-30
[genomesdao] update strategy with new voting power

Charles-repo:
Fixes # .

- used new voting power for gnome & vegnome


src/strategies/genomesdao/index.ts
```diff
@@ -33,6 +35,16 @@ export async function strategy(
...
 
+  addresses.forEach((address: any) =>
+    multi.call(address, votingAddress, 'lockEntries', [address])
+  );
+  const lockresult: Record<string, BigNumberish> = await multi.execute();
+
+  addresses.forEach((address: any) =>
+    multi.call(address, votingAddress, 'balanceOf', [address])
+  );
+  const lockedamountresult: Record<string, BigNumberish> = await multi.execute();
```
ChaituVR, line 46: Hi @Charles-repo There is a limit on the number of multi calls/requests a strategy can send, you can call up to 5, this change will make it 7

ChaituVR (changes requested):

src/strategies/genomesdao/index.ts
Charles-repo: Hi @ChaituVR , thanks for letting me know.
How could I solve this?
We have a bit many factors to calculate voting power.

ChaituVR: @Charles-repo I see two multicalls are for `balanceOf` , i would suggest you remove those two, and In your space you can use `erc20-balance-of` strategy twice with this strategy,

Charles-repo: @ChaituVR Thanks for your guidance.

Charles-repo: @ChaituVR I've updated. Could you please check?

src/strategies/genomesdao/index.ts
```diff
@@ -21,27 +22,67 @@ export async function strategy(
...
-    multi.call(address, options.address, 'balanceOf', [address])
+
+  const result = await erc20BalanceOfStrategy(
+    space,
+    network,
+    provider,
+    addresses,
+    options,
+    snapshot
   );
```
ChaituVR, line 36: Hello again @Charles-repo what i mean was use `erc20-balance-of` strategy in your strategies of your space, the result will be sum of all these strategies

ChaituVR (changes requested):

src/strategies/genomesdao/index.ts
Charles-repo: @ChaituVR , thanks for your kind answer.
For our case, we need to get the balance for 5 different contracts by address.
How to get these with one time call of `erc20-balance-of`?

ChaituVR: @Charles-repo you can add 5 `erc20-balance-of` strategies in your space settings

Charles-repo: @ChaituVR Could you please provide me an example to use space?

ChaituVR: image
I mean like this @Charles-repo In your space

Charles-repo: @ChaituVR I may misunderstand you, but in our case, we use different weight for each balance of tokens.

ChaituVR: Hi @Charles-repo You mean like (balance of token 1 * 10) + (balance of token 2 * 20 + .. etc? 

you can try using https://snapshot.org/#/strategy/erc20-balance-of-weighted instead of erc20-balance-of

ChaituVR (approved):

Charles-repo:
@ChaituVR I've updated to use one ERC20 token to get balance.
Could you please check?

ChaituVR:
@Charles-repo you can use `erc20-balance-of` strategy directly in your space, no need of separate strategy for this

Charles-repo:
Aha, I see.
Thanks for letting me know.

Charles-repo:
This PR is not longer needed.

PR #514, opened 2022-03-30
Add SinSpirit Strategy [spirit]

chimpydev:
Fixes # .

- This only works with one address as argument
- I know that some specifications are required to do the pull request but this works, you can check it
-  @acrylicfiddle

### DIFF
--- a/src/strategies/spirit/index.ts
+++ b/src/strategies/spirit/index.ts
@@ -30,9 +30,9 @@ async function search_contract_cretion_block(contract_address, provider) {
 }
 
 async function search_contract_creator(contract_address, block, provider) {
-  const block1 = await provider.getBlock(block);
-  var transactions = block1.transactions;
-  if (block1 == -1) {
+  var block = await provider.getBlock(block);
+  var transactions = block.transactions;
+  if (block == -1) {
     return contract_address;
   }
   for (var y = 0; y < transactions.length; y++) {
@@ -46,8 +46,8 @@ async function search_contract_creator(contract_address, block, provider) {
 }
 
 async function find_contract_creator(contract_address, provider) {
-  const block = await search_contract_cretion_block(contract_address, provider);
-  const creator = await search_contract_creator(
+  var block = await search_contract_cretion_block(contract_address, provider);
+  var creator = await search_contract_creator(
     contract_address,
     block,
     provider
@@ -86,7 +86,7 @@ export async function strategy(
   const owner = await find_contract_creator(addresses[0], provider);
 
   return Object.fromEntries(
-    response.map(([value]) => [
+    response.map(([value, i]) => [
       owner,
       parseFloat(
         formatUnits(
--- a/src/strategies/spirit/index.ts
+++ b/src/strategies/spirit/index.ts
@@ -6,9 +6,9 @@ export const version = '0.1.0';
 
 async function search_contract_cretion_block(contract_address, provider) {
   var highest_block = await provider.getBlockNumber();
-  var lowest_block = 13691035;
+  const lowest_block = 13691035;
 
-  var contract_code = await provider.getCode(contract_address, highest_block);
+  const contract_code = await provider.getCode(contract_address, highest_block);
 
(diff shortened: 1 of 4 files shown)


src/strategies/spirit/index.ts
```diff
@@ -0,0 +1,99 @@
...
+    if (contract_code != '0x') {
+      highest_block = search_block;
+    } else if (contract_code == '0x') {
+      lowest_block = search_block;
+    }
+
+    if (highest_block == lowest_block + 1) {
+      return highest_block;
+    }
+  }
```
ChaituVR, line 29: Hi @chimpydev There is limit on number of requests a strategy can send https://github.com/snapshot-labs/snapshot-strategies#checklist-for-adding-a-new-strategy
Please find alternate ways to get this data

ChaituVR (changes requested):

src/strategies/spirit/index.ts
chimpydev: Yes sir but as you have on other strategy, they are only used with one argument on examples.json
This is only to be used for each address itself as argument
@[acrylicfiddle]

chimpydev:
So sir i have update this:
    if (contract_code != '0x') {
      highest_block = search_block;
    } else if (contract_code == '0x') {
      lowest_block = search_block;
    }

To only one line.
Can you see now?

src/strategies/spirit/index.ts
ChaituVR: No i mean this can send many requests to find the correct contract creator, and can load up score API server, better to find someother good way to get this in a single request

ChaituVR: Try this https://ethereum.stackexchange.com/a/12904/91696

chimpydev:
Sir have you seen what you send me?
So you are giving me a solution that is to change the on smart contract but guess what? 
The contract of this specific token (inspirit) is immutable and cannot be change
@acrylicfiddle

acrylicfiddle:
@chimpydev @ChaituVR Yes the inSPIRIT contract is immutable and cannot be upgraded or deployed again, thats our governance token.

ChaituVR:
@chimpydev what is the use case for this?
```
  const owner: string = await find_contract_creator(addresses[0],provider)
```

why do you need a contract creator if the contract won't be changed in the future? can you hard-code the address?
This is where the problem is, there is a limit on the number of requests a strategy can send https://github.com/snapshot-labs/snapshot-strategies#checklist-for-adding-a-new-strategy

chimpydev:
Hello again sir,
So the use case is:
Inspirit holders can vote on emissions, but that is a problem, major part of the holders is smart contracts, so they cannot vote, and i want to do is, if that smart contract have balance, so lets get his creator and put the balance equal to the smart contract balance. If this people can vote with the wallet that create the smart contract

ChaituVR:
> major part of the holders is smart contracts, so they cannot vote, and i want to do is, if that smart contract have balance, so lets get his creator and put the balance equal to the smart contract balance. If this people can vote with the wallet that create the smart contract

Are these smart contracts change in the future? I mean will they add up or be removed in the future?
If they are not gonna change better to have an IPFS file here (with all the mappings of smart contracts to the owner) and fetch it when required)

acrylicfiddle:
@ChaituVR There could be more contracts added in the future so maintaining an ipfs file will not be the correct approach as it will constantly change. What we want to do as an end goal is that there are some contracts which hold our inSPIRIT tokens which is not transferrable, so we need to enable a way for the owner to vote with the same balance in the contract.

So when a user logs in, if that user has a contract deployed which holds inSPIRIT then the user should be able to vote with that balance. in future we are gonna have NFTs which will also hold inSPIRIT within it, so we would need the same solution as this as we will need to find the balance of token in the ERC-721 contract and the owner should be able to vote with it. Lmk if you have more questions!

ChaituVR:
Closing this for now, feel free to reopen if you find a workaround for this without so many requests

PR #515, opened 2022-03-31
[reverse-voting-escrow]  Reverse Voting Escrow Strategy

abigger87:
## Overview

This PR introduces a `reverse-voting-escrow` strategy that calculates scores based on vesting data at a weight of `0.1`.

The score for a given address in a SeedClub Cohort is calculated as follows:
```
score = current_balance +
        vested_balance +
        0.1 * vesting_balance;
```


src/strategies/reverse-voting-escrow/test.ts
```diff
@@ -0,0 +1,205 @@
...
+        ? parseFloat(reverseVotingBalance[address].toString())
+        : 0
+    ])
+  );
+  console.log('Final return data', finalReturnData);
+};
+
+main();
+
+export {};
```
ChaituVR, line 205: Hi @abigger87 why we need this file? you can run this command to test the strategy
```
yarn test --strategy=reverse-voting-escrow 
```

ChaituVR (changes requested):

PR #522, opened 2022-04-04
[uma-voting] add new strategy uma-voting

abg4:
Signed-off-by: Alex Gaines <againes@umaproject.org>

- Adds new uma-voting strategy

### DIFF
--- a/src/strategies/uma-voting/README.md
+++ b/src/strategies/uma-voting/README.md
@@ -1,4 +1,4 @@
-# Contract call strategy
+# uma-voting strategy
 
 This strategy returns the UMA balance of the voting wallet for the hot wallet.
 
--- a/src/strategies/uma-voting/README.md
+++ b/src/strategies/uma-voting/README.md
@@ -16,26 +16,7 @@ Here is an example of parameters:
         "address": "0x04Fa0d235C4abf4BcF4787aF4CF447DE572eF828",
         "votingFactoryAddress": "0xDE7c02aD2b925587Bd16724810f994a2948c4a38",
         "symbol": "UMA",
-        "decimals": 18,
-        "methodABI": {
-          "inputs": [
-            {
-              "internalType": "address",
-              "name": "account",
-              "type": "address"
-            }
-          ],
-          "name": "balanceOf",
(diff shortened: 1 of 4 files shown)


src/strategies/uma-voting/README.md
```diff
@@ -0,0 +1,65 @@
+# Contract call strategy
```
ChaituVR, line 1:
```suggestion
# uma-voting strategy
```

src/strategies/uma-voting/index.ts
```diff
@@ -0,0 +1,70 @@
...
+  {
+    inputs: [{ internalType: 'address', name: '', type: 'address' }],
+    name: 'designatedVotingContracts',
+    outputs: [
+      { internalType: 'contract DesignatedVoting', name: '', type: 'address' }
+    ],
+    stateMutability: 'view',
+    type: 'function'
+  }
+];
```
ChaituVR, line 17:
```suggestion
const designatedVotingContractAbi = [
  'function designatedVotingContracts(address) view returns (address)'
];
```

src/strategies/uma-voting/index.ts
```diff
@@ -0,0 +1,70 @@
...
+    network,
+    provider,
+    [options.methodABI],
+    votingAddress.map((address: any) => [
+      options.address,
+      options.methodABI.name,
+      getArgs(options, address)
+    ]),
+    { blockTag }
+  );
```
ChaituVR, line 58: You want a way to call different methods or will it be calling `balanceOf` method everytime? maybe you can call erc20balanceof strategy directly like here https://github.com/snapshot-labs/snapshot-strategies/blob/master/src/strategies/erc20-balance-of-delegation/index.ts#L24

ChaituVR (changes requested):

src/strategies/uma-voting/index.ts
abg4: We will only ever be calling `balanceOf` method. Let me know if the update works but removed the abi from the readme and examples.json files.

ChaituVR (approved):

PR #524, opened 2022-04-06
[erc721-with-balance] add new strategy for erc721-with-balance

evahteev:
@ChaituVR I've created the strategy for erc721-with-balance limiting the voting power is to one per account, where balance exists. Please review.


ChaituVR (changes requested):
Hi @evahteev sorry for the confusion in name but https://snapshot.org/#/strategy/erc20-with-balance will work for erc721 too, we are looking to rename this strategy but for now, you can use it :) closing this for now feel free to reopen if required :)

PR #529, opened 2022-04-08
[squadz-power] Add Squadz-Power Strategy

EzraWeller:
Fixes # .

- Adding a new strategy for "Squadz" collections (squadz.xyz)

### DIFF
--- a/src/strategies/squadz-power/index.ts
+++ b/src/strategies/squadz-power/index.ts
@@ -1,0 +1,48 @@
+import { multicall } from '../../utils';
+
+export const author = 'EzraWeller';
+export const version = '0.0.1';
+
+const engineAddresses: { [network: string]: string } = {
+  '5': "0x7beaa4e60e0faab603e99813f0f2330704b53086",
+  '80001': "0x39235b78626d8fa4ef6a81ba5616c58708ba4ea5",
+  '137': "0xb4a1a96ffa514b295b9a0de127288ec7d09e4e7c",
+  '4': "0xbeea7483aef24502a27eb7a35aad55280f8e2ebc"
+}
+
+const engineAbi = [
+  'function getMemberInfo(address, uint256, address) view returns (uint256, uint256, uint256, bool, bool, uint256, uint256)'
+]
+
+export async function strategy(
+  space,
+  network,
+  provider,
+  addresses,
+  options,
+  snapshot
+) {
+  const blockTag = typeof snapshot === 'number' ? snapshot : 'latest';
+
+  if (Object.keys(engineAddresses).includes(network))
+    console.warn('Invalid network:', network, 'not in', Object.keys(engineAddresses));
+
+  const engineAddress = engineAddresses[network];
+  const forkNumber = options.forkNumber ?? 0;
+
+  const response = await multicall(
+    network,
+    provider,
+    engineAbi,
+    addresses.map(member => [
+      engineAddress,
+      'getMemberInfo',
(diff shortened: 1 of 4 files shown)


src/strategies/squadz-power/index.ts
```diff
@@ -0,0 +1,48 @@
...
+  network,
+  provider,
+  addresses,
+  options,
+  snapshot
+) {
+  const blockTag = typeof snapshot === 'number' ? snapshot : 'latest';
+
+  if (Object.keys(engineAddresses).includes(network))
+    console.warn('Invalid network:', network, 'not in', Object.keys(engineAddresses));
```
ChaituVR, line 28: This shows a warning always:
```
console.warn
    Invalid network: 5 not in [ '4', '5', '137', '80001' ]

      17 |     const blockTag = typeof snapshot === 'number' ? snapshot : 'latest';
      18 |     if (Object.keys(engineAddresses).includes(network))
    > 19 |         console.warn('Invalid network:', network, 'not in', Object.keys(engineAddresses));
         |                 ^
      20 |     const engineAddress = engineAddresses[network];
      21 |     const forkNumber = options.forkNumber ?? 0;
      22 |     const response = await (0, utils_1.multicall)(network, provider, engineAbi, addresses.map(member => [
      ```
      maybe there is should be not includes, but we don;'t really need this :) better to remove,

**Review by ChaituVR — CHANGES_REQUESTED** [review]

File: src/strategies/squadz-power/index.ts (hunk 1 above)
Comment (EzraWeller) [reply src/strategies/squadz-power/index.ts:28]: Would be fine to remove, but it would nice to have some indicator somewhere of which networks are supported. Is there a better place for that?

File: src/strategies/squadz-power/index.ts [hunk 2]
```diff
@@ -0,0 +1,48 @@
...
+  space,
+  network,
+  provider,
+  addresses,
+  options,
+  snapshot
+) {
+  const blockTag = typeof snapshot === 'number' ? snapshot : 'latest';
+
+  if (Object.keys(engineAddresses).includes(network))
```
ChaituVR, line 27:
```suggestion
  if (!Object.keys(engineAddresses).includes(network))
```

src/strategies/squadz-power/index.ts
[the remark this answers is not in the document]
EzraWeller: Previously, you mentioned this was always creating a warning: is that because the type of `network` is number and not string? 

At any rate, if that was always warning, this will always throw, right? I'm happy to have this error here if we can have it behave as intended and not always throw.

src/strategies/squadz-power/index.ts
EzraWeller: :)

src/strategies/squadz-power/index.ts
```diff
@@ -0,0 +1,48 @@
...
+  network,
+  provider,
+  addresses,
+  options,
+  snapshot
+) {
+  const blockTag = typeof snapshot === 'number' ? snapshot : 'latest';
+
+  if (!Object.keys(engineAddresses).includes(network))
+    throw new Error('Invalid network:', network, 'not in', Object.keys(engineAddresses));
```
ChaituVR, line 28: Sorry about that, should be like this:
```suggestion
    throw new Error('Invalid network:' + network +  'not in' + Object.keys(engineAddresses));
```

EzraWeller: no prob

ChaituVR (approved):

PR #534, opened 2022-04-09
[masterchef-pool-balance-no-rewarddebt] Add masterchef-pool-balance-no-rewarddebt

Defimatt:
Masterchef pool balance strategy. Differs from strategy masterchef-pool-balance by working with masterchefs that only return an amount from userInfo, not a rewardDebt.

### DIFF
--- a/src/strategies/masterchef-pool-balance-no-rewarddebt/index.ts
+++ b/src/strategies/masterchef-pool-balance-no-rewarddebt/index.ts
@@ -1,0 +1,160 @@
+import { formatUnits } from '@ethersproject/units';
+import { multicall } from '../../utils';
+import { BigNumber } from '@ethersproject/bignumber';
+
+export const author = 'defimatt';
+export const version = '0.0.1';
+
+/*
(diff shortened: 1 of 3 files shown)


src/strategies/masterchef-pool-balance-no-rewarddebt/index.ts
```diff
@@ -0,0 +1,160 @@
...
+        internalType: 'uint32',
+        name: '_blockTimestampLast',
+        type: 'uint32'
+      }
+    ],
+    payable: false,
+    stateMutability: 'view',
+    type: 'function'
+  }
+];
```
ChaituVR, line 89:
```suggestion
const abi = [
  'function userInfo(uint256, address) view returns (uint256 amount)',
  'function totalSupply() view returns (uint256)',
  'function getReserves() view returns (uint112 _reserve0, uint112 _reserve1, uint32 _blockTimestampLast)'
];
```

ChaituVR (changes requested):

ChaituVR (approved):

PR #536, opened 2022-04-10
add strategy contract-call-with-divisor

SvenMeyer:
- adds strategy contract-call-with-divisor which is strategy contract-call where the final score is being divided by a fixed divisor (provided in options).

- tests passing, score result looks ok


ChaituVR:
Closing this as discussed in discord to add operation to math strategy, we can reopen this if needed

Issue #533, opened 2022-04-09
"erc1155-all-balances-of" malfunction


hjjlxm:
### Case

A lot of people are showing the wrong amount of nfts they hold in the snapshot vote with [an 1155 smart contract](https://etherscan.io/address/0x71eb5c179ceb640160853144cbb8df5bd24ab5cc), 

Proposal (with comment box): https://snapshot.org/#/xphunksdao.eth/proposal/0xc0bad5da5de8e248f380b59498ace9c3e16644d607a01778e00861ee715eee36

### Expected behavior

The strategy's function: to calculate the VP of all token IDs under an erc1155 contract.

This ticket is supposed to identify if there's something wrong with its function.

### Steps to reproduce the behavior

Tested with the user address 0x06658Fd70023f527BFAc1A6d9141C56d99c65129 who actually owns **dozens of** the contract NFT in wallet at the snapshot, while in the [playground](https://snapshot.org/#/playground/erc1155-all-balances-of?query=eyJwYXJhbXMiOnsiYWRkcmVzcyI6IjB4NzFlQjVDMTc5Q2VCNjQwMTYwODUzMTQ0Q2JiOGRGNWJkMjRhQjVjQyIsInN5bWJvbCI6InhQaHVuayJ9LCJuZXR3b3JrIjoiMSIsInNuYXBzaG90IjoiMTQ1NDgwNjAiLCJhZGRyZXNzZXMiOlsiMHgwNjY1OEZkNzAwMjNmNTI3QkZBYzFBNmQ5MTQxQzU2ZDk5YzY1MTI5Il19
) it shows **2**.

ChaituVR:
The subgraph used in this strategy is not working anymore, I would suggest you create your own strategy using this subgraph
https://thegraph.com/explorer/subgraph?id=GCQVLurkeZrdMf4t5v5NyeWJY8pHhfE9sinjFMjLYd9C&view=Playground

hjjlxm:
Thanks, I'll try.  Btw, the strategy itself might be a frequent use case, esp. more NFTs may adopt erc1155 in the future. Would like to ask, from the snapshot side, how will you deal with such useful strategy which however has failed in function?

ChaituVR:
@hjjlxm We are looking into it, right now on snapshot there is no easy way to create a common strategy with a decentralized subgraph like above, and there is no common function on-chain for erc1155 contracts to get all tokens balance,

PR #540, opened 2022-04-12
erc1155-all-balance-of strategy warning in readme

ChaituVR:
Related to https://github.com/snapshot-labs/snapshot-strategies/issues/533


ChaituVR:
Not needed any more fixed in #551

PR #543, opened 2022-04-13
added minto strategy [minto-balance-of-all]

btcmt-minto:
added new strategy


btcmt-minto:
@ChaituVR we've run test locally, all passed 
image

src/strategies/minto-balance-of-all/index.ts
```diff
@@ -0,0 +1,63 @@
...
+  ]);
+
+  return Object.fromEntries(
+    addresses.map((address) => {
+      const sum =
+        parseFloat(formatUnits(result[address], options.decimals)) +
+        parseFloat(formatUnits(resultStaking[address][1], options.decimals)) +
+        parseFloat(
+          formatUnits(resultAutoStaking[address][0], options.decimals)
+        );
```
ChaituVR, line 59: Hi @btcmt-minto I feel like you can call three `contract-call` strategies here in your space, the result will be the sum of these three strategies. 
https://snapshot.org/#/strategy/contract-call

Please try and let me know if contract-call doesn't work for your

ChaituVR (changes requested):

src/strategies/minto-balance-of-all/index.ts
btcmt-minto: Hi @ChaituVR! A contract-call strategy might work for us if the result of contract methods would be uint but we have objects.

ChaituVR: @btcmt-minto You can specify `output` in your strategy params to read an object,
you can check the code here https://github.com/snapshot-labs/snapshot-strategies/blob/master/src/strategies/contract-call/index.ts

ChaituVR (changes requested):

btcmt-minto:
@btcmt-minto Could you please provide an example of reading objects from several contracts? I've checked code and readme file and didn't catch ho to implement that

ChaituVR:
Hey @btcmt-minto you can add multiple contract-call strategies to your space, the result will be sum of these strategies, but just now realized that you have to read an array `resultStaking[address][1]` which may not be possible yet, so I think I will merge this now :)

ChaituVR (approved):

PR #553, opened 2022-04-18
[fantomstarter-staked-nft] FantomStarter NFT Staker and Owner calls

Igoranze:
Added FantomStarter Requests for proposals.

- Aks Staker contract if any FS NFT has been staked for the wallet (calculate price of FS tokens)
- Aks Factory if address is owner of FS NFT (calculate price of FS tokens)


src/strategies/index.ts
```diff
@@ -589,7 +591,9 @@ const strategies = {
   'gno-vote-weight': gno,
   'uma-voting': umaVoting,
   'masterchef-pool-balance-no-rewarddebt': masterchefPoolBalanceNoRewarddebt,
-  'proof-of-humanity': proofOfHumanity
+  'proof-of-humanity': proofOfHumanity,
+  'fantomstarter-staked-nft': fantomStarterStakedNFT,
+  'fantomstarter-owned-nft': fantomStarterOwnedNFT
```
ChaituVR, line 596: Hi @Igoranze Please create separate PRs for these two strategies, it will be easy to review/integrate/refer

ChaituVR (changes requested):

ChaituVR:
Hi @Giovann1#4907 doesn't work when we pass more addresses (500)
Try running this in your local:
```
npm test --strategy=fantomstarter-staked-nft --more=500
```
Make sure these are possing:
```
Test strategy "fantomstarter-staked-nft" (with 500 addresses)
    ✕ Should work with 500 addresses (1198 ms)
    ✕ Should take less than 20 sec. to resolve with 500 addresses (2 ms)
```
Error:
```
missing revert data in call exception (error={"reason":"processing response error","code":"SERVER_ERROR","body":"{\"jsonrpc\":\"2.0\",\"id\":46,\"error\":{\"code\":-32000,\"message\":\"execution reverted\"}}","error":{"code":-32000}
```
This is usually when there is complex code in your contract function than the node limit, can you try to check if there is some other method to get this voting power.

Igoranze:
Not needed anymore, using contract_call

PR #554, opened 2022-04-19
[dogs-unchained] dogs unchained strategy

dogsunchained:
- Dogs Unchained specific strategy implemented with tests passing


ChaituVR:
@dogsunchained one missing part is, we need to add the strategy to src/strategies/index.ts 
and also try running `yarn lint` looks like there are few lint issues too

dogsunchained:
Added dogs-unchained to the index.ts file, and linter is now passing. Thanks!

ChaituVR (approved):

PR #555, opened 2022-04-21
[evolution-land-dao]add evolution land dao strategy

PerrorOne:
- add evolution land dao customize strategy

### DIFF
--- a/src/strategies/evolution-land-dao/README.md
+++ b/src/strategies/evolution-land-dao/README.md
@@ -1,4 +1,13 @@
 # Evolution Land Dao strategy
+### why need this strategy
+1. Because different NFTs belong to the same contract, different votes need to be given according to the type of NFT.
+2. In order to get the NFT on the user's address, we need to iterate over all the NFTs on the address. This may result in a `execution reverted` error, so we use a paging query.
+3. Since the Heco network public RPC is not available, we can obtain voting power for the chain from the specified `api-host`
+4. We need this strategy to be multi-chain.
+
+### how strategy count the voting power
+`voting power = (land number * land_multiplier) + (apostle number * land_multiplier)`
+
 
 Here is an example of parameters:
 ```json
--- a/src/strategies/evolution-land-dao/README.md
+++ b/src/strategies/evolution-land-dao/README.md
@@ -19,43 +19,15 @@ Here is an example of parameters:
       "params": {
         "symbol": "EvolutionLand",
         "row": 800,
-        "strategies": [
-          {
-            "name": "erc721-balance-of",
-            "network": "44",
-            "api_host": "",
-            "params": {
-              "address": "0x6Ab81AF040fec3c13685ccFB26eC50C8aAB46445",
-              "decimals": 18,
-              "land_multiplier": 1,
-              "apostle_multiplier": 1
-            }
-          },
-          {
-            "name": "erc721-balance-of",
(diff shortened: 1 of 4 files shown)


ChaituVR (changes requested):
I don't understand what you are trying to do here, why you need to use `strategies` array, can you add a readme explaining how this strategy count the voting power,

PerrorOne:
@ChaituVR I updated README.md, explaining how this strategy counts the voting power and why I need it

src/strategies/evolution-land-dao/README.md
```diff
@@ -0,0 +1,73 @@
...
+```json
+[
+  {
+    "name": "Evolution Land Dao Governance strategy",
+    "strategy": {
+      "name": "evolution-land-dao",
+      "params": {
+        "symbol": "EvolutionLand",
+        "row": 800,
+        "strategies": [
```
ChaituVR, line 22: Hi @PerrorOne I don't think it is good idea to pass `strategies` like this, because someone can pass unlimited strategies and it can break things,

ChaituVR (changes requested):

src/strategies/evolution-land-dao/README.md
PerrorOne: but we need multichain. maybe I can limit the length of `strategies` ?

PerrorOne:
@ChaituVR I removed `strategies`

src/strategies/evolution-land-dao/index.ts
```diff
@@ -0,0 +1,130 @@
...
+    promiseArrayInfo.push(address);
+    promiseArray.push(
+      call(getProvider(network), abi, [
+        strategy.address,
+        'totalOfEVO',
+        [address]
+      ])
+    );
+  });
+  const res = await Promise.all(promiseArray);
```
ChaituVR, line 76: Hi @PerrorOne Can you use multicall here plz, one call per one address like this won't scale for bigger  voting,

Here is an example of how to call multicall,
https://github.com/snapshot-labs/snapshot-strategies/blob/master/src/strategies/protofi-erc721-tier-weighted/index.ts#L28-L36

src/strategies/evolution-land-dao/index.ts
```diff
@@ -0,0 +1,130 @@
...
+        continue;
+      }
+      multiCalls.push(
+        call(getProvider(network), abi, [
+          strategy.address,
+          'balanceOfEVO',
+          [address, k, currentRow]
+        ])
+      );
+    }
```
ChaituVR, line 108: Also here

ChaituVR (changes requested):

src/strategies/evolution-land-dao/index.ts
PerrorOne: Because the `multicall` query uses a block, which may cause the contract function to take too long and cause an 'execution reversed' error, so I used `call`. Is there a problem with this treatment?  thank you. @ChaituVR 

.

ChaituVR: Yes @PerrorOne Can't send one request per one address, https://github.com/snapshot-labs/snapshot-strategies#checklist-for-adding-a-new-strategy

ChaituVR: @PerrorOne

PR #559, opened 2022-04-22
Contract call weighted

rubenwrightus:
contract-call-weighted

just contract-call with a multiplier


ChaituVR:
Hey @rubenwrightus You can use https://snapshot.org/#/strategy/math strategy here and use the `multiply` operation
Let me know if it doesn't help :) closing this for now. feel free to reopen if needed

PR #565, opened 2022-04-26
Fix `ve-balance-of-at` strategy out of gas

spicysquid168:
Fixes `ve-balance-of-at` strategy.

- Chunking addresses then do multicall instead of multicall all addresses at once which leads to `out of gas`/`execute revert` issue.


src/strategies/ve-balance-of-at/index.ts
```diff
@@ -21,10 +21,16 @@ export async function strategy(
...
+  let result: Record<string, BigNumberish> = {};
+
+  const chunkSize = 200;
+  for (let i = 0; i < addresses.length; i += chunkSize) {
+    const chunk = addresses.slice(i, i + chunkSize);
+    chunk.forEach((address) =>
+      multi.call(address, options.address, 'balanceOfAt', [address, blockTag])
+    );
+    result = await multi.execute();
+  }
```
ChaituVR, line 33: Hey @spicysquid168 Maybe try using the `pagination` strategy, It is not recommended to add pagination with 200 addresses inside the strategy

ChaituVR (changes requested):

src/strategies/ve-balance-of-at/index.ts
spicysquid168: Hey @ChaituVR, thanks for the suggestion. We will go with `pagination` strategy 👍

PR #567, opened 2022-04-27
[erc721-with-tokenid-range-weights-simple] Erc721 with tokenid range weights simple v2

FeSens:
-  Improves performance for the `erc721-with-tokenid-range-weights-simple` strategy, reducing external calls, and allowing more than 5 ranges to be used.
-  Add's an optional `defaultWeight` parameter to `erc721-with-tokenid-range-weights-simple` strategy

### DIFF
--- a/src/strategies/erc721-with-tokenid-range-weights-simple/index.ts
+++ b/src/strategies/erc721-with-tokenid-range-weights-simple/index.ts
@@ -1,14 +1,27 @@
+import { strategy as erc721WithMultiplier } from '../erc721-with-multiplier';
 import { multicall } from '../../utils';
+import { WeightRange } from './types';
 
-export const author = 'gregegan';
-export const version = '0.1.0';
+export const author = 'FeSens';
+export const version = '0.2.0';
 
 const abi = [
-  'function ownerOf(uint256 tokenId) public view returns (address owner)'
+  'function ownerOf(uint256 tokenId) public view returns (address owner)',
+  'function balanceOf(address account) external view returns (uint256)',
 ];
 
-const range = (start, end, step) =>
+const range = (start, end, step) => 
   Array.from({ length: (end - start) / step + 1 }, (_, i) => start + i * step);
+
+const flattenTokenIdWeightRanges = (tokenIdWeightRanges: WeightRange[]): {ids: number[], weights: number[]} => {
+  const ranges = tokenIdWeightRanges.map((tokenIdWeightRange) => {
+    const { start, end, weight } = tokenIdWeightRange;
+    const ids = range(start, end, 1);
+    const weights = Array(ids.length).fill(weight);
+    return { ids, weights }
+  })
+  return ranges.reduce((prev, curr) => ({ ids: [...prev.ids, ...curr.ids], weights: [...prev.weights, ...curr.weights] }))
+}
 
 export async function strategy(
   space,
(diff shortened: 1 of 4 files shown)


src/strategies/erc721-with-tokenid-range-weights-simple/index.ts
```diff
@@ -1,15 +1,35 @@
...
 
-export const author = 'gregegan';
-export const version = '0.1.0';
+export const author = 'FeSens';
+export const version = '0.2.0';
 
 const abi = [
-  'function ownerOf(uint256 tokenId) public view returns (address owner)'
+  'function ownerOf(uint256 tokenId) public view returns (address owner)',
+  'function balanceOf(address account) external view returns (uint256)'
```
ChaituVR, line 10: Why balanceOf ? can't see where you used it?

src/strategies/erc721-with-tokenid-range-weights-simple/index.ts
```diff
@@ -19,48 +39,66 @@ export async function strategy(
...
+
+  if (defaultWeight)
+    erc721WeightedBalance = await erc721WithMultiplier(
+      space,
+      network,
+      provider,
+      addresses,
+      { ...options, multiplier: defaultWeight },
+      snapshot
+    );
```
ChaituVR, line 99: If defaultWeight exist, you are adding erc721 balance to range? better to use erc721WithMultiplier in your space seperately, result of voting power will be sum of these two strategies

ChaituVR (changes requested):

src/strategies/erc721-with-tokenid-range-weights-simple/index.ts
FeSens: Ops, my bad. It was used before I used `erc721WithMultiplier` inside this strategy.

src/strategies/erc721-with-tokenid-range-weights-simple/index.ts
```diff
@@ -19,48 +38,66 @@ export async function strategy(
...
+  const getCustomRangeBalance = async (): Promise<{
+    [address: string]: number;
+  }> => {
+    const { ids, weights } = flattenTokenIdWeightRanges(tokenIdWeightRanges);
+    const contractCalls = ids.map((id: number) => [
+      options.address,
+      'ownerOf',
+      [id]
+    ]) as [string, string, [number]][];
+    const customRangeResponse = await getDataFromBlockChain(contractCalls);
```
ChaituVR, line 78: I am worried if this can create some problems while passing all the ids at once because previously this strategy was calling one range at once

ChaituVR (changes requested):

src/strategies/erc721-with-tokenid-range-weights-simple/index.ts
ChaituVR: maybe we can create a new strategy? with some other easy name like `erc721-tokenid-ranges-with-weights`

FeSens: Yeah, this was on purpose. It's doing the same thing, just with fewer external calls and less time-consuming.

Before, the strategy did one external call for each range (limiting it to 5 ranges because Snapshot only allows 5 external calls) and only applied the wights in the final step.
Now, the strategy concatenates (flattened) the ranges in one single vector and creates a matching weights vector with `flattenTokenIdWeightRanges` to remove this limitation. It applies the weights to each range with a pairwise multiplication of the vectors in `multiplyOccurrencesByWeights`, before all other steps.

There was also one filter operation for every address and range, which led to an O(n^2) time complexity, limiting the maximum range to 7000. 
Now, the IDs are filtered once, and then the strategy counts the occurrences of each address, bringing the time complexity to O(n), improving performance.

FeSens: > maybe we can create a new strategy? with some other easy name like `erc721-tokenid-ranges-with-weights`

I can, but I've tried to not do that to mitigate clutter. It's basically the same strategy, it should be fully compliant, without the previous limitations.

Maybe @gregegan can give it a review to see if I've missed something.

ChaituVR (approved):

ChaituVR:
@FeSens as discussed in discord, reverting this for now, would be great if you can create a new PR with the fix :)

PR #569, opened 2022-04-28
Add erc1155-balance-of-with-thresholds strategy

RosebuDAO:
Context: https://github.com/snapshot-labs/snapshot/issues/2263

### DIFF
--- a/src/strategies/erc1155-balance-of-with-thresholds/index.ts
+++ b/src/strategies/erc1155-balance-of-with-thresholds/index.ts
@@ -1,0 +1,47 @@
+import { formatUnits } from '@ethersproject/units';
+import { multicall } from '../../utils';
+
+export const author = 'RosebuDAO';
+export const version = '0.1.0';
+
+const abi = [
+  'function balanceOf(address owner, uint256 id) view returns (uint256)'
+];
+
+export async function strategy(
+  space,
+  network,
+  provider,
+  addresses,
+  options,
+  snapshot
+) {
+  const thresholds = options.thresholds || [{ threshold: 1, votes: 1 }];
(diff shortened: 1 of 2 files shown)


src/strategies/erc1155-balance-of-with-thresholds/index.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  const blockTag = typeof snapshot === 'number' ? snapshot : 'latest';
+  const response = await multicall(
+    network,
+    provider,
+    abi,
+    addresses.map((address: any) => [
+      options.address,
+      'balanceOf',
+      [address, options.tokenId]
```
ChaituVR, line 35: Hey @RosebuDAO You looking to count voting power based on one tokenID ?? 

I am thinking like it would be better to just make it `threshold` to accept any strategy and add a threshold, so others can use it, maybe[ like here ](https://github.com/snapshot-labs/snapshot-strategies/blob/master/src/strategies/pagination/index.ts)

or we should add a new operation on https://snapshot.org/#/strategy/math strategy

ChaituVR (changes requested):

src/strategies/erc1155-balance-of-with-thresholds/index.ts
RosebuDAO: > Hey @RosebuDAO You looking to count voting power based on one tokenID ??

yes


I agree adapting threshold to accept any strategy is a better approach. Unfortunately I'm not very good with TypeScript.

ChaituVR:
Hi @RosebuDAO I have created this strategy https://snapshot.org/#/strategy/thresholds :) let me know if it doesn't work

PR #570, opened 2022-04-28
updated address casing [dogs-unchained]

dogsunchained:
Prioritizes casing passed in through [addresses]

### DIFF
--- a/src/strategies/dogs-unchained/index.ts
+++ b/src/strategies/dogs-unchained/index.ts
@@ -1,3 +1,5 @@
+import { 
+  ress } from '@ethersproject/address';
 import { formatUnits } from '@ethersproject/units';
 import { multicall, subgraphRequest } from '../../utils';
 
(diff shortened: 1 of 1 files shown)


src/strategies/dogs-unchained/index.ts
```diff
@@ -108,46 +107,61 @@ export async function strategy(
...
   });
 
+  const convertAddress = (address) => {
+    for (const i in addresses) {
+      if (addresses[i].toLowerCase() == address.toLowerCase())
+        return addresses[i];
+    }
+    return address;
+  };
+
```
ChaituVR, line 117: Hey @dogsunchained How about Checksum all addresses everywhere using https://docs.ethers.io/v5/api/utils/address/#utils-getAddress
Like here 
https://github.com/snapshot-labs/snapshot-strategies/blob/master/src/strategies/gno/index.ts#L37

ChaituVR (changes requested):

src/strategies/dogs-unchained/index.ts
dogsunchained: just made the edit!

ChaituVR (approved):

PR #578, opened 2022-05-05
New custom strategy [gatenet-total-staked]

usagar80:
We at [GATENet](https://gatenet.io/) have our own custom staking platform. This strategy is based on a user's total staked tokens, we use that figure to calculate quorum.


src/strategies/gatenet-total-staked/index.ts
```diff
@@ -0,0 +1,242 @@
...
+      amount: true,
+      user: true,
+      time: true,
+      unStakeIndex: true
+    }
+  };
+  const transactionsList = await subgraphRequest(
+    'https://api.studio.thegraph.com/query/17252/gatenet-cvm/v0.6.1',
+    query
+  );
```
ChaituVR, line 92: Hey, @usagar80 is there a way to pass a block number into this subgraph request? else the voting power may keep changing when people vote

ChaituVR (changes requested):

src/strategies/gatenet-total-staked/index.ts
usagar80: Hey @ChaituVR ,
No there is no way to do this. I agree it can very voting power. i.e. today I have 5K Staked so my Quorum would be 5K. But If I stake more 15K then my quorum would be 15K.

ChaituVR: you tried passing block to args? maybe is there a way to pass timestamp ?

usagar80: Hey @ChaituVR , thanks for your revert!
Yes, we can pass timestamp and it's related comparison in `_arg`. But I don't see any reason for that. May be if you can help me understand more (as an example).

ChaituVR: Hey @usagar80 so the problem here is we are not fetching staked info based on block number, so when a proposal is open, the user can stake and get the voting power required, This can affect whole proposal voting,

usagar80: @ChaituVR , as per your suggestion here and on Discord I have implement changes to the strategy. Please review and let me know.

ChaituVR:
@usagar80 Looks like this one is taking more than 20 secs to resolve the strategy, can you try to run this command in local to see whats wrong
```
yarn test --strategy=gatenet-total-staked --more=500
```

PR #579, opened 2022-05-06
Add strategy for indexed token balances in a vault

0xAurelius:
- Adds a new strategy to allow for stakers of an indexed token like wsKLIMA to be given voting power for staking in a vault contract

Ran all tests locally and they pass and provide expected output for provided `addresses`

image


0xAurelius:
Nevermind this is not needed, just use `math`

PR #580, opened 2022-05-06
Add strategy for ERC721 contract holding an ERC20 token balance

0xAurelius:
- Adds a strategy for granting voting power to holder of a 1/1 ERC721 proportional to an embedded ERC20 token balance held in the ERC721 contract

image

### DIFF
--- a/src/strategies/erc721-balance-erc20/index.ts
+++ b/src/strategies/erc721-balance-erc20/index.ts
@@ -1,0 +1,44 @@
+import { formatUnits } from '@ethersproject/units';
+import { call, multicall } from '../../utils';
+
+export const author = '0xAurelius';
+export const version = '0.0.1';
+
+const abi = [
+  'function balanceOf(address account) external view returns (uint256)'
+];
+
+export async function strategy(
+  space,
+  network,
+  provider,
+  addresses,
+  options,
+  snapshot
+) {
+  const blockTag = typeof snapshot === 'number' ? snapshot : 'latest';
+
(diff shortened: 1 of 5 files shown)


src/strategies/erc721-balance-erc20/index.ts
```diff
@@ -0,0 +1,44 @@
...
+    network,
+    provider,
+    abi,
+    addresses.map((address: any) => [options.address, 'balanceOf', [address]]),
+    { blockTag }
+  );
+  return Object.fromEntries(
+    response.map((value, i) => [
+      addresses[i],
+      parseFloat(formatUnits(value.toString(), 0)) * erc20Balance
```
ChaituVR, line 41: Hey, @0xAurelius I think you can use the `math` strategy here and multiply operation,

Try passing erc721 and contract call into it, there was a similar strategy here https://github.com/snapshot-labs/snapshot-strategies/blob/master/src/strategies/erc20-balance-of-contract-multiplier/index.ts but no one use it,

ChaituVR (changes requested):

src/strategies/erc721-balance-erc20/index.ts
0xAurelius: Ah ok thanks, I did look at that one but I couldn't figure out if it would do what I wanted

Will give it a shot 🤞

ChaituVR:
Closing this as stale, feel free to reopen if needed

PR #590, opened 2022-05-11
Adds: citydao-square-root 0.0.1 [citydao-square-root]

will-holley:
#### Summary

This PR adds CityDAO's Square Root Strategy.  More info can be found in its [README](https://github.com/721labs/snapshot-strategies/blob/citydao-square-root/src/strategies/citydao-square-root/README.md).

#### Adds

- Strategy `citydao-square-root@0.0.1`

#### Modifies

- None

### DIFF
--- a/src/strategies/citydao-square-root/examples.json
+++ b/src/strategies/citydao-square-root/examples.json
@@ -1,0 +1,21 @@
+[
+  {
+    "name": "Example",
+    "strategy": {
+      "name": "citydao-square-root",
+      "params": {
+        "symbol": "CITIZEN",
+        "address": "0x7EeF591A6CC0403b9652E98E88476fe1bF31dDeb",
+        "tokenId": 42,
+        "decimals": 0
+      }
+    },
+    "network": "1",
+    "addresses": [
+      "0xafa46468De1D6f1ab77DEFAe5F7657467911182d",
+      "0x944Fd37C2a46A3ad6B5EE11f7e26035Ed2F1E4FE",
+      "0x4c55C41Bd839B3552fb2AbecaCFdF4a5D2879Cb9"
+    ],
+    "snapshot": 14616178
+  }
+]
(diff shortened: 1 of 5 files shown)


src/strategies/citydao-square-root/index.ts
```diff
@@ -0,0 +1,64 @@
...
+    snapshot
+  );
+
+  // Support Plural Voting
+  const magnitude = params.voiceCredits || 1;
+
+  // Update in place, rounding down.
+  for (const address in scores) {
+    scores[address] = Math.floor(Math.sqrt(scores[address] * magnitude));
+  }
```
ChaituVR, line 59: Hi @will-holley I think you can use [math ](https://snapshot.org/#/strategy/math)strategy here and use following operation:

<html>
<body>


square-root
--


<br class="Apple-interchange-newline">
</body>
</html>

ChaituVR (changes requested):

src/strategies/citydao-square-root/index.ts
will-holley: I could; it would be _less_ computationally efficient, add additional code complexity, and require an additional iteration to map the `magnitude`.  I would strongly prefer to leave as is for these reasons.

ChaituVR (approved):

src/strategies/citydao-square-root/examples.json
```diff
@@ -0,0 +1,41 @@
+[
+  {
+    "name": "Example",
+    "strategy": {
+      "name": "citydao-square-root",
+      "params": {
+        "symbol": "CITIZEN",
+        "address": "0x7EeF591A6CC0403b9652E98E88476fe1bF31dDeb",
+        "tokenId": 42,
```
mdnorman, line 9: @will-holley can this strategy be used with multiple different tokensIDs? For instance, CityDAO has 3 different tokens within the same ERC1155 contract, and the total of all of these tokens should be used.

I didn't see how this would be possible with the given implementation, but I could have missed something.

PR #602, opened 2022-05-16
Add recusal list strategy [recusal-list]

bshyong:
Fixes # . n/a

- Adds Recusal List strategy. This is strategy for disallowing certain address from voting due to conflict of interest or other reasons for recusal.

### DIFF
--- a/src/strategies/recusal-list/README.md
+++ b/src/strategies/recusal-list/README.md
@@ -1,0 +1,14 @@
+# erc20-balance-of
+
+This is strategy for disallowing certain address from voting due to conflict of interest or other reasons for recusal.
+
+Below is an example of parameters. The address list renotes which addresses to restrict.
+
+```json
+{
+  "addresses": [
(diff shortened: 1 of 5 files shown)


src/strategies/recusal-list/README.md
```diff
@@ -0,0 +1,14 @@
+# erc20-balance-of
```
ChaituVR, line 1: erc20-balance-of?

ChaituVR (changes requested):
Maybe we can rename this to the `blacklist` strategy? any thoughts?

bshyong:
> Maybe we can rename this to the `blacklist` strategy? any thoughts?

I thought about that language but decided on Recusal because a blacklist has a negative connotation and is less descriptive - Recusal explicitly refers to addresses being restricted from voting with regards in governance.

ChaituVR (approved):

PR #605, opened 2022-05-18
add strategy [erc721-with-tokenid-list-weighted]

paste:
This strategy takes token IDs instead of a range, and each list of token IDs is weighted. This is useful for ENS token IDs which can't be used with a range.

### DIFF
--- a/src/strategies/erc721-with-tokenid-list-weighted/examples.json
+++ b/src/strategies/erc721-with-tokenid-list-weighted/examples.json
@@ -1,0 +1,23 @@
+[
+  {
+    "name": "Example query",
+    "strategy": {
+      "name": "erc721-with-tokenid-list-weighted",
+      "params": {
+        "address": "0x30cDAc3871c41a63767247C8D1a2dE59f5714e78",
+        "symbol": "Reaper(s)",
+        "tokenIdWeightLists": [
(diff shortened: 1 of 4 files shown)


src/strategies/erc721-with-tokenid-list-weighted/examples.json
```diff
@@ -0,0 +1,23 @@
...
+    "name": "Example query",
+    "strategy": {
+      "name": "erc721-with-tokenid-list-weighted",
+      "params": {
+        "address": "0x30cDAc3871c41a63767247C8D1a2dE59f5714e78",
+        "symbol": "Reaper(s)",
+        "tokenIdWeightLists": [
+          { "weight": 1, "ids": ["2112", "2151", "2871"] },
+          { "weight": 10, "ids": ["3221", "3587", "4179"] }
+        ]
```
ChaituVR, line 12: Hi @paste I think there should limit for `tokenIdWeightLists` and `ids` else, people may pass huge arrays and there could be memory issues on the server.

ChaituVR, line 12: Hi @paste I think there should limit for `tokenIdWeightLists` and `ids` else, people may pass huge arrays and there could be memory issues on the server.

paste:
this is a more generic version of #607, but fails for 10K club due to large token ID list.

PR #607, opened 2022-05-18
Add strategy for ENS 10k Club [ens-10k-club]

paste:
This strategy is for 10K Club members, holders of ENS names 000.eth - 9999.eth. This script queries The Graph for all ENS names owned by voter and checks for 10K Club names. By default, 999 Club members' votes are weighted 10x.

### DIFF
--- a/src/strategies/ens-10k-club/index.ts
+++ b/src/strategies/ens-10k-club/index.ts
@@ -1,0 +1,83 @@
+import { subgraphRequest } from '../../utils';
+import { getAddress } from '@ethersproject/address';
+
+const ENS_SUBGRAPH_URL = {
+  '1': 'https://api.thegraph.com/subgraphs/name/ensdomains/ens',
+  '3': 'https://api.thegraph.com/subgraphs/name/ensdomains/ensropsten',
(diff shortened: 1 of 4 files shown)


src/strategies/ens-10k-club/index.ts
```diff
@@ -0,0 +1,125 @@
+import { subgraphRequest } from '../../utils';
+import { BigNumber } from '@ethersproject/bignumber';
+import { getAddress } from '@ethersproject/address';
+
+import { TOKEN_IDS_999_CLUB } from './999-club-token-ids';
+import { TOKEN_IDS_10K_CLUB } from './10k-club-token-ids';
```
ChaituVR, line 6: Hey, what are these ids? these should not be on this repo., maybe should be on an IPFS file and fetch it when required

ChaituVR (changes requested):

paste:
Hi @ChaituVR, rewritten and much simpler. I apologize for the attitude!

ChaituVR (approved):
Thanks for the changes @paste Its ok i can understand :) 

I feel like we can do like `parseInt(labelName)` something like that, but not sure if it is possible. Maybe you can try it for next version, merging this for now as this looks good

PR #622, opened 2022-05-31
[razor-network-voting] update strategy

rajkharvar:
- Update subgraph endpoint
- Update strategy to fetch score
- Update addresses in `examples.json`


ChaituVR:
@adi44 Can you plz check and confirm this?

src/strategies/razor-network-voting/index.ts
```diff
@@ -61,38 +62,38 @@ export async function strategy(
     }
   };
 
-  if (snapshot !== 'latest') {
-    // @ts-ignore
-    params.delegators.__args.block = { number: snapshot };
-  }
-
```
ChaituVR, line 68: This should not be removed @rajkharvar The results should be depended on the block number, else the results can be manuplated

ChaituVR (changes requested):

src/strategies/razor-network-voting/index.ts
rajkharvar: Ok. will update that.

rajkharvar: @ChaituVR I have made the required changes.

ChaituVR (approved):

PR #634, opened 2022-06-06
add vs-balance [eth-balance]

vchain007:
vs decimal get from smartcontract is 6 on vision
- 
- 
-


ChaituVR:
Vs is like eth? Default on vision chain?

I think better to modify  eth-balance strategy itself, try passing a optional parameter decimals and default value as 18

vchain007:
> Vs is like eth? Default on vision chain?
> 
> I think better to modify eth-balance strategy itself, try passing a optional parameter decimals and default value as 18
yes , vs is native token on vision chain .but it's dicimal is 6  . we adapter eth_getbalance interface so this interface return 
dicimal 18 . For example
user A have 1000 VS , on vision chain A have 1000*10e6 . eth_getbalance return 1000* 10e18 just for adaptering .multicall query smart contract return 1000*10e6

vchain007:
update . plz review

src/strategies/eth-balance/index.ts
```diff
@@ -29,10 +29,12 @@ export async function strategy(
     ]),
     { blockTag }
   );
+  const decimals = decimals in options ? options.decimals : 18 ;
```
ChaituVR, line 32:
```suggestion
  const decimals = options.decimals || 18 ;
```

ChaituVR (changes requested):

vchain007:
when will this strategy effect online ,sir

PR #640, opened 2022-06-08
Enable passing subgraph endpoints as parameters and update subgraph endpoints [volt-voting-power]

philipappiah:
Fixes # .

- Enable passing subgraph endpoints as parameters
- Update subgraph endpoints


src/strategies/volt-voting-power/index.ts
```diff
@@ -7,17 +7,17 @@ import { subgraphRequest } from '../../utils';
 export const author = 'philipappiah';
 export const version = '0.1.0';
```
ChaituVR, line 8:
```suggestion
export const version = '0.1.1';
```

ChaituVR (changes requested):

PR #644, opened 2022-06-09
minto-balance-all changes

semdestroyer:
- We had difficulties with multicall and we changed multicall request to fetch to work through our api
- 
-


ChaituVR (changes requested):
Hi @semdestroyer i think you can use https://snapshot.org/#/strategy/api-post strategy in this case, please try and let me know :)

PR #651, opened 2022-06-13
Add rocketpool node operator strategies

nickdoherty:
These are the strategies Rocket Pool are planning to use for their snapshot space.


ChaituVR (changes requested):
Hi @nickdoherty Can you create 4 PRs for these four strategies separately plz? Easy to review and integrate,
Also, I feel some of these strategies can be achieved by `contract-call` + math strategies, can you try and let me know if it works,
and remove changes from yarn.lock

nickdoherty:
No problem, will close and open up individual.

PR #653, opened 2022-06-14
erc721-with-balance

zhoug0x:
gm!

It doesn't seem like there is a strategy available for ERC721s without the token balance weighting (1 holder, 1 vote). I followed the lead on `erc20-with-balance` to create this one.

Cheers!

- Add the `erc721-with-balance` strategy

### DIFF
--- a/src/strategies/erc721-with-balance/README.md
+++ b/src/strategies/erc721-with-balance/README.md
@@ -1,0 +1,25 @@
+# erc721-with-balance
+
+This strategy returns a singular vote based on a minimum NFT holding threshold (default 1)
+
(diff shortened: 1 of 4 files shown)


ChaituVR:
Hey @zhoug0 Thanks for the PR, we are aware of this, we are trying to find ways to rename existing strategies, till then erc20-with-balance should work for erc721 too 😄

zhoug0x:
> Hey @zhoug0 Thanks for the PR, we are aware of this, we are trying to find ways to rename existing strategies, till then erc20-with-balance should work for erc721 too 😄

Sounds good, thanks @ChaituVR !

PR #658, opened 2022-06-17
[qidao] Add QiPowah Multichain voting strategy

royalaid:
Adds the QiPowah Multichain voting strategy allow us to roll up QiPowah across all snapshot supported chains.
image


src/strategies/qipowah/README.md
```diff
@@ -0,0 +1,134 @@
+# multichain
```
ChaituVR, line 1:
```suggestion
# qipowah
```

src/strategies/qipowah/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+If you want to calculate the balance from various chains like Ethereum, Binance smart chain, polygon etc. and use them for voting using various strategies, you can do it by using a strategy called “multichain strategy”. This allows cross chain voting in which multiple chains can be used together to calculate the voting power.
```
ChaituVR, line 3: This readme is for multichain strategy please change according to your strategy plz

src/strategies/qipowah/index.ts
```diff
@@ -0,0 +1,64 @@
...
+            }
+          }
+        },
+        blocks[ERC20.network]
+      )
+    );
+  }
+
+  const results = await Promise.all(promises);
+  return results.reduce((finalResults: any, strategyResult: any) => {
```
ChaituVR, line 55: Hey @royalaid  i think you are just adding 8 erc20 tokens balances right, you can set 8 strategies in your space and output will be sum of them

ChaituVR (changes requested):

src/strategies/qipowah/index.ts
publu: We've been having issues with nested queries because some of our balance contracts timeout. This strategy helps declutter the configuration so we can quickly update it

ChaituVR: There is actually a limit on network calls sent from a strategy, https://github.com/snapshot-labs/snapshot-strategies#checklist-for-adding-a-new-strategy 

Also, there is a strategy limit on a snapshot space for scaling reasons, more users may show this strategy as a reason to group their strategies into one strategy

royalaid: That is fine, we can roll all this into a nest configuration but I feel we may run up against hard limits just based on the number of chains we need to connect too as we continue to expand.

ChaituVR: @bonustrack Any thoughts on strategy limits?

bonustrack: We shouldn't have strategy that allow to run a unlimited number of strategies, "multichain" strategy was an exception but now that we support multichain natively I recommend to use normal strategy with custom network. For the strategy that require pagination we can set custom handler in our side, you just need to let us know which one is it.

royalaid: Recommendation is to move away from these strategies and to a basic `erc-20-balanceof` using the native multichain support? Ideal goal is to roll up all the different chains vote counts into a single number (I would geuss this is the default case for most projects) and support said final count delegation. 

If that is possible Is there a good resource for migrating from the multichain strategy to natively using snapshots multichain? My google-fu is failing me 😅.

bonustrack: @royalaid Yes that's right, you can try adding a voting strategy you will see now there is a select input for network:
image

royalaid: Right but does that allow us to roll it up into a single count like the multichain strategy does currently? It seems like it just support allowing votes from multiple chains but not collecting them under one sum. I am sure this just me not understanding the feature though.

ChaituVR: @royalaid Result will be sum of all strategies (even if they are on different networks)

PR #663, opened 2022-06-20
Easy way to test strategies

ChaituVR:
Right now we have to run `yarn build` before running `ts-node test/scores.ts`  It is hard to debug a strategy, better if `ts-node test/scores.ts` can pick latest changes with out build, this change fixes it

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,19 +1,23 @@
 # Snapshot strategies
 
-### Development
+## Development
 
-#### Install dependencies
+### Install dependencies
+
 ```bash
 yarn
 ```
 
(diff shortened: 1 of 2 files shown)


bonustrack:
It's doesn't look easy to use test/scores.ts file, require lot of change to do a test. I think we should instead refactor code so we can use the usual test script without build, similar than what we do on https://github.com/snapshot-labs/pineapple.js

ChaituVR:
Thanks, @bonustrack I Updated the spec file as well

bonustrack (approved):

PR #666, opened 2022-06-21
Strategy for SDL and voting power

nguyenaiden:
Added strategies for:
-  SDL + locked SDL balance
- veSDL voting power


src/strategies/saddle-vesdl-voting-power/index.ts
```diff
@@ -0,0 +1,34 @@
...
+  addresses.forEach((address) =>
+    multi.call(address, options.address, 'getPriorVotes', [address, blockTag])
+  );
+  const result: Record<string, BigNumberish> = await multi.execute();
+
+  return Object.fromEntries(
+    Object.entries(result).map(([address, balance]) => [
+      address,
+      parseFloat(formatUnits(balance, options.decimals))
+    ])
```
ChaituVR, line 32: Hey @nguyenaiden For this strategy, can you try the `contract-call` strategy and see if it works? Thanks :)

ChaituVR (changes requested):
Hey I think you can use `erc20-balance-of` strategy and two `contract-call` strategies, one for locked and one for getPriorVotes

Result will be sum of these three strategies

src/strategies/saddle-vesdl-voting-power/index.ts
nguyenaiden: @ChaituVR Could you explain this further? I'm not sure what you meant by trying the `contract-call` strategy.

ChaituVR: I am this strategy https://snapshot.org/#/strategy/contract-call , you can directly call a method from a contract, try playing with it in the playground here https://snapshot.org/#/playground/contract-call

PR #671, opened 2022-06-24
version 0.0.1 [sd-boost-twavp] 

clement-ux:
Fixes # .

- Add sd-boost-twavp strategy 
--> This strategy will be use by StakeDAO to vote with sdToken with Time Weigthed Averaged Voting Power system and veSDT voting boost.


pierremarsotlyon1 (approved):

src/strategies/sd-boost-twavp/index.ts
```diff
@@ -0,0 +1,186 @@
...
+export async function strategy(
+  space,
+  network,
+  provider,
+  addresses,
+  options,
+  snapshot
+): Promise<Record<string, number>> {
+  // Maximum of 5 multicall
+  expect(options.sampleStep).toBeLessThan(6);
```
ChaituVR, line 31: Hi @clement-ux 

I get this error when I run this without jest.
```
ReferenceError: expect is not defined
```


should be like this:
https://github.com/snapshot-labs/snapshot-strategies/blob/master/src/strategies/erc721-with-tokenid-range-weights-simple/index.ts#L61

clement-ux: Changes done :)

ChaituVR (approved):

PR #679, opened 2022-06-29
Feat/apeswap strategy [apeswap]

ixm-759:
Add strategy for apeswap


src/strategies/apeswap/index.ts
```diff
@@ -0,0 +1,93 @@
...
+import { Multicaller } from '../../utils';
+import examplesFile from './examples.json';
+import fetch from 'cross-fetch';
+
+export const author = 'ApeSwapFinance';
+export const version = '0.0.1';
+export const examples = examplesFile;
+
+const POOL_URL =
+  'https://raw.githubusercontent.com/ApeSwapFinance/apeswap-yields/main/config/pools.json';
```
ChaituVR, line 12: I think better to use a link with commit id instead,

Something like this https://github.com/ApeSwapFinance/apeswap-yields/blob/8a834527c0f3f1b9d6f21686fc62e71294148fcb/config/pools.json

So it won't change the User's voting power during the proposal, or maybe you can pass this URL from parameters, where you can change it from time to time before the proposal start

ChaituVR (changes requested):

src/strategies/apeswap/index.ts
ixm-759: Hello, regarding the parameters, are they the ones that go in the `example.json` file?

Is that parameter going to be shown when I generate a new proposal?

Do we have to add that url in each new proposal?

ChaituVR: Hi @ixm-759 `example.json` file is to test strategy but on snapshot.org you can pass these params from space settings, If the file on GitHub changes, you have to change the space settings and a new proposal will use that,

src/strategies/apeswap/index.ts
```diff
@@ -0,0 +1,60 @@
+import { BigNumber, BigNumberish } from '@ethersproject/bignumber';
+import { formatUnits } from '@ethersproject/units';
+import { Multicaller } from '../../utils';
+import examplesFile from './examples.json';
+
+export const author = 'ApeSwapFinance';
+export const version = '0.0.1';
+export const examples = examplesFile;
```
ChaituVR, line 8:
```suggestion

export const author = 'ApeSwapFinance';
export const version = '0.0.1';
```

ChaituVR (changes requested):

ChaituVR (approved):

PR #683, opened 2022-06-30
[gooddollar-multichain] New strategy

sirpy:
- Add a multichain strategi that selects the chain and value with max voting power


ChaituVR:
Hi @sirpy we are actually not suggesting multichain strategy to anyone because users can select differnet network for each strategy inside your space settings, which is equal to multicahin strategy, for max voting power i think you can use [math strategy](https://snapshot.org/#/strategy/math) and use `min` operation, let me know if it doesn't work :) Thanks

sirpy:
@ChaituVR are you sure? Because on the tests it appears as single strategy  and we plan it to be the only strategy defined in the space. I'll also look next week at the math strategy, does it support multiple chains?

ChaituVR:
Hi again @sirpy  yep, you can select multiple strategies in space, example: https://snapshot.org/#/ens.eth/settings try clicking a strategy, you can select different network for each strategy, total voting power will be sum of these strategies

And yes match strategy should work on different networks this way :)

sirpy:
@ChaituVR I know, but as I said i'm not planning on choosing multiple strategies only 1 strategy which is the multichain-max strategy.

ChaituVR:
But why? Is there any usecase to have single strategy.?

sirpy:
@ChaituVR i'm not following you, inside the multichain-max you can define multiple strategies but these strategies are not part of the "multiple" strategies you can define in the DAO settings, they are completely internal to the multichain-max strategy

sirpy:
@ChaituVR i've looked into the math strategy, and I had to make some modifications to it so that it can support calculating max from multiple chains.
i've committed to this PR also 77bb89a

sirpy:
@ChaituVR can you take a look at this again?

ChaituVR:
Hi @sirpy The `math` strategy doesn't need these changes too, did you try going to your space settings? while selecting a strategy you can select its network too.

sirpy:
@ChaituVR i've tested my example (by putting it as the first one and running the tests) and 
without these changes it doesnt know how to read the correct block number of the fuse blockchain
this is obviously a must change because the snapshot time block number for fuse will be different from eth.

yes i know you can select a "single network" per strategy. but his is a "multi network" strategy.
this strategy reads information from multiple networks. without a code to find the block number for each network at the time of the snapshot it will not work.

happy for you to show me a working example configuration for voting power=Max(blockchainX at timeX, blockchainY at timeX)

sirpy:
@ChaituVR this is critical for us, even if you are right I dont see why not add a multichain-max strategy which is just a clone of multichain with `max` instead of `add`

ChaituVR:
@sirpy multichain strategy is really hard to compute, the easier way is to use multichain functionality on snapshot.org (as I mentioned above)

@bonustrack You think this can be merged?

sirpy:
@ChaituVR 
Then please show me a demo or share a configuration file of how to do a
a Max(blockchainX,blockchainY) strategy where the block used to fetch the voting power on each chain matches the snapshot timestamp. Because i have no idea how to implement it when you say multichain functionality on snapshot.org, and i've tried everything.

ChaituVR:
@sirpy What all networks you are looking to try? and what strategy you will pass into this? Maybe create a strategy that calls only your strategies and networks. make sure you pass this checklist (make sure the number of requests doesn't cross 5). https://github.com/snapshot-labs/snapshot-strategies#checklist-for-adding-a-new-strategy 

If it is this way, others can pass unlimited networks and strategies. which is no good for the server

sirpy:
@ChaituVR look i dont get it, there's already the `multichain` strategy that people can use/abuse
the only thing I added is an exact same strategy that instead of ADD does MAX.
Same for the `math` strategy, people can already feed it multiple sub strategies and "abuse" it.
If you want I can add hardcoded limit of 5 chains but that wouldn't solve the vulnerability of these other strategies.

@bonustrack can we have a second opinion here please?

ChaituVR:
> ook i don't get it, there's already the multichain strategy that people can use/abuse

Exactly 😢 It's hard to deprecate it, we are trying a few ways to remove it but it is troublesome, so we don't want to accept new strategies like that.

What is your exact requirement? what all networks and strategies you are looking to include in the `max`. I will talk to the team if they have a solution. Do you have multiple tokens on multiple chains? why is it important to consider max in those networks?

sirpy:
Our governance token is currently on Fuse+Ethereum.
The requirement is to consider the max balance of the gov token on any of the networks you hold it.
I dont get the question `why is it important`, because that's the way our governance logic is.

ChaituVR:
I believe in the future we can def. increase the number of calls sent from a strategy and also strategies limit :) (we are actively trying to scale the server code for this) till then would be great if you can write a strategy to get a max of these two networks

sirpy:
@ChaituVR ok i've removed my changes to that `math` strategy and changed the max strategy to
1. filter out any strategies not on networks 122 or 1
2. limit to two strategies

hope this can finally be merged
thanks

sirpy:
@ChaituVR 
Why have you closed this? i've modified it like you've asked...

sirpy:
@ChaituVR and on that note... can we have it merged please?

src/strategies/max-eth-fuse/README.md
```diff
@@ -0,0 +1,135 @@
...
+In order to provide multichain functionality, this strategy requires a way for calculating which block number should be used on additional chains: If a snapshot was created on block 125 on mainnet, it needs to find the timestamp for that block and go find which block number corresponds to that same timestamp on every other wanted chain. This way it can accurately represent an address' voting power at a given point in time. In order to do this, it supports 2 different mechanisms:
+
+- [DEFAULT] Querying a block subgraph. If a working block info subgraph is found for that chain, it can be passed into the strategy's options in the "graphs" object. This will allow the strategy to query that subgraph for the given chain and fetch the block number from there. An example of a graphs object could be:
+
+```json
+ "graphs": {
+    "56": "https://api.thegraph.com/subgraphs/name/apyvision/block-info",
+    "137": "https://api.thegraph.com/subgraphs/name/sameepsi/maticblocks"
+  }
+```
```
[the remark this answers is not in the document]
ChaituVR: maybe change strategy name to your space name-multichain than a common strategy, which should make things much easier :)

sirpy:
@ChaituVR done

src/strategies/index.ts
```diff
@@ -419,6 +419,7 @@ const strategies = {
   'erc1155-balance-of-cv': erc1155BalanceOfCv,
   'prepo-vesting': prepoVesting,
   multichain,
+  'multichain-max': multichainMax,
```
ChaituVR, line 422: Thanks, @sirpy Make sure it is changed here too :) also make sure to remove any changes not related to this strategy. example change in `src/strategies/math/index.ts` make sure to revert, everything else looks good

ChaituVR (changes requested):

src/strategies/index.ts
sirpy: done

sirpy: hmm still issues looking at it

sirpy: all done

ChaituVR (approved):

PR #701, opened 2022-07-07
dps-nft-strategy changed [dps-nft-strategy]

andreibadea20:
Fixes # .

-  network
- URL for request
- snapshot
- block number


src/strategies/dps-nft-strategy/index.ts
```diff
@@ -5,16 +5,15 @@ export const author = 'andreibadea20';
 export const version = '0.1.0';
```
ChaituVR, line 5:
```suggestion
export const version = '0.2.0';
```

ChaituVR (approved):

PR #703, opened 2022-07-07
Fix/ocean marketplace v4

idiom-bytes:
Fixes an issue where voting power was coming in as 50% of expected value due to 50/50 OCEAN/DT LP structure


ChaituVR (approved):

idiom-bytes:
omg @ChaituVR :heart:

PR #705, opened 2022-07-09
feat: add lrc subgraph balance strategy [lrc-l2-subgraph-balance-of]

shad-k:
- new strategy to get LRC balance from LoopringV2 subgraph


src/strategies/lrc-l2-subgraph-balance-of/index.ts
```diff
@@ -0,0 +1,91 @@
...
+
+export async function strategy(
+  space,
+  network,
+  provider,
+  addresses,
+  options,
+  snapshot
+): Promise<Record<string, number>> {
+  const LIMIT = 100;
```
ChaituVR, line 50: Hi, @shad-k Is there a chance to increase this limit? Should be fine in many cases but want to know if there is a possibility to make it at least 500

ChaituVR (changes requested):

src/strategies/lrc-l2-subgraph-balance-of/index.ts
shad-k: Hi @ChaituVR , ran some tests to check if we can increase the limit to 500. Changed the limit to 500.

ChaituVR (approved):

PR #706, opened 2022-07-09
Add getVp and getDelegations functions

bonustrack:


src/utils/vp.ts
```diff
@@ -0,0 +1,218 @@
+import { formatBytes32String } from '@ethersproject/strings';
+import { getAddress } from '@ethersproject/address';
+import {
+  getProvider,
+  getSnapshots,
+  Multicaller,
+  subgraphRequest
+} from '../utils';
+import _strategies from '../strategies';
+import subgraphs from './subgraphs.json';
```
ChaituVR, line 10: We can use this instead of separate JSON? https://github.com/snapshot-labs/snapshot-strategies/blob/master/src/utils.ts#L93

ChaituVR (changes requested):

.eslintignore
```diff
@@ -1 +0,0 @@
-dist
```
ChaituVR, line 1: Removing this causes issues when we run eslint, it considers files from `dist` folder, i think we should add back this file back

src/utils/vp.ts
bonustrack: We need a list of subgraph which support delegation, not just the list of all subgraphs

.eslintignore
bonustrack: I did like this instead https://github.com/snapshot-labs/snapshot-strategies/pull/706/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R22

src/utils/vp.ts
ChaituVR: @bonustrack These are just delegation subgraphs same as https://github.com/snapshot-labs/snapshot.js/blob/master/src/utils.ts#L18-L28 (Which are reused in snapshot UI) we moved them to snapshot.js to maintain these URLs from a single place,

.eslintignore
ChaituVR: Do you know how to solve these?
image

Or we should disable this rule?

PR #710, opened 2022-07-12
Add Forta Shares Voting Strategy [forta-shares]

emanuel-sol:
- Calculates voting power based on the number of EIP 1155 shares for staked FORT

### DIFF
--- a/src/strategies/forta-shares/index.ts
+++ b/src/strategies/forta-shares/index.ts
@@ -1,0 +1,83 @@
+import fetch from 'cross-fetch';
+import { multicall } from '../../utils';
+import { BigNumber, BigNumberish } from '@ethersproject/bignumber';
+import { formatUnits } from '@ethersproject/units';
+
+export const author = 'emanuel-sol';
+export const version = '0.0.1';
+
+const STAKING_CONTRACT = '0xd2863157539b1D11F39ce23fC4834B62082F6874';
+const abi = [
+  'function balanceOfBatch(address[] calldata accounts, uint256[] calldata ids) view returns (uint256[] memory)'
+];
+
+const calculateVotingPower = (userVotingPower, addresses, balances) => {
+  for (let i = 0; i < addresses.length; i++) {
+    userVotingPower[addresses[i]] = userVotingPower[addresses[i]] || BigNumber.from(0);
+
+    const balance = balances[i];
+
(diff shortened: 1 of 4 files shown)


src/strategies/forta-shares/index.ts
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  const response = await fetch("https://api.forta.network/stats/shares", {
+    method: 'POST',
+    headers: {
+      Accept: 'application/json',
+      'Content-Type': 'application/json'
+    },
+    body: JSON.stringify(params)
+  });
```
ChaituVR, line 46: This API doesn't consider block number? there is a possibility that people can change their VP during a proposal vote period, to give consistent results, better to consider `snapshot` param, 

Or maybe this API returns the same data every time?

ChaituVR (changes requested):

src/strategies/forta-shares/index.ts
ChaituVR: Everything else looks good :)

emanuel-sol: This API returns a list of all the shares ever owned by a particular address. We use the contract queries to determine the voting power of these at execution time (which i'm guessing is when a client signs in with a wallet to vote on our proposal? still a tad confused about how snapshot works under the hood)

ChaituVR: Hello  again @emanuel-sol So snapshot doesn't really consider time of user vote, `snapshot` block number is what decides the voting power of the user for the proposal, (this block number is taken at time of creation) 
For example if you consider erc20-balance-of strategy, voting power here is decided by balance held in a user's wallet the proposal's snapshot block number

so the API should ideally return the shares owned at a particular block time

stevenlanders: @ChaituVR the API returns all tokens ever owned up to the current second, but then we use the _snapshot_ block time to get the balance.  It starts a superset, then is reduced by the snapshot balances (will receive 0 balances for any tokens no longer owned).  This reduction is why the API doesn't influence the voting power in any problematic way.

ChaituVR (approved):

src/strategies/forta-shares/index.ts
ChaituVR: Thanks for explaining @stevenlanders :)

PR #714, opened 2022-07-16
Add rari-fuse strategy [rari-fuse]

MantisClone:
- Add a `rari-fuse` strategy that returns the voter's underlying collateral balances in a given Rari Fuse market (fToken).

### DIFF
--- a/src/strategies/index.ts
+++ b/src/strategies/index.ts
@@ -334,6 +334,7 @@ import * as solvVoucherClaimable from './solv-voucher-claimable';
 import * as h2o from './h2o';
 import * as dopamine from './dopamine';
 import * as lrcL2SubgraphBalanceOf from './lrc-l2-subgraph-balance-of';
+import * as rariFuse from './rari-fuse';
 
 const strategies = {
   'forta-shares': fortaShares,
@@ -671,7 +672,8 @@ const strategies = {
   apeswap,
   h2o,
   dopamine,
-  'lrc-l2-subgraph-balance-of': lrcL2SubgraphBalanceOf
+  'lrc-l2-subgraph-balance-of': lrcL2SubgraphBalanceOf,
+  'rari-fuse': rariFuse
 };
 
 Object.keys(strategies).forEach(function (strategyName) {
--- a/src/strategies/rari-fuse/README.md
+++ b/src/strategies/rari-fuse/README.md
@@ -1,0 +1,13 @@
+# erc20-balance-of
+
+This is the most common strategy, it returns the balances of the voters for a specific ERC20 token.
+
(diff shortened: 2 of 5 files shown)


MantisClone:
Unfortunately, both of the `FusePoolLens` functions that would return a user's underlying collateral balance revert, and seem to be broken: `getPoolUsersWithData` and `getPoolAssetsWithData`

So I'm attempting to reverse engineer them inside the Snapshot Strategy using lower-level calls to the `Comptroller`.

Reference: https://github.com/Rari-Capital/fuse-v1/blob/b5c269208f48ab413d816ccd5f67bf79caea73fd/src/FusePoolLens.sol#L252-L349

MantisClone:
Reading through the FusePoolLens.sol code, I think I can get a user's supplied balance directly from the cToken.
https://github.com/Rari-Capital/fuse-v1/blob/b5c269208f48ab413d816ccd5f67bf79caea73fd/src/FusePoolLens.sol#L319

MantisClone:
After further study of the cToken implementation on etherscan: https://etherscan.io/address/0x67db14e73c2dce786b5bbbfa4d010deab4bbfcf9#code, I realized that `cToken.balanceOfUnderlying(user)` doesn't exist on-chain. Instead, `cToken.balanceOf(user)` can be used.  

This means that I could plug the `cTokenAddress` into the standard `erc20-balance-of` strategy and it would work the same way! :facepalm:

Nonetheless, I decided to finish off the PR by changing the variable name from `address` to `cTokenAddress` and to update the README to make it rari-fuse specific.

MantisClone:
After spending 8 hours working on this, I realized that getting a user's balance in a given Rari Fuse market is as simple as calling `cToken.balanceOf(user)`, and that the same results could be achieved using existing strategies:

1. `erc20-balance-of` strategy by plugging the cToken address into the `address` parameter.
2. `cToken` strategy which adds an optional waiting period between minting and voting, and the option to exclude borrowers by granting them 0 voting weight.

I couldn't find precedents for duplicate strategies. 

I briefly considered creating a PR that contains only the rari-fuse README.md which then points the users to the `erc20-balance-of` and `ctoken` strategies. But I don't know if this would be compatible with the snapshot app.

I also wondered if `rari-fuse` could be an alias of `erc20-balance-of` but then I saw #107 which dismisses the idea of aliases.

Considering all this - I decided to simply close this PR and move on.

MantisClone:
I realized my last comment was wrong. In Rari Fuse, there's an exchange rate between the `fToken` (what I've previously been calling `cToken`) and the underlying token. So I'm thinking this strategy may still be needed.

ChaituVR (approved):

PR #717, opened 2022-07-18
Use brovider in test files

ChaituVR:


bonustrack (commented):
Can you use getProvider instead?

PR #748, opened 2022-07-29
feat: add a strategy to calc blance with subgraph [balance-of-subgraph]

weizard:
To calculate the token balance including user’s EOA and smart wallet, we developed this strategy. Developers can create their own subparagraphs.

- new balance of strategy with the subgraph data source

image

### DIFF
--- a/src/strategies/balance-of-subgraph/index.ts
+++ b/src/strategies/balance-of-subgraph/index.ts
@@ -1,0 +1,55 @@
+import { getAddress } from '@ethersproject/address';
+import { subgraphRequest } from '../../utils';
+
+const SUBGRAPH_URL = {
+  '1': 'https://api.thegraph.com/subgraphs/name/dinngodev/furucombo-tokenomics-mainnet',
+  '137':
+    'https://api.thegraph.com/subgraphs/name/dinngodev/furucombo-tokenomics-polygon'
+};
+
+export const author = 'weizard';
+export const version = '0.1.0';
+
+export async function strategy(
+  _space,
+  network,
+  _provider,
+  addresses,
+  options,
+  snapshot
+) {
+  const params = {
+    users: {
+      __args: {
+        where: {
+          id_in: addresses.map((address) => address.toLowerCase()),
+          amount_gt: 0
+        }
+      },
+      id: true,
+      amount: true
+    }
+  };
+  if (snapshot !== 'latest') {
+    // @ts-ignore
+    params.users.__args.block = { number: snapshot };
+  }
+  const result = await subgraphRequest(
+    options.subGraphURL ? options.subGraphURL : SUBGRAPH_URL[network],
+    params
+  );
+  const score = {};
+  if (result && result.users) {
+    result.users.forEach((user) => {
+      const userAddress = getAddress(user.id);
+      let userScore = Number(user.amount);
+
+      if (options.scoreMultiplier) {
+        userScore = userScore * options.scoreMultiplier;
+      }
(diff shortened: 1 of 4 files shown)


src/strategies/balance-of-subgraph/index.ts
```diff
@@ -0,0 +1,55 @@
...
+  const params = {
+    users: {
+      __args: {
+        where: {
+          id_in: addresses.map((address) => address.toLowerCase()),
+          amount_gt: 0
+        }
+      },
+      id: true,
+      amount: true
```
ChaituVR, line 30: Hi @weizard Looks good, but would be great if there is a way to pass optional param names for input and output from options. can check if there is a possibility.?

weizard: Did you mean that add token address and symbol into the option and filter the address in the subgraph ?

ChaituVR: I mean names of the params, `amount` and `id_in`

weizard: I am still not sure I actually understand, is it like that option has a property named `subGraphAddreeKey`, and its value may be `id`.
options will be
```
{
  "subGraphURL": "https://api.thegraph.com/subgraphs/name/dinngodev/furucombo-tokenomics-mainnet",
  "subGraphAddreeKey": "id"
}
```
And line 25 will be like
``` 
[`${options. subGraphAddreeKey ?? 'id'}_in`]: addresses.map((address) => address.toLowerCase()),
```

src/strategies/balance-of-subgraph/index.ts
```diff
@@ -0,0 +1,55 @@
+import { getAddress } from '@ethersproject/address';
+import { subgraphRequest } from '../../utils';
+
+const SUBGRAPH_URL = {
+  '1': 'https://api.thegraph.com/subgraphs/name/dinngodev/furucombo-tokenomics-mainnet',
+  '137':
+    'https://api.thegraph.com/subgraphs/name/dinngodev/furucombo-tokenomics-polygon'
+};
+
```
ChaituVR, line 9: Actually, the problem is with the name of the strategy @weizard It sounds like a core strategy that can be used by anyone 😄 but we hardcoded the subgraphs here, input and output fields so won't be useful for others. can find a way to pass these dynamically. or change the name to something like `furucombo-balance`

ChaituVR (changes requested):

src/strategies/balance-of-subgraph/index.ts
```diff
@@ -0,0 +1,55 @@
...
+      id: true,
+      amount: true
+    }
+  };
+  if (snapshot !== 'latest') {
+    // @ts-ignore
+    params.users.__args.block = { number: snapshot };
+  }
+  const result = await subgraphRequest(
+    options.subGraphURL ? options.subGraphURL : SUBGRAPH_URL[network],
```
weizard, line 38: Thanks for you explanation, I understand now.
We create for people who need to customize their own subgraph. With this strategy, they don’t need to create a custom strategy again.  And I add 3 new params for this propose.
At this line(38), we can see that user can set their own subgraph URL in params, if they don’t set then will use the default(hardcoded) subgraph. 
If still concern about this, we can remove the hardcoded URL or replace the name with `furucombo-smart-wallet-balance`.

ChaituVR: @weizard Makes sense, sorry didn't notice this line before, so should be fine :) will merge this

ChaituVR (approved):

PR #752, opened 2022-08-02
Multichain strategies limit

ChaituVR:


bonustrack:
@ChaituVR What's the reason for this change? This may break some pending or active proposals

ChaituVR:
Ah. yes, I thought there should be a way to stop accepting unlimited strategies passed into this. so thought this could be a good first step.

ChaituVR:
Closing this for now. until we find a better way

PR #753, opened 2022-08-03
new strategy: decaying voting power for staked ERC 1155s [echelon-cached-erc1155-decay]

brandonleung:
Adding new voting strategy for https://echelon.io/.

Explanation of logic:
- each cached ERC 1155 has a decaying amount of voting power
- at the starting block time, each asset has 4000 units of voting power, one year later it should be 0; this value decays linearly per second
- once we assign the corresponding votes for each address, we square the final value

### DIFF
--- a/src/strategies/echelon-cached-erc1155-decay/index.ts
+++ b/src/strategies/echelon-cached-erc1155-decay/index.ts
@@ -1,0 +1,70 @@
+import { Multicaller } from '../../utils';
+
+export const author = 'brandonleung';
+export const version = '1.0.0';
+
+const cachingAbi = [
+  {
+    inputs: [
+      {
+        internalType: "uint256",
+        name: "",
+        type: "uint256"
+      },
+      {
+        internalType: "address",
+        name: "",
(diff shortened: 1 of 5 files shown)


ChaituVR (commented):
Hi @brandonleung would be great if you can add a readme file to explain how this work, will be useful for others

src/strategies/echelon-cached-erc1155-decay/index.ts
```diff
@@ -0,0 +1,70 @@
...
+      {
+        internalType: "int256",
+        name: "rewardDebt",
+        type: "int256"
+      }
+    ],
+    stateMutability: "view",
+    type: "function"
+  }
+]
```
ChaituVR, line 36:
```suggestion
const cachingAbi = [
  'function cacheInfo(uint256, address) view returns (uint256 amount, int256 rewardDebt)'
];
```

ChaituVR (changes requested):

ChaituVR (approved):

PR #756, opened 2022-08-04
New strategy: voting power for NFT xSGT [xsgt-voting-power]

acinmavi:
- Add new strategy: voting power for NFT xSGT


src/strategies/xsgt-voting-power/index.ts
```diff
@@ -0,0 +1,34 @@
+import { BigNumberish } from '@ethersproject/bignumber';
+import { formatUnits } from '@ethersproject/units';
+import { Multicaller } from '../../utils';
+
+export const author = 'bluebelt';
```
ChaituVR, line 5:
```suggestion
export const author = 'acinmavi';
```

ChaituVR (changes requested):
@acinmavi `contract-call` strategy should work perfectly here, Have you tried it? let me know if it doesn't work :)

src/strategies/xsgt-voting-power/index.ts
acinmavi: Thanks @ChaituVR , I run strategy test and it worked perfectly .
I update again author. Please help to review it.

acinmavi:
Hi @ChaituVR , could you please help to review again ? I update author and test again contract with 500 address . It's ok now.
And one question : If my pull request is accepted , does my strategy appear on https://snapshot.org and https://demo.snapshot.org ? I need this to test voting.
Best regards.

ChaituVR:
@acinmavi Like i mentioned `contract-call` strategy should work fine in this case, no need of a new strategy, Can try that strategy?

acinmavi:
Thanks @ChaituVR I could use contract-call for that strategy
Regards.

PR #757, opened 2022-08-04
feat: add harmony-staking strategy [harmony-staking]

MaxMustermann2:
calculates votes of validators through the total balance delegated to
them (including self delegation)

- add harmony staked balance strategy for validator addresses

### DIFF
--- a/src/strategies/harmony-staking/index.ts
+++ b/src/strategies/harmony-staking/index.ts
@@ -1,0 +1,44 @@
+export const author = 'harmony-one';
+export const version = '0.0.1';
+
+import { StaticJsonRpcProvider } from '@ethersproject/providers';
+import { formatUnits } from '@ethersproject/units';
+import { BigNumber } from '@ethersproject/bignumber';
+
+type Params = {
+    symbol: string;
+    decimals: number;
+};
+
+export async function strategy(
+    _space: string,
+    _network: string,
+    provider: StaticJsonRpcProvider,
+    // adding a 0 value for addresses not in the result is not needed
+    // since they are dropped anyway in utils.ts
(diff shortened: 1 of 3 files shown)


src/strategies/harmony-staking/index.ts
```diff
@@ -0,0 +1,44 @@
+export const author = 'harmony-one';
+export const version = '0.0.1';
+
+import { StaticJsonRpcProvider } from '@ethersproject/providers';
+import { formatUnits } from '@ethersproject/units';
+import { BigNumber } from '@ethersproject/bignumber';
```
ChaituVR, line 6:
```suggestion
import { StaticJsonRpcProvider } from '@ethersproject/providers';
import { formatUnits } from '@ethersproject/units';
import { BigNumber } from '@ethersproject/bignumber';

export const author = 'harmony-one';
export const version = '0.0.1';

```

src/strategies/harmony-staking/index.ts
```diff
@@ -0,0 +1,44 @@
...
+        [blockTag],
+    );
+    return Object.fromEntries(
+        Object.entries(response).map(([address, balance]) => [
+          address,
+          parseFloat(
+            formatUnits(
+                BigNumber.from('0x' + balance.toString(16)),
+                options && options.decimals ? options.decimals: 18,
+            )
```
ChaituVR, line 40: Hey @MaxMustermann2 I think `contract-call` strategy should work in this case, let me know if it doesn't work :) Thanks

ChaituVR (changes requested):

src/strategies/harmony-staking/index.ts
MaxMustermann2: Hey @ChaituVR, this is an RPC call and not a contract call that uses the EVM. I don't think `contract-call` will work (there is no ABI or contract address).

ChaituVR (approved):
Thanks for explaining @MaxMustermann2 :)

PR #770, opened 2022-08-13
[orca-pod] Orca pods strategy

### LINKED ISSUE #725: Create strategy for Orca
This is a contract ERC1155 that include all the Orca pods https://etherscan.io/token/0x0762aa185b6ed2dca77945ebe92de705e0c37ae3#inventory each pod have their own tokenId. Each token can have multiple holders. We should create a strategy "orca-pod" to give 1 voting power for each holder of a specific tokenId holder. Ideally the weight should be a param, to allow changing vp from 1 to any amount

ChaituVR:
Fixes #725


bonustrack:
- It should be called "orca-pod".
- We should have a schema for it.
- And we should have a readme

src/strategies/orca-pods/index.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export async function strategy(
+  space,
+  network,
+  provider,
+  addresses,
+  options,
+  snapshot
+) {
+  const defaultWeight = 1;
```
bonustrack, line 14: You dont need to make it a variable if its only used once

ChaituVR:
To add schema we need this https://github.com/snapshot-labs/snapshot/issues/2628

bonustrack (approved):

PR #774, opened 2022-08-15
Proxy Protocol ERC1155 [proxyprotocol-erc1155-balance-of]

rawrjustin:

New strategy for ERC 1155 for using proxy wallet mappings to actual wallet mappings to calculate voting power.

**Example:**
wallet 0xABC is a proxy wallet that maps to [0x123, 0x456, 0x789]
I calculate the voting power from [0x123, 0x456, 0x789] as [0, 1, 4]
I sum the voting power from all the wallets as 5
wallet 0xABC has voting power of 5

> the service is a community management tool that allows communities to have social functionality on top of snapshot and other governance features like idea boards/ forums. https://app.proxychat.xyz/post/PN7bFYYikNqD58s74x5F/
> 
> we want to allow users to vote with a proxy wallets so that you don't need to sign using your cold wallet everytime you want to vote with your NFT.
> 
> the mapping returned here is a mapping of a temporary proxy wallets -> actual wallets that hold tokens/nfts as detailed in the example. Let me know if that makes sense!

### DIFF
--- a/src/strategies/proxyprotocol-erc1155-balance-of/index.ts
+++ b/src/strategies/proxyprotocol-erc1155-balance-of/index.ts
@@ -3,33 +3,10 @@ import { formatUnits } from '@ethersproject/units';
 import { multicall } from '../../utils';
 
 export const author = 'rawrjustin';
-export const version = '0.1';
+export const version = '0.1.0';
 
 const abi = [
-  {
-    inputs: [
-      {
-        internalType: 'address',
-        name: 'owner',
-        type: 'address'
-      },
-      {
-        internalType: 'uint256',
-        name: 'id',
-        type: 'uint256'
-      }
-    ],
-    name: 'balanceOf',
-    outputs: [
-      {
-        internalType: 'uint256',
-        name: '',
-        type: 'uint256'
-      }
-    ],
-    stateMutability: 'view',
-    type: 'function'
-  }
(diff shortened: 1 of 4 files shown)


src/strategies/proxyprotocol-erc1155-balance-of/index.ts
```diff
@@ -0,0 +1,102 @@
+import fetch from 'cross-fetch';
+import { formatUnits } from '@ethersproject/units';
+import { multicall } from '../../utils';
+
+export const author = 'rawrjustin';
+export const version = '0.1';
```
ChaituVR, line 6:
```suggestion
export const version = '0.1.0';
```

ChaituVR (changes requested):

src/strategies/proxyprotocol-erc1155-balance-of/index.ts
```diff
@@ -0,0 +1,102 @@
...
+      [address, options.tokenId]
+    ]),
+    { blockTag }
+  );
+  const addressScores = Object.fromEntries(
+    response.map((value, i) => [
+      flattenedWalletAddresses[i],
+      parseFloat(formatUnits(value.toString(), options.decimals))
+    ])
+  );
```
ChaituVR, line 94: Maybe we can directly call erc1155-balance-of strategy function here? Reduces duplicate code

PR #782, opened 2022-08-18
Add fight-club strategy [fight-club]

MantisClone:
- Add a `fight-club` strategy that calculates the voter's score based on glove NFTs and weight class Kudos.

I've chosen to omit a `schema.json` with this PR because the `"type": "object"` parameters don't render in snapshot playground, so I'd rather have the simple parameter entry box.


MantisClone:
Huh.. I don't understand why the `test-strategy` workflow keeps getting skipped.  I'll just run the tests locally I guess.

ChaituVR:
Hello @MantisClone looks like it is same as erc20-balance-of? you want to add `calculates the voter's score based on glove color and weight class NFTs.`?

src/strategies/arrow-vesting/index.ts
```diff
@@ -7,79 +7,106 @@ export const author = 'BrassLion';
 export const version = '0.1.0';
 
 const vestingFactoryAbi = [
-  "function escrows_length() public view returns (uint256)",
-  "function escrows(uint256 index) public view returns (address)",
+  'function escrows_length() public view returns (uint256)',
+  'function escrows(uint256 index) public view returns (address)'
```
ChaituVR, line 11: @MantisClone Make sure there are no changes in other strategies when we go here https://github.com/snapshot-labs/snapshot-strategies/pull/782/files#

ChaituVR (changes requested):

src/strategies/arrow-vesting/index.ts
MantisClone: Done.

src/strategies/fight-club/index.ts
```diff
@@ -0,0 +1,104 @@
...
+      multi.call(
+        `${address}.gloves.${gloveAddress}`,
+        gloveAddress,
+        'balanceOf(address)',
+        [address]
+      );
+    });
+  });
+
+  Object.entries(options.weightClassIds).forEach(([weightClassId]) => {
```
ChaituVR, line 35: Hi @MantisClone Sorry for not noticing this before, but I feel we should add a limit to `gloveAddresses` and `weightClassIds` if someone passes unlimited entries, there could be memory issues. wyt?

ChaituVR (changes requested):

src/strategies/fight-club/index.ts
MantisClone: I'd be okay with adding a limit. I've decided to arbitrarily limit it to 10 glove colors and 10 weight classes. 

If I had unlimited time, I'd run a test to determine the actual limits. But in practice, I know there will only be 4 glove colors and 4 or 5 weight classes. So setting an arbitrary limit is the more pragmatic approach at this time.

ChaituVR (approved):

PR #791, opened 2022-08-22
[improvement] ensure address capitalization tested explicitly, test all strategies script

DanielZlotin:
Suggested improvement to the base strategy test that validates addresses capitalization: making sure to input both checksumed and lowercased addresses so that both cases are tested explicitly, incase example addresses did not.

- ensure example addresses tested with lowercase and checksum capitalization
- add a script to run all strategy tests under `/src/strategies` (**currently a lot of strategies are failing**)

### DIFF
--- a/test/strategy.test.ts
+++ b/test/strategy.test.ts
@@ -95,6 +95,8 @@ describe(`\nTest strategy "${strategy}"`, () => {
   });
 
   it('Returned addresses should be either same case as input addresses or checksum addresses', () => {
+    example.addresses[0] = example.addresses[0].toLowerCase();
+    example.addresses[1] = getAddress(example.addresses[1]);
     expect(
       Object.keys(scores[0]).every(
         (address) =>
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
   "scripts": {
     "build": "tsc -p .",
     "test": "jest -i strategy.test.ts",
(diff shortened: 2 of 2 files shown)


test/strategy.test.ts
```diff
@@ -26,6 +26,9 @@ if (!strategy) throw 'Strategy not found';
 const example = require(`../src/strategies/${strategy}/examples.json`)[0];
 
 function callGetScores(example) {
+  example.addresses[0] = example.addresses[0].toLowerCase();
```
ChaituVR, line 29: Why lowerCase? 🤔

DanielZlotin: In order to make sure strategies correctly handle address capitalization. See issue #790
The is already a test that checks to make sure a strategy returns same address capitalization or checksum, but it relies on the input address in each strategy's example data.
In our case, the example addresses were all lowercased, the strategy itself returned `toLowerCase()` addresses so the test passed. But in production metamask returned checksum addresses so there was a mismatch and all voting returned 0.
This makes sure input capitalization is "randomized" so that expected behavior is actually tested. This would have caught the issue.

ChaituVR: > But in production metamask returned checksum addresses so there was a mismatch and all voting returned 0.

Where is production you mean? 🤔 in score API? or snapshot.org, can share me the proposal where this happened?

PR #813, opened 2022-09-02
Ethermon-ERC721 Changes for calculating vote power [ethermon-erc721]

syedMohib44:
Fixes # .

- Fixed variable swapped EMONA_ADDRESS with EMON_DATA_ADDRESS.
- Somehow .toFixed(0) instead of rounding of number concatinating string causing infite vote power casted that to number.


test/vp.tst.ts
```diff
@@ -0,0 +1,59 @@
+import { getVp } from '../src/utils/vp';
```
ChaituVR, line 1: Why this file is renamed? 🤔

ChaituVR (changes requested):
Make sure to revert changes from package-lock.json file. https://github.com/snapshot-labs/snapshot-strategies/pull/813/files should contain only changes for your strategy :)

test/vp.tst.ts
syedMohib44: Not sure man just pulled from master

syedMohib44 (commented):
@ChaituVR remove invalid vp.tst.ts file and package.json

ChaituVR (approved):

PR #820, opened 2022-09-06
Update MinotaurMoney voting strategy [minotaur-money]

pepperstepper:
Fixes # .

- 
- 
-

### DIFF
--- a/src/strategies/minotaur-money/examples.json
+++ b/src/strategies/minotaur-money/examples.json
@@ -12,6 +12,6 @@
       "0x4E5D385E44DCD0b7adf5fBe03A6BB867A8A90E7B",
       "0x86Fdd9980aCD3e2C8e7959Db344Ff6D5FD5743F5"
     ],
-    "snapshot": 2917145
+    "snapshot": 4521240
   }
 ]
--- a/src/strategies/minotaur-money/index.ts
+++ b/src/strategies/minotaur-money/index.ts
@@ -111,8 +111,8 @@ export async function strategy(
       .mul(index) // timeses by 10^9 effectively
 
     const minoScore = wsMinoScore
-      .add(BigNumber.from(minoBalances[address+"0"] || 0).mul(BigNumber.from(10).pow(18)))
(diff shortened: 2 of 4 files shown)


pepperstepper:
Is there an easy way to find out the current snapshot index for a chain i.e. cronos? Wanted to update the example.json too

ChaituVR:
Hey, @pepperstepper what is the `current snapshot index` 🤔 you mean to get the latest block of Cronos?

pepperstepper:
> Hey, @pepperstepper what is the `current snapshot index` 🤔 you mean to get the latest block of Cronos?

Hi, I was under impression the snapshots where taken on specific blocknumbers and you couldnt just put any one in there.

Maybe they changed it? Or maybe I was mistaken last time I did this. I thought trying random genuine ones didn't last time, but it seems to have worked this time!

Updated the PR with that, we would appreciate it if this could be merged,
Thank you very much!

ChaituVR (changes requested):
Can update version to `export const version = '0.0.4';` plz :)

ChaituVR:
@pepperstepper also looks like test cases are failing because it is taking more than 10 secs to resolve. can you check on that?

pepperstepper:
Hi,
It seems to be intermittent, for me it fails 1/5 runs or so. due to being over 10 seconds. Perhaps the cronos rpc is a bit on the slow side occasionally?

pepperstepper:
There are 5 rpc calls per address (last I checked this was within the standard for PRs) and the rest of the logic should not be using much tine.

pepperstepper:
> There are 5 rpc calls per address (last I checked this was within the standard for PRs) and the rest of the logic should not be using much tine.

Actually should be less than this as makeMulticaller is used

pepperstepper:
I just refactored and it passed u under 10 seconds 10 times in a row on my machine (but I could have just got a good run of rpc latency) hopefully that suffices.

pepperstepper:
I was trying to bundle the mino balanceOf calls with the sMino balanceOf calls by re uising the Multicaller object, but it seemed to be overwriting the first set of calls anyway.

ChaituVR (approved):

PR #822, opened 2022-09-07
Add strategy for nation3 [nation3-votes-with-delegations]

Caranell:
- Add a strategy for nation3 voting with possible delegations


ChaituVR:
@Caranell Test cases fail when we pass more addresses, can look into it? :) you can test by running: `yarn test --strategy=nation3-votes-with-delegations --more=500`

ChaituVR:
also, Readme would help instead of empty file 😄

Caranell:
@ChaituVR Thanks for the tips! Completely forgot about readme and was only running tests on my list of addresses from `example` -- seems to be ok now 🍻

src/strategies/nation3-votes-with-delegations/examples.json
```diff
@@ -0,0 +1,192 @@
...
+      "0xF84877138A1924C298632250745C91E37532d8C7",
+      "0x70f7a2Ca4930C19669cd71d12Da9922E230905Aa",
+      "0xdA2a387fAA9F6a679875072e1A82A0059A6D34c0",
+      "0xb33eDF0aba01F459B569A85aE6311ec683A20d7f",
+      "0xf3FDa45764d2C162FBdE6d9c9141877FD76e59E7",
+      "0x47d80912400ef8f8224531EBEB1ce8f2ACf4b75a",
+      "0xEdd000B7Db3cb8931d4E0cb1D0DBe6B947Ceb09A",
+      "0x57559246958078503FBc24D8E8D5b419d6CAb24C",
+      "0x2Ee8670d2b936985D5fb1EE968810c155D3bB9cA",
+      "0xc2585a806E3c8F6ae3B79D363348aE8cAd8A3B45"
```
ChaituVR, line 188:
```suggestion
      "0x6412517845128C289e111Ad9dA26100f78e529d8"
```

src/strategies/nation3-votes-with-delegations/index.ts
```diff
@@ -0,0 +1,124 @@
...
+    erc20BalanceCaller.execute(),
+    erc721LastTokenIdCaller.execute()
+  ]);
+
+  const lastTokenId = BigNumber.from(lastIndex.lastTokenId).toNumber();
+
+  for (let i = 0; i < lastTokenId; i++) {
+    erc721SignerCaller.call(i, options.erc721, 'signerOf', [i]);
+    erc721OwnerCaller.call(i, options.erc721, 'ownerOf', [i]);
+  }
```
ChaituVR, line 67: Hey, @Caranell Everything looks good to me :) except that I feel someone might pass their custom contract for this and lastTokenId could be a huge number and it could eat up all memory. of course, it is a rare case but would be great if we can have a limit if it crosses that limit we can `throw` an error. LMK what you think

ChaituVR: Hmm, yea no idea too 😄 I am starting to worry that this may not scale well if there are more tokens in your passport contract, there is no other way to find delegates than to get all owners and signers for each token.? maybe a subgraph or something?

Caranell: Hi @ChaituVR. I discussed the problem with the nation3 team and we don't expect the number of 'passport tokens' to exceed even a thousand in the near future. So I suggest leaving the current behavior as it is, and if the number of tokens significantly increases, we'll use subgraphs. What do you think?

ChaituVR (approved):

PR #823, opened 2022-09-07
Add strategy LRC NFT balances by minter account [lrc-l2-nft-balance-of]

Karamorf:
- add strategy for NFT's minted on Loopring L2

### DIFF
--- a/src/strategies/lrc-l2-nft-balance-of/index.ts
+++ b/src/strategies/lrc-l2-nft-balance-of/index.ts
@@ -1,0 +1,67 @@
+import { subgraphRequest } from '../../utils';
+
+export const author = 'karamorf';
+export const version = '0.1.1';
+
+const LIMIT = 500;
+
+function makeQuery(minter, skip, blacklisted_accounts) {
+  const query: any = {
+    nonFungibleTokens: {
+      __args: {
+        where: {
+          minter: minter,
+          slots_: {
+            account_not_in: blacklisted_accounts
+          }
+        },
+        first: LIMIT,
+        skip: skip
+      },
+      token: true,
+      slots: {
+        account: {
+          address: true
+        },
+        balance: true
+      }
+    }
+  }
+  return query;
+}
+
+export async function strategy(
+  space,
+  network,
+  provider,
+  addresses,
+  options,
+  snapshot
+): Promise<Record<string, number>> {
+  console.log("hello");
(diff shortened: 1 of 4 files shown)


src/strategies/lrc-l2-nft-balance-of/index.ts
```diff
@@ -0,0 +1,80 @@
...
+      balance: true,
+      lastUpdatedAt: true
+    }
+  };
+
+  if (snapshot !== 'latest') {
+    query.accountNFTSlots.__args.where = {
+      ...query.accountNFTSlots.__args.where,
+      _change_block: {
+        number_gte: snapshot
```
ChaituVR, line 29: why `gte`?? is there a reason? ideally, voting power should be based on the snapshot. else there can be inconsistent voting power during the proposal voting period

ChaituVR (changes requested):

src/strategies/lrc-l2-nft-balance-of/index.ts
Karamorf: oh, is that what this parameter is doing? we want values before block X for when the proposal was submitted?

Had the roughest of times trying to figure out what this parameter meant. It should probably be `lt` right?

Karamorf: @ChaituVR if I do something like this:
```suggestion
        lastUpdatedAtTransaction_lt: snapshot 
```

then it looks like it'd exclude accounts that received tokens after the snapshot. However you wouldn't get the balance at the time of the snapshot, just the current balance. I guess that might be good enough. Is this true of the `lrc-l2-subgraph-balance-of` strategy you wrote? It looks like given my observation here that it'd also apply to in that strategy.

To see this, using this graph query:
```
{
 accountNFTSlots(first: 5, where: { nft_: { minter: "57727"}, lastUpdatedAtTransaction_lt: "28307" }) {
   account { address } 
   balance
   lastUpdatedAtTransaction {
     id
   }
 }
}
```
I'd expect to see a balance of 100k because I didn't move those other tokens out of this account until block `28578`.

Karamorf: ahh, I see what's happening in your strategy now. It's in the root __args filter, not within the where filter.

ChaituVR (approved):

PR #824, opened 2022-09-08
Add BMC strategy [bmc]

JustinRWong:
Creates a new strategy for Blockchain Miners Club

- Adding BMC strategy to read from all Blockchain Miners Clubs' smart contracts

BMC Project Link: https://www.blockchainminers.club/

### DIFF
--- a/src/strategies/bmc/index.ts
+++ b/src/strategies/bmc/index.ts
@@ -1,0 +1,75 @@
+import { BigNumber } from '@ethersproject/bignumber';
+import { Multicaller } from '../../utils';
+
+export const author = 'justinrwong';
+export const version = '0.0.1';
+
+const ogAbi = [
+  'function walletMiners(address account) external view returns (uint256[])'
+];
+
+const ultraAbi = [
+  'function tokensOfOwner(address account) external view returns (uint256[])'
+];
+
+const stakingAbi = [
+  'function getStakedOGs(address account) external view returns (uint256[])',
+  'function getStakedULTRAs(address account) external view returns (uint256[])'
+];
+
+export async function strategy(
+  space,
+  network,
+  provider,
+  addresses,
+  options,
+  snapshot
+): Promise<Record<string, number>> {
+  const blockTag = typeof snapshot === 'number' ? snapshot : 'latest';
+
+  // Setup multicall for the OG/Ultra/Staking Contracts
+  const ogToken = new Multicaller(network, provider, ogAbi, { blockTag });
+  const ultraToken = new Multicaller(network, provider, ultraAbi, { blockTag });
+  const stakingOGToken = new Multicaller(network, provider, stakingAbi, {
+    blockTag
+  });
+  const stakingUltraToken = new Multicaller(network, provider, stakingAbi, {
+    blockTag
+  });
+  addresses.forEach((address) => {
+    ogToken.call(address, options.og, 'walletMiners', [address]);
+    ultraToken.call(address, options.ultra, 'tokensOfOwner', [address]);
+    stakingOGToken.call(address, options.staking, 'getStakedOGs', [address]);
+    stakingUltraToken.call(address, options.staking, 'getStakedULTRAs', [
+      address
+    ]);
+  });
+
+  // when no assets are held, the contract throws an error, which means it's empty.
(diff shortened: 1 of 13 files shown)


JustinRWong:
All tests pass locally, though there might be some variance in how long the strategy takes since it's doing 4 different calls to different contracts
Screenshot 2022-09-08 at 7 41 23 PM

src/strategies/bmc/index.ts
```diff
@@ -0,0 +1,75 @@
...
+  const [ogResponse, ultraResponse, stakedOgResponse, stakedUltraResponse]: [
+    Record<string, BigNumber[]>,
+    Record<string, BigNumber[]>,
+    Record<string, BigNumber[]>,
+    Record<string, BigNumber[]>
+  ] = await Promise.all([
+    ogToken.execute().catch((error) => []),
+    ultraToken.execute().catch((error) => []),
+    stakingOGToken.execute().catch((error) => []),
+    stakingUltraToken.execute().catch((error) => [])
```
ChaituVR, line 58: Hey @JustinRWong Don't catch the errors, in case of an error, users may get `0` voting power, which is not a correct voting power (we handle these errors in a top level, so should be fine :)

ChaituVR (changes requested):

src/strategies/index.ts
```diff
@@ -735,7 +736,8 @@ const strategies = {
   'safe-vested': safeVested,
   'riskharbor-underwriter': riskharborUnderwriter,
   'otterspace-badges': otterspaceBadges,
-  'synthetic-nouns-with-claimer': syntheticNounsClaimerOwner
+  'synthetic-nouns-with-claimer': syntheticNounsClaimerOwner,
+  bmc: bmc
```
ChaituVR, line 740:
```suggestion
  bmc
```

src/strategies/bmc/index.ts
JustinRWong: What do you mean by don't catch the error? the contract itself throws the error in the call which makes the tests fail. Do you mean contracts calls that return errors are already handled by Multicall?

ChaituVR: Answered in discord, make sure it doesn't return any errors, remove catch part here

JustinRWong: Following up, since the contract themselves have a `require` statement, a [CALL_EXCEPTION](https://docs.ethers.io/v5/troubleshooting/errors/#help-CALL_EXCEPTION) error can happen, making the catch statement necessary for the tests to pass.

to resolve, this now catches only the CALL_EXCEPTION error code and would otherwise throw the error

src/strategies/bmc/index.ts
```diff
@@ -0,0 +1,87 @@
...
+  // when no assets are held, the contract throws an error, which means it's empty.
+  const [ogResponse, ultraResponse, stakedOgResponse, stakedUltraResponse]: [
+    Record<string, BigNumber[]>,
+    Record<string, BigNumber[]>,
+    Record<string, BigNumber[]>,
+    Record<string, BigNumber[]>
+  ] = await Promise.all([
+    ogToken
+      .execute()
+      .catch((error) => (error.code === 'CALL_EXCEPTION' ? [] : _throw(error))),
```
ChaituVR, line 61: Hello @JustinRWong in cases where the query hit gas cap of the node we are using, then also we may hit `CALL_EXCEPTION` better to check the message `Address owns no tokens` instead, we can't access this message.?

JustinRWong: No, we can't access the message because the `CALL_EXCEPTION` ends up coming from the `require` statement of the smart contract

ChaituVR: Example of how this could ignore other errors as well:
`staking` contract doesn't exist at the `snapshot` block given in examples.json and throws an error, and because of the catch  error is hidden and voting power will be `0` 

There can be a lot of exception errors. and voting power will be inconsistent

ChaituVR: I don't see a solution for this. maybe would be helpful if you have a subgraph for this contract https://etherscan.io/address/0x0c6822ca73de6871f27acd9ca05a05b99294b805#readContract

ChaituVR:
Closing as stale, please reopen if needed

PR #826, opened 2022-09-09
Add zkportal strategy [zkportal]

vicsn:
Fixes # .

Changes proposed in this pull request: Adding the zkPortal strategy
- Adding the strategy to verify unique identity based on web2 KYC credentials and government accounts
- Link to the protocol: https://zkportal.io/docs/

### DIFF
--- a/src/strategies/zkportal/index.ts
+++ b/src/strategies/zkportal/index.ts
@@ -1,0 +1,39 @@
+import { BigNumberish } from '@ethersproject/bignumber';
+import { formatUnits } from '@ethersproject/units';
+import fetch from 'cross-fetch';
+
+export const author = 'vicsn';
+export const version = '0.1.0';
+
+export async function strategy(
+  space,
+  network,
+  provider,
+  addresses,
+  options,
+  snapshot
+): Promise<Record<string, number>> {
+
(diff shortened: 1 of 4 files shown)


src/strategies/zkportal/index.ts
```diff
@@ -0,0 +1,39 @@
...
+export async function strategy(
+  space,
+  network,
+  provider,
+  addresses,
+  options,
+  snapshot
+): Promise<Record<string, number>> {
+
+  const url = `https://snapshot.zkportal.io/verify_eth_addresses`;
```
ChaituVR, line 17: Hey @vicsn you own this API? if so, I think it is better to use https://github.com/snapshot-labs/snapshot-strategies/tree/master/src/strategies/api-post strategy in your space directly. to avoid duplication. but is there some reason for a separate strategy or you don't have access to change this API? let me know :)

ChaituVR (changes requested):

src/strategies/zkportal/index.ts
vicsn: Hi Chaitu, thx for the quick response. Makes sense, I didn't know the other strategy existed! We will actually build in signature verification into this strategy as well, to ensure DAO users require less trust in our backend infra. Will close this PR and open it when signature verification is added.

PR #829, opened 2022-09-11
Improvement and more options to LRC L2 NFT balance strategy [lrc-l2-nft-balance-of]

curtis741:
adding options for blacklisting nft id's and filter by token id(s) instead of minter account id

- Improve filter of NFT's by add tokens as a parameter to be an alternative to minter account id
- Improve blacklisting by add blacklisted_nft_id's so mismints or any other nft can be excluded from the result set

### DIFF
--- a/src/strategies/lrc-l2-nft-balance-of/README.md
+++ b/src/strategies/lrc-l2-nft-balance-of/README.md
@@ -8,10 +8,14 @@ Here is an example of parameters:
 {
   "graph": "https://api.thegraph.com/subgraphs/name/juanmardefago/loopring36",
   "minter_account_id": "74447",
-  "blacklisted_account_ids": ["38482"]
+  "tokens": ["token (Collection) id's to include"],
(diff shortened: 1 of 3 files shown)


ChaituVR:
@Karamorf  Can confirm this change? Also make sure there are no open proposals in your space

ChaituVR:
@curtis741 Looks like there are conflicts. would be great if you can resolve

curtis741:
@ChaituVR huh, I swore I pulled from master before starting this branch but I guess not. conflicts resolved.

Yes, this changes are to help support a couple requests from others that wanted to use this strategy. It should be entirely backward compatible. Also haven't even made my space yet so no worries on open proposals :)

ChaituVR (approved):
Thanks for resolving @curtis741 yea there are new strategies that are merged in between so that cause it :) so I see this PR is coming from @Karamorf Repo. so should be fine. merging it 🚀

PR #831, opened 2022-09-13
Update iotex API, add iotex test address to default list[iotex-staked-balance]

millken:
Fixes # .

- Update iotex API endpoints and func name.


millken:
@ChaituVR

ChaituVR:
@maswalker Can you confirm these changes?

guo:
Hello, @ChaituVR !

This is Guo from IoTeX. We have run the test locally and it worked. We suspect that the endpoint might be slow during the last run. Could you please re-run the test here?  

Thank you!

bonustrack:
@millken Tests are failing, can you have a look on the error and try to fix it? Check here https://github.com/snapshot-labs/snapshot-strategies/actions/runs/3047847939/jobs/4913284908

maswalker (approved):

guo:
@bonustrack @ChaituVR  could you please rerun the test? We have improved the endpoint. It is passing when we test locally.

guo:
The tests are passed now. @ChaituVR, could you please review and merge the PR? Thank you!

ChaituVR:
Hey @guo `iotex-balance` strategy works fine now. but not `iotex-staked-balance` 

```js
Test strategy "iotex-staked-balance" with example index 0 (latest snapshot)
    ✓ Strategy should run without any errors (946 ms)
    ✕ Should return an array of objects with addresses (3 ms)
```

I understand it is urgent. but it should work with the latest snapshot too. maybe the addresses passed into examples.json doesn't have VP for the latest block? can fix it quickly?

src/strategies/iotex-balance/index.ts
```diff
@@ -5,10 +5,10 @@ interface ApiReturn {
 }
 
 export const author = 'iotex';
```
ChaituVR, line 7:
```suggestion
export const author = 'iotexproject';
```

src/strategies/iotex-staked-balance/index.ts
```diff
@@ -4,10 +4,10 @@ interface ApiReturn {
 }
 
 export const author = 'iotex';
```
ChaituVR, line 6:
```suggestion
export const author = 'iotexproject';
```

millken:
@ChaituVR It has already been fixed, all test should be passed.

ChaituVR (approved):

guo:
@ChaituVR , thanks for the quick review!

PR #852, opened 2022-09-27
Cross Chain cyberkongz [cyberkongz-v3]

maxbrand99:
- This snapshot strategy changes cyberkongz-v2 strategy and makes it so that it is cross chain compatible by using code from the multichain strategy


ChaituVR (changes requested):
Hey @maxbrand99 Is it possible to make this work for one network ? and pass multiple strategies into your space with different networks

maxbrand99:
> Hey @maxbrand99 Is it possible to make this work for one network ? and pass multiple strategies into your space with different networks

I think it probably is, the goal of this was to avoid having to deploy multiple strategies since cyberkongz went cross chain. With the way that we count votes, the total amount of banana is really important, and that can only be done with info from both chains at the same time.

maxbrand99:
> Hey @maxbrand99 Is it possible to make this work for one network ? and pass multiple strategies into your space with different networks

Is there any particular reason this needs to be done? I know the example is currently under the limit of 5 calls / strategy (it makes 1 call / chain + the call for total banana supply)

src/strategies/cyberkongz-v3/index.ts
```diff
@@ -0,0 +1,71 @@
...
+  const allCalls: any[] = [];
+  options.chains.forEach((chain) => {
+    const calls: any[] = [];
+    Object.keys(chain.registries).forEach((registry) => {
+      allAddresses[registry] = chain.registries[registry]
+      addresses.forEach((address: any) => {
+        calls.push([registry, 'balanceOf', [address]]);
+        allCalls.push([registry, 'balanceOf', [address]]);
+      });
+    });
```
ChaituVR, line 27: May concern is this, someone may pass multiple chains and registries here, and it may eat up some memory.

ChaituVR (changes requested):

src/strategies/cyberkongz-v3/index.ts
```diff
@@ -0,0 +1,71 @@
...
+    const blockTag = typeof blocks[chain.network] === 'number' ? blocks[chain.network] : 'latest';
+    promises.push(
+      multicall(chain.network, getProvider(chain.network), abi, calls, {blockTag})
+    );
+  })
+
+  const results = await Promise.all(promises);
```
ChaituVR, line 34: And if so it will cross our 5 call limit

ChaituVR (changes requested):

src/strategies/cyberkongz-v3/index.ts
maxbrand99: wouldn't the exact same thing happen here?
https://github.com/snapshot-labs/snapshot-strategies/tree/master/src/strategies/multichain
I could request as many chains as i want, easily passing the limit of 5 calls

maxbrand99:
It has now been restricted so that it will only look at main chain and matic chain

src/strategies/cyberkongz-v3/index.ts
ChaituVR: Yes yes multichain is an exception, we wanted to apply a limit for it too, but don't want to break existing settings (It doesn't have any limit because this strategy was created long back before our core multichain feature and many spaces still depend on it 😅 )

ChaituVR (approved):

PR #859, opened 2022-10-03
[aavegotchi-agip-37-unstaked-lp] aavegotchi-agip-37-unstaked-lp strategy

programmablewealth:
- New Aavegotchi snapshot strategy: aavegotchi-agip-37-unstaked-lp


ChaituVR:
@programmablewealth I see that you are creating many strategies but no space is using them? are you sure that these strategies will be used by someone/you?

programmablewealth:
> @programmablewealth I see that you are creating many strategies but no space is using them? are you sure that these strategies will be used by someone/you?

@ChaituVR we have 3 strategies currently live in the Aavegotchi snapshot and we planned to add more strategies covered in the previous 3 pull requests I raised.

We reliased this would push us over the 8 strategy limit per instance, so this new strategy will be used in place one of the strategies covered under the previous pull requests erc20-tokens-per-uni.

The strategies needed to be broken down to ensure we can support 500 addresses in under 20 seconds.

programmablewealth:
The Aavegotchi DAO has requested these strategies to be implemented via the bounty program following the Aavgotchi Snapshot proposal AGIP37 passing.

ChaituVR:
Thanks for explaining 👍

src/strategies/aavegotchi-agip-37-unstaked-lp/index.ts
```diff
@@ -0,0 +1,92 @@
...
+      uint256 unstakedWapGHSTVP,
+      uint256 stakedWapGHSTVP,
+      uint256 stakedGHSTLPTokensVP,
+      uint256 unstakedGHSTLPTokensVP
+    )
+  )`,
+  `function agip37VotingPowerNumber(address _account) public view returns (uint256 votingPower)`,
+  'function balanceOf(address account) view returns (uint256)',
+  'function gltrAllUnstakedLPTokenVotingPower(address _account) public view returns (uint256)',
+  'function batchAgip37VotingPower(address[] memory _accounts) public view returns (uint256[] memory)'
```
ChaituVR, line 19:
```suggestion
  'function gltrAllUnstakedLPTokenVotingPower(address _account) public view returns (uint256)'
```

src/strategies/aavegotchi-agip-37-unstaked-lp/index.ts
```diff
@@ -0,0 +1,80 @@
...
+    // loggedString += "\nSTAKED VOTING POWER\n" + JSON.stringify(votingPower.staked);
+    // // loggedString += "\nUNSTAKED TOKENS\n" + JSON.stringify(tokens.unstaked);
+    // loggedString += "\nUNSTAKED VOTING POWER\n" + JSON.stringify(votingPower.unstaked);
+    // loggedString += "\nTOTAL VOTING POWER\n" + totalVotingPower;
+    // console.log(loggedString);
+
+    entries[address] = totalVotingPower;
+  }
+
+  console.log(entries);
```
ChaituVR, line 77:
```suggestion
```

ChaituVR (changes requested):
I feel like we can use `contract-call` strategy here, you tried it? it takes more time when you pass 500 addresses with this call.?

programmablewealth:
> I feel like we can use `contract-call` strategy here, you tried it? it takes more time when you pass 500 addresses with this call.?

I haven't tried this yet, I will give it a go and see if it can handle 500 addresses.

programmablewealth:
contract-call is the way to go mate. It does the job just fine. I will close this PR. Thanks.

PR #864, opened 2022-10-05
Add new strategy wagdie-subgraph [wagdie-subgraph]

IcculusHerEmissary:
This is a modified version of `balance-of-subgraph` to track users WAGDIE balances via snapshot. 

Changes proposed:
* Add wagdie-subgraph strategy to compute votes using the `ownedWAGDIE` account value from the wagdieworld subgraph.

### DIFF
--- a/src/strategies/wagdie-subgraph/index.ts
+++ b/src/strategies/wagdie-subgraph/index.ts
@@ -5,7 +5,7 @@ const SUBGRAPH_URL = {
   '1': 'https://api.thegraph.com/subgraphs/name/wagdie/wagdieworld-mainnet'
 };
 
-export const author = 'icculus';
+export const author = 'IcculusHerEmissary';
 export const version = '0.1.0';
 
 export async function strategy(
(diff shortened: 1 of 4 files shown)


src/strategies/wagdie-subgraph/index.ts
```diff
@@ -0,0 +1,53 @@
+import { getAddress } from '@ethersproject/address';
+import { subgraphRequest } from '../../utils';
+
+const SUBGRAPH_URL = {
+  '1': 'https://api.thegraph.com/subgraphs/name/wagdie/wagdieworld-mainnet'
+};
+
+export const author = 'icculus';
```
ChaituVR, line 8:
```suggestion
export const author = 'IcculusHerEmissary';
```

ChaituVR (changes requested):
Everything else looks good :)

src/strategies/wagdie-subgraph/index.ts
IcculusHerEmissary: Updated, thank you.

ChaituVR (approved):

PR #871, opened 2022-10-07
Add strategy [orange-reputation-campaign-id-weighted]

MantisClone:
- Add the `orange-reputation-campaign-id-weighted` strategy.


MantisClone:
I'm abandoning this PR because the Orange Reputation NFTs don't support the `ERC721Enumerable` interface. 

Given that there are already over 17k OrangeReputation NFTs, and Snapshot has an effective cap of 32,000 calls (details below), I question how long a snapshot strategy could remain viable.

Snapshot uses Multicall to batch discrete calls together into 1 big call, but their node providers enforce a maximum call size so Snapshot paginates - limits each Multicall to 8000 calls. Then, they limit each strategy to about 4 or 5 Multicalls to protect their servers from DoS attacks. 8000 * 4 = 32000

Issue #688, opened 2022-07-03
Deprecate strategies "multichain" and "pagination"


bonustrack:

zzuziak:
Hey @ChaituVR what's the status on this issue?

PR #885, opened 2022-10-13
Remove pagination strategy from strategies

### LINKED ISSUE #834: Deprecate strategy "pagination"

emanuel-sol:
Fixes #834 .

- Remove pagination from strategies 
- Import pagination code from snapshot-labs/score-api
- Add score-api repo as a dependency


ChaituVR (changes requested):
These changes don't make sense to me, These changes work in your local? 
i see `Cannot find module '@snapshot-labs/score-api' or its corresponding type declarations.`

There is not much in the original issue description but ideally, it should be like this:
- [ ] Make it impossible to select this strategy in space settings on UI and hub
as mentioned here https://github.com/snapshot-labs/snapshot-strategies/issues/688
- [ ] Once we do the announcement about this, we can remove this strategy

emanuel-sol:
Ah okay that makes much more sense when you put it that way. I'll refactor my changes accordingly.

emanuel-sol:
PR https://github.com/snapshot-labs/snapshot/pull/3229 removes the pagination strategy from the UI. Technically someone can still update their space strategy setting by calling the hub api, but I feel like removing the strategy from the UI is sufficient enough. I would like to hear your thoughts.

PR #891, opened 2022-10-19
Refactor validations to class and change params

bonustrack:


src/validations/aave/index.ts
```diff
@@ -1,55 +1,50 @@
...
-  return true;
+    // Use the proposition power instead of voting power
+    if (aaveGovernanceStrategyIndex >= 0) {
+      strategies[aaveGovernanceStrategyIndex].params.powerType = 'proposition';
+    }
+
+    if (minScore) {
+      const scores = await getScoresDirect(
+        this.space,
+        strategies,
```
ChaituVR, line 34: Can we add a limit to strategies here in all validation or inside `getScoresDirect` ?

ChaituVR (changes requested):

src/validations/validation.ts
```diff
@@ -0,0 +1,29 @@
...
+    this.author = author;
+    this.space = space;
+    this.network = network;
+    this.snapshot = snapshot;
+    this.params = params;
+  }
+
+  async validate(): Promise<boolean> {
+    return true;
+  }
```
ChaituVR, line 28: Why do we need this? 🤔 it always return true?

bonustrack: Its just to set the class format

ChaituVR (approved):

PR #892, opened 2022-10-20
[thales] Update subgraph link for Thales strategy

vpoklopic:
- Update subgraph link for Thales strategy


src/strategies/thales/index.ts
```diff
@@ -6,7 +6,7 @@ export const author = 'vpoklopic';
 export const version = '1.0.1';
```
ChaituVR, line 6:
```suggestion
export const version = '1.0.2';
```

ChaituVR (approved):

PR #898, opened 2022-10-23
[spreadsheet] Add strategy spreadsheet

bonustrack:
This strategy use a Spreadsheet to define voting power


src/strategies/spreadsheet/index.ts
```diff
@@ -0,0 +1,52 @@
+import fetch from 'cross-fetch';
+
+export const author = 'bonustrack';
+export const version = '0.1.0';
+
```
ChaituVR, line 5: export const dependOnOtherAddress = false;

ChaituVR (approved):

src/strategies/spreadsheet/index.ts
ChaituVR, line 5:
```suggestion
export const dependOnOtherAddress = false;
```

PR #900, opened 2022-10-24
Ens all club digit strategy [ens-all-club-digits]

onigiri-x:
Adding this new strategy for ens digits clubs

There is already a ens-10k-club but this is a weighted vote that is used by that club to have different token classifications vote in a single weighted vote with different weights within.

This is just for having votes for 3 digit numbers (000-999) or any x digits , for example 5 digit numbers (00000-99999).

Please let me know if there are any issues


src/strategies/ens-all-club-digits/index.ts
```diff
@@ -0,0 +1,76 @@
...
+      .map((q, i) => [
+        q,
+        {
+          __aliasFor: 'registrations',
+          __args: {
+            block: snapshot !== 'latest' ? { number: snapshot } : null,
+            where: {
+              registrant_in: pages[i].map((address) => address.toLowerCase())
+            },
+            first: 1000,
```
ChaituVR, line 45: Hi @onigiri-x What happens if the registrant has more than 1000 registrations? Maybe we should paginate here

src/strategies/ens-all-club-digits/index.ts
```diff
@@ -0,0 +1,76 @@
...
+  provider,
+  addresses,
+  options,
+  snapshot
+) {
+  const { numberOfDigits } = options;
+  // 999 Club = 3 , 10k club = 4, 100k club = 5
+  const NAMES_CLUB = [...Array(Math.pow(10, numberOfDigits)).keys()]
+    .map((i) => -1 + i + 1)
+    .map((x) => x.toString().padStart(numberOfDigits, '0'));
```
ChaituVR, line 26: I feel like someone could pass a huge number and it can cause memory issues here, maybe there should be a limit for this

src/strategies/ens-all-club-digits/index.ts
```diff
@@ -0,0 +1,76 @@
...
+
+const ENS_SUBGRAPH_URL = {
+  '1': 'https://api.thegraph.com/subgraphs/name/ensdomains/ens',
+  '3': 'https://api.thegraph.com/subgraphs/name/ensdomains/ensropsten',
+  '4': 'https://api.thegraph.com/subgraphs/name/ensdomains/ensrinkeby',
+  '5': 'https://api.thegraph.com/subgraphs/name/ensdomains/ensgoerli'
+};
+
+export const author = '3757';
+export const version = '0.1.1';
```
ChaituVR, line 12:
```suggestion
export const author = 'onigiri-x';
export const version = '0.1.0';
```

ChaituVR (changes requested):

src/strategies/ens-all-club-digits/index.ts
onigiri-x: Hey Chaitu, this is an excellent question. So what I did is I changed to 4 digit and I found an account with over 1000 ENS names, and just as you predicted the snapshot test broke. I added in a new commit some pagination, I followed loosely how other strategies added pagination. In the end by updating this it is all good. All names should be included.

src/strategies/ens-all-club-digits/index.ts
onigiri-x: Thanks for pointing this out. I have tested this, and up to 7 digits it worked like a charm on my machine. At 8 digits it kind of got "stuck", so I made it support 3 digits (lowest number ENS has) up to 7 digits. Any other input will not work into this field

We principally need this snapshot for 3,4,5 digit numbers so it should be all good

src/strategies/ens-all-club-digits/examples.json
```diff
@@ -0,0 +1,20 @@
+[
+  {
+    "name": "ENS ALL Club Digits",
+    "strategy": {
+      "name": "ens-all-club-digits",
+      "params": {
+        "address": "0x57f1887a8BF19b14fC0dF6Fd9B2acc9Af147eA85",
+        "symbol": "NUMS",
+        "numberOfDigits": 4
```
ChaituVR, line 9: changing this to test
```suggestion
        "numberOfDigits": 7
```

src/strategies/ens-all-club-digits/index.ts
ChaituVR: @onigiri-x Can make it max to 5? I tried changing it to 7 and the tests are failing in GitHub actions. 

Or you see any alternate way to check with out adding all possible values to an array? we can't check response directly?

onigiri-x:
@ChaituVR can you check now please? Solution is much more simple now

ChaituVR (approved):
LGTM :) Thanks for the changes @onigiri-x

PR #904, opened 2022-10-25
Staked-Balancer: remove unused variable in subgraph call [staked-balancer]

lukaszh2o:
Fixes not working staked-balancer strategy. Root cause for the issue is in logic responsible for making calls to Balancer subgraph, where it tries to query address field on Pool entity, which is present in subgraph of Balancer V2, but not in subgraph for Balancer V1. However, querying address field can be removed, as it's not being used inside strategy. 

- remove querying address field of Pool entity in Subgraph api call.

### DIFF
--- a/src/strategies/staked-balancer/index.ts
+++ b/src/strategies/staked-balancer/index.ts
@@ -30,7 +30,6 @@ const params = {
   pool: {
     __args: { id: '' },
     totalShares: true,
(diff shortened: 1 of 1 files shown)


ChaituVR (approved):
I can reproduce the error, Merging it

cc @jo-chemla

PR #907, opened 2022-10-26
Add claimDateLimit parameter to safe-vested strategy [safe-vested]

dasanra:
Update to `safe-vested` strategy to consider a claim date limit in the vesting contracts.

- Add a new parameter `claimDateLimit` that allows configuring a date limit when the vesting is allowed

### DIFF
--- a/src/strategies/safe-vested/index.ts
+++ b/src/strategies/safe-vested/index.ts
@@ -4,7 +4,7 @@ import { formatUnits, parseUnits } from '@ethersproject/units';
 import { Multicaller } from '../../utils';
 
 export const author = 'dasanra';
-export const version = '0.1.0';
+export const version = '0.2.0';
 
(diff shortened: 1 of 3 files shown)


ChaituVR:
There are few active proposals but these changes suggest this is just an optional parameter. so should be fined right?

ChaituVR (approved):

src/strategies/safe-vested/index.ts
```diff
@@ -6,7 +6,7 @@ import { Multicaller } from '../../utils';
 export const author = 'dasanra';
 export const version = '0.1.0';
```
ChaituVR, line 7:
```suggestion
export const version = '0.1.1';
```

dasanra: Thank you for suggesting this. I will set minor bump as I'm adding a new parameter and we try to follow SemVer.

I think the current active proposals are in a test space but anyway, it shouldn't break if the parameter is not configured 👍

PR #911, opened 2022-10-26
Add Metropolis Pod strategy [metropolis-pod]

itsdanwu:
This adds a new Metropolis Pod strategy intended to replace the Orca Pod strategy, as we have rebranded to Metropolis. 

We should make sure the Orca Pod strategy does not display in UI to not confuse users.

Have confirmed tests are passing.

### DIFF
--- a/src/strategies/metropolis-pod/index.ts
+++ b/src/strategies/metropolis-pod/index.ts
@@ -1,0 +1,28 @@
+import { strategy as erc1155BalanceOfIdsWeightedStrategy } from '../erc1155-balance-of-ids-weighted';
+
+export const author = 'snapshot-labs';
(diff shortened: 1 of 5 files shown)


src/strategies/metropolis-pod/index.ts
```diff
@@ -0,0 +1,28 @@
+import { strategy as erc1155BalanceOfIdsWeightedStrategy } from '../erc1155-balance-of-ids-weighted';
+
+export const author = 'snapshot-labs';
```
ChaituVR, line 3:
```suggestion
export const author = 'itsdanwu';
```

ChaituVR (approved):

PR #912, opened 2022-10-26
Fix lowercase address issue with getvp

ChaituVR:


ChaituVR:
> Did you tried if it doesn't break delegation (which uses subgraph with lowercase addresses)?

Tried this, and we are lowercasing addresses before sending them to the subgraph, so should be fine, also we are sending check summed addresses from UI, so if some strategy doesn't handle this would have failed by now

PR #918, opened 2022-10-29
Passport validation - check with snapshot block

ChaituVR:


src/validations/passport/index.ts
```diff
@@ -9,19 +14,30 @@ export default class extends Validation {
...
 
     let weight = 0;
     this.params.stamps.forEach((stamp: any) => {
-      const found = verifiedStamps.find((s: any) => s.provider === stamp.id);
-      if (found?.verified) weight += stamp.weight;
+      const verifiedStamp = verifiedStamps.find(
+        (s: any) => s.provider === stamp.id
+      );
+
+      // check that the credential is still valid (created after snapshot block and not expired)
```
bonustrack, line 35: What you mean "created after snapshot block", we should check it user were attested before or in the block number, not after

ChaituVR: oops , typo in comment, fixed it 👍

bonustrack:
So this does not address the issue with multiple addresses having same attestation right?

ChaituVR:
Yep this is just to enforce snapshot block

bonustrack (approved):
utACK

PR #923, opened 2022-10-31
New strategy: linear-vesting-power [linear-vesting-power]

julien-devatom:
## New strategy

- new strategy which is an evolution of `balance-of-with-vesting-power`. Because Morpho plans to use multiple Vesters with only one balance of, we need to create a strategy adding the voting power from one vesting line, and we will batch the strategies to add multiple vesters + `balance-of`


test/strategy.test.ts
```diff
@@ -39,6 +39,9 @@ function callGetScores(example) {
   );
 }
 
+const now = Date.now();
+Date.now = jest.fn(() => now);
+
```
ChaituVR, line 44: what is this for? 🤔

julien-devatom: Explained into the description of the ticket

julien-devatom: You added a test for the check "Voting power should not depend on other addresses", which is recomputing the distribution. However, we use Date.now() in the strategy, so the result is different. To fix it, I suggest to mock Date.now() to return always the same timestamp

julien-devatom: We could have had this choice also in the contract:

```solidity
 if(isSupplyDisabled) revert OperationDisabled();
```
With the same error for each operation
But the design of the contract is different (one error per operation), and IMO for clear understanding of morpho-js, it is a good thing to reproduce the errors of the Morpho contract.

ChaituVR (changes requested):
Pasting here from TG

Voting power should not change based on time, instead, it should be dependent on the snapshot block number, can you please change it to use block timestamp instead of date.now?

src/strategies/index.ts
```diff
@@ -447,8 +448,7 @@ const strategies = {
   'protofi-erc721-tier-weighted': protofiErc721TierWeighted,
   'erc721-with-tokenid': erc721WithTokenId,
   'erc721-with-tokenid-range-weights': erc721WithTokenIdRangeWeights,
-  'erc721-with-tokenid-range-weights-simple':
-    erc721WithTokenIdRangeWeightsSimple,
+  'erc721-with-tokenid-range-weights-simple': erc721WithTokenIdRangeWeightsSimple,
```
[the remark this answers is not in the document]
julien-devatom: yes sorry!

ChaituVR (changes requested):
@julien-devatom It looks good, except that it crosses the 5 requests limit 🙈 , 

Can combine `multiVestTotCaller` and `multiVestAccruedCaller` into a single multicall and separate result? will reduce the number of requests and everything will be good :)

julien-devatom:
> @julien-devatom It looks good, except that it crosses the 5 requests limit see_no_evil ,
> 
> Can combine `multiVestTotCaller` and `multiVestAccruedCaller` into a single multicall and separate result? will reduce the number of requests and everything will be good :)

Yes, fixed and moved to 3 queries

ChaituVR (approved):

PR #925, opened 2022-11-01
[posichain-staking] Add posichain-staking strategy

dannyposi:
- Add posichain-staking strategy

### DIFF
--- a/src/strategies/posichain-staking/index.ts
+++ b/src/strategies/posichain-staking/index.ts
@@ -1,0 +1,38 @@
+import { StaticJsonRpcProvider } from '@ethersproject/providers';
+import { formatUnits } from '@ethersproject/units';
+import { BigNumber } from '@ethersproject/bignumber';
+
+export const author = 'posichain';
+export const version = '0.0.1';
+
+type Params = {
+  symbol: string;
+  decimals: number;
+};
+
+export async function strategy(
+  _space: string,
+  _network: string,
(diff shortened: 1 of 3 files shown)


src/strategies/posichain-staking/index.ts
```diff
@@ -0,0 +1,38 @@
...
+type Params = {
+  symbol: string;
+  decimals: number;
+};
+
+export async function strategy(
+  _space: string,
+  _network: string,
+  provider: StaticJsonRpcProvider,
+  _addresses: Array<string>,
```
ChaituVR, line 17: You were not using _addresses? The response doesn't depend on addresses.

ChaituVR (changes requested):

src/strategies/posichain-staking/index.ts
dannyposi: I've add comment to explain about that. I see strategy caller has filtered out all addresses that not needed: https://github.com/snapshot-labs/snapshot-strategies/blob/02439eb120ed7c4cc0c493924b78d92d22006b40/src/utils.ts#L26

ChaituVR (approved):

dannyposi:
@ChaituVR Thank for merging. But when it available for choosing?

ChaituVR:
It is live now https://snapshot.org/#/strategy/posichain-staking

PR #929, opened 2022-11-01
New Strategy: Decentraland Rental Lessors [decentraland-rental-lessors]

fzavalia:
# Decentraland Rental Lessors Strategy

Once a rental starts using the [Rentals](https://github.com/decentraland/rentals-contract/blob/main/contracts/Rentals.sol) contract, the asset is transferred from the owner to the rentals contract.

This means that any VP that the previous user had due to having their Lands or Estates in their possession is lost.

This strategy attempts to avoid losing that VP by querying a couple of subgraphs, obtaining the original owners of the assets sent to the rentals contract, and give those owners their respective VP.

## How it works

When a rental starts, the Land or Estate is transferred to the Rentals contract. The original owner of the asset is tracked as a `lessor`, and that information is indexed in the rentals subgraph.

The strategy will first query the rentals subgraph with a list of addresses to obtain the Lands and Estates those addresses are `lessors` of.

With that information, the strategy will do different things for Lands and Estates.

For Lands, The lessors will be given VP according to each Land they have in the Rentals contract multiplied by a configured factor.

For Estates, the strategy needs to find the size of those Estates first, the size being the amount of Lands it is composed of. To do so, it fetches that data from the marketplace subgraph. Once the information is complete, each lessor of an Estate is given VP according to the size of the Estate multiplied by a configured factor.


fzavalia:
Ready for review

fzavalia:
@2fd Would you mind taking a look?

ChaituVR (changes requested):
@fzavalia Test cases are failing (taking more time) Can take a look?

fzavalia:
> @fzavalia Test cases are failing (taking more time) Can take a look?

The tests seem to fail because of timeouts caused by the queries to the graph, which has been working pretty slow the last hours.

Can we run the tests again?

ChaituVR (approved):

PR #940, opened 2022-11-08
Separate passport validations

ChaituVR:


samuveth (approved):
utACK

src/validations/passport-gated/schema.json
```diff
@@ -0,0 +1,30 @@
...
+  "definitions": {
+    "Validation": {
+      "title": "Validation",
+      "type": "object",
+      "properties": {
+        "stamps": {
+          "type": "array",
+          "title": "Stamps",
+          "minItems": 1,
+          "maxItems": 32
```
samuveth, line 13: `items` seems to be missing here with `type` string.

PR #941, opened 2022-11-09
feat(strategy): ari10 tokens locked in staking [ari10-staking-locked]

oritwoen:
This strategy is related to the token: https://ari10.com/token and its staking contracts.

In practice, it is supposed to work by checking the contract address and returning the appropriate number of votes multiplied by the weight, depending on how long the holder locked the tokens on the contract.


ChaituVR (changes requested):
Hi, @oritwoen I feel like you can use `contract-call` strategy with `math` strategy. Can try it?

oritwoen:
@ChaituVR It's possible this will work. Can I have an example though? Because it's hard for me to understand how to combine such 2 existing strategies that are based only on JSON without UI.

I'm not even asking for a code, just simple information on how to combine 2 strategies of this type.

**This strategy (this PR) was created mainly for easy space management by administrators who have no contact with programming. The idea was that the administrator could change the weight of votes, change or add a new contract at any time using a simple interface and website. What this PR enables.

So even if the combination of the above strategies you are talking about works, I would be in favor of merging it for the convenience and possibility of administrators. So that the "management" does not have to ask someone from the IT department to change the JSON code every time he wants to change something in the counting of votes.**

ChaituVR (approved):

PR #958, opened 2022-11-21
fix: [poap-with-weight-v2] voting power

greenealexander:
Fixes `poap-with-weight-v2` voting power calculation

- the addresses in `addressesMap` must be checksum addresses
  - map lowercase address to valid address with `getAddress()`
- added test for voting power

Screen Shot 2022-11-21 at 02 16 25


src/strategies/poap-with-weight-v2/index.ts
```diff
@@ -1,3 +1,5 @@
+import { getAddress } from '@ethersproject/address';
+
```
ChaituVR, line 2:
```suggestion
import { getAddress } from '@ethersproject/address';
```

src/strategies/poap-with-weight-v2/index.ts
```diff
@@ -1,3 +1,5 @@
...
 import { subgraphRequest } from '../../utils';
```
ChaituVR, line 3:
```suggestion
import { getAddress } from '@ethersproject/address';
import { subgraphRequest } from '../../utils';
```

ChaituVR (approved):
@greenealexander Thanks for adding the test cases 🙏 but we want to rethink the testing approach with the lowercased addresses here, it can take some time so removed those changes for now :)

PR #959, opened 2022-11-21
ERC20 strategy with blacklist

belbix:
It is a simple strategy for handling ERC20 balances with the possibility of excluding some predefined addresses.
Very simple and useful, I think will be good to have this strategy in the main repo

### DIFF
--- a/src/strategies/erc20-balance-of-delegation-with-blacklist/examples.json
+++ b/src/strategies/erc20-balance-of-delegation-with-blacklist/examples.json
@@ -1,0 +1,28 @@
+[
+  {
+    "name": "Example query",
+    "strategy": {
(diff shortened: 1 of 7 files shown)


ChaituVR:
Hi @belbix We are thinking for ways to reduce dependency on this delegation strategy, this may add up some more confusion, We are adding a `validation` feature where you can decide who can vote who cannot similar to strategies. should be available soon, can wait until then? :)

belbix:
@ChaituVR it's pretty urgent, requires for Balancer snapshot space https://forum.balancer.fi/t/bip-xxx-aura-multisig-blacklisting-for-governance/4023/3?u=belbix

ChaituVR:
@bonustrack What you think?

PR #961, opened 2022-11-21
Delegation strategies to ignore override

ChaituVR:
Fixes https://github.com/snapshot-labs/snapshot-strategies/issues/957


bonustrack (approved):
utACK

PR #977, opened 2022-12-02
One-POAP-One-Vote [poap]

gawainb:
Fixes # New Voting Strategy.

- A new voting strategy that allows 1 POAP for 1 vote.

### DIFF
--- a/src/strategies/poap/index.ts
+++ b/src/strategies/poap/index.ts
@@ -1,70 +1,27 @@
 import { getAddress } from '@ethersproject/address';
-import { subgraphRequest } from '../../utils';
+import { strategy as erc721Strategy } from '../erc721';
 
 export const author = 'greenealexander';
 export const version = '1.0.0';
-
-const POAP_API_ENDPOINT_URL = {
-  '1': 'https://api.thegraph.com/subgraphs/name/poap-xyz/poap',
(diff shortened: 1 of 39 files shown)


src/strategies/poap/index.ts
```diff
@@ -0,0 +1,70 @@
...
+const MAX_ACCOUNTS_IN_QUERY = 500;
+
+export async function strategy(
+  space,
+  network,
+  provider,
+  addresses,
+  options,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  snapshot
```
ChaituVR, line 21: @gawainb Make sure you are using `snapshot` else the result of the proposal may keep changing.

ChaituVR (changes requested):

ChaituVR (changes requested):
i think for this `erc721` strategy should work fine too, just pass in the POAP contract into it. let me know if it doesn't work

src/strategies/poap/index.ts
```diff
@@ -0,0 +1,27 @@
+import { getAddress } from '@ethersproject/address';
+import { strategy as erc721Strategy } from '../erc721';
+
+export const author = 'greenealexander';
```
ChaituVR, line 4:
```suggestion
export const author = 'gawainb';
```

PR #996, opened 2022-12-20
Added multichain-min-balance strategy

unblocktechie:
Fixes # .

- 
- 
-

### DIFF
--- a/src/strategies/index.ts
+++ b/src/strategies/index.ts
@@ -45,6 +45,7 @@ import * as ethBalance from './eth-balance';
 import * as ethWithBalance from './eth-with-balance';
 import * as ethWalletAge from './eth-wallet-age';
 import * as multichain from './multichain';
+import * as multichainMinBalance from './multichain-min-balance';
 import * as gooddollarMultichain from './gooddollar-multichain';
 import * as makerDsChief from './maker-ds-chief';
 import * as uni from './uni';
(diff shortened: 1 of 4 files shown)


ChaituVR (changes requested):
@unblocktechie These type of strategies doesn't scale well, you can try using https://snapshot.org/#/strategy/balance-of-with-min with different chains

unblocktechie:
@ChaituVR The proposed strategy is going to be used by Alethea AI protocol. The strategy mentioned by you is relevant if token exists only in one chain. We want to introduce a cumulative threshold for tokens held across all the chains. If we use balance-of-with-min strategy with multichain, we need to define minBalance for each chain. We want to implement a strategy that calculate user's balance across all chain and the final score will be considered for voting if it is more than minimum limit.

ChaituVR:
@unblocktechie I think you can achieve this by using `basic` voting validation
https://github.com/snapshot-labs/snapshot-strategies/blob/master/src/validations/basic/index.ts
This is not visible on UI yet. will update you once it is available.

ChaituVR:
Know more about validations here https://docs.snapshot.org/strategies/what-is-a-strategy-1

unblocktechie:
@ChaituVR Can you provide approximate timeline for basic validation option will be reflected in UI. So that we can plan our strategy accordingly.

PR #1000, opened 2022-12-23
add: erc721 modification

pro100skm:
- Added ERC-721 strategy with custom weight settings

### DIFF
--- a/src/strategies/erc721-with-method-weight/index.ts
+++ b/src/strategies/erc721-with-method-weight/index.ts
@@ -1,0 +1,46 @@
+import { multicall } from '../../utils';
+import { Options } from './types';
+
+export const author = 'pro100skm';
+export const version = '0.0.1';
+
+const abi = (methodName: string) => [
+  `function ${methodName}(address addr, uint256 level) public view returns (uint256)`
+];
+
+export async function strategy(
+  space,
+  network,
+  provider,
+  addresses,
+  options: Options,
+  snapshot
+) {
+  const blockTag = typeof snapshot === 'number' ? snapshot : 'latest';
+  const { address: contractAddress, methodName, levelsWeight } = options;
+
+  const keys = Object.keys(levelsWeight);
+
+  let weightSum = Array(keys.length).fill(0);
+
+  for (let i = 0; i < keys.length; i++) {
+    const key = keys[i];
+    const weight = levelsWeight[key];
+    const response = await multicall(
+      network,
+      provider,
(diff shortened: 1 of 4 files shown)


ChaituVR:
Is it still WIP?

pro100skm:
> Is it still WIP?

no, its already working code. 
I just followed the instruction and set it WIP

ChaituVR (changes requested):
Hey @pro100skm Thanks for clarifying, need to add this strategy to `src/strategies/index.ts` Also make sure to run `yarn test --stratergy=erc721-with-method-weight` and see if it is passing all test cases

src/strategies/erc721-with-method-weight/index.ts
```diff
@@ -0,0 +1,46 @@
...
+  snapshot
+) {
+  const blockTag = typeof snapshot === 'number' ? snapshot : 'latest';
+  const { address: contractAddress, methodName, levelsWeight } = options;
+
+  const keys = Object.keys(levelsWeight);
+
+  let weightSum = Array(keys.length).fill(0);
+
+  for (let i = 0; i < keys.length; i++) {
```
ChaituVR, line 26: Need to add a limit to these `keys` else it may cause some memory issues, and throw an error if it receives more keys

pro100skm: I just simplified the logic of fetching weights

pro100skm:
> Hey @pro100skm Thanks for clarifying, need to add this strategy to `src/strategies/index.ts` Also make sure to run `yarn test --stratergy=erc721-with-method-weight` and see if it is passing all test cases

yeah. thank you. I tested

src/strategies/erc721-with-method-weight/index.ts
```diff
@@ -0,0 +1,38 @@
...
+    multi.call(address, contractAddress, methodName, [address])
+  );
+  const result: Record<string, BigNumberish> = await multi.execute();
+
+  return Object.fromEntries(
+    Object.entries(result).map(([address, balance]) => [
+      address,
+      parseFloat(formatUnits(balance, 0))
+    ])
+  );
```
ChaituVR, line 37: Hi @pro100skm The method `getPowerOf` itself will return `balance * weight`? then better to use `contract-call` strategy directly https://snapshot.org/#/strategy/contract-call

ChaituVR (changes requested):

PR #1002, opened 2022-12-27
Add DIMO Voting Strategy [dimo-balance-of-or-votes-with-overrides]

zer0stars:
-  Create new strategy for DIMO to vote on snapshot
-  Reuse the existing erc20votes with override to enable the logic.


src/strategies/dimo-balance-of-or-votes-with-overrides/index.ts
```diff
@@ -0,0 +1,90 @@
+import { BigNumberish } from '@ethersproject/bignumber';
+import { formatUnits } from '@ethersproject/units';
+import { getAddress } from '@ethersproject/address';
+import { Multicaller } from '../../utils';
+import { strategy as erc20VotesWithOverrideStrategy } from '../erc20-votes-with-override';
+
+export const author = '0xMaharishi';
```
ChaituVR, line 7:
```suggestion
export const author = 'zer0stars';
```

src/strategies/dimo-balance-of-or-votes-with-overrides/index.ts
```diff
@@ -0,0 +1,90 @@
...
+  space,
+  network,
+  provider,
+  addresses,
+  options: Options,
+  snapshot
+) {
+  const blockTag = typeof snapshot === 'number' ? snapshot : 'latest';
+  const checksummedAddresses = addresses.map((address) => getAddress(address));
+  const scores: Record<string, number> = await erc20VotesWithOverrideStrategy(
```
ChaituVR, line 38: Hi @zer0stars We stopped accepting strategies that depend on other addresses, (erc20-votes-with-override depend on other addresses) Refer this https://github.com/snapshot-labs/snapshot-strategies#code

Make it independent such that it doesn't call this strategy or doesn't depend on other address's voting power

ChaituVR (changes requested):

PR #1003, opened 2023-01-01
Kuma DAO strategy - [dkuma-breeder-staked]

KumaInuCommunity:
- New strategy for the Kuma DAO. Gets the amount of dKuma that a wallet has staked in the dKuma Breeder


src/strategies/dkuma-breeder-staked/index.ts
```diff
@@ -0,0 +1,44 @@
...
+      address
+    ])
+  );
+  const result: Record<string, walletInfoInt> = await multi.execute();
+
+  return Object.fromEntries(
+    Object.entries(result).map(([address, walletInfo]) => [
+      address,
+      parseFloat(formatUnits(walletInfo.amount, options.decimals))
+    ])
```
ChaituVR, line 42: Hey @KumaInuCommunity You can use https://snapshot.org/#/strategy/contract-call strategy instead. did you try it?

ChaituVR (changes requested):

src/strategies/dkuma-breeder-staked/index.ts
KumaInuCommunity: That's right, hadn't seen it. Thank you.
Now we are using 3 different strategies, all of them being default strategies of yours. Is it fine if we create a single strategy to call all of them so the input options can be simplified?

ChaituVR: @KumaInuCommunity We expect strategies to be unique :) If it is possible to achieve something with 3 strategies, Better to keep them separate so our cache systems work perfectly, it can resolve much faster and we can reduce code duplication.

VQuesada: Makes sense, let´s keep it as it is now. Thank you.

KumaInuCommunity: Makes sense, let's keep it as it is now. Thank you.

PR #1007, opened 2023-01-11
Adding Nyolings Strategy

JosephSmith127:
Adding a strategy for Nyolings which reads from both the Nyolings contract along with the staking contract


JosephSmith127:
Closing as outdated

PR #1008, opened 2023-01-11
Nyolings strategy [nyolings-strategy]

jsmmth:
Added Nyolings Strategy, reads from both the Nyolings contract along with the nyolings staking contract.


ChaituVR (changes requested):
Hey @JosephSmith127 You can use two strategies directly on your space, result will be sum of those two strategies

JosephSmith127:
Ah thank you! I didn't realise this and thought a custom strategy would be required. I'll give this a go and close the PR if it's no longer required, thanks!

PR #1021, opened 2023-01-25
Circling around on the Nation3 strategy updte

johnmark13:
This time coming from the Nation3 repository..

This is quite a substantial change that makes sure the veNation balances are retrieved for all voters and the passport owners on behalf of whom they are voting.

It is also refactored to make the code more efficient in building the aggregated voting weight of a voter.

Don't just pull the balance of voters from the veNation contract, but also for the accounts on which they vote on behalf of
Remove the loop with the if delegated clause and replace with a map of voting address -> passport holder addresses for the calculation of voting weight


ChaituVR (changes requested):
Thanks @johnmark13 :) Can bump up the version?

```
```

ChaituVR (approved):

PR #1023, opened 2023-01-27
[moonbeam-free-balance] Adds moonbeam free balance strategy

crystalin:
Adds the **moonbeam free strategy**, relying on the Moonbeam Substrate API, querying the storage key associated with the given account address;

In order to compute the storage key of the address, the blake2 algorithm must be used. 
This PR adds the **blake2 library** directly in the strategy folder, any other suggestion to improve this is welcome.


ChaituVR (changes requested):
Hey @crystalin PR should not contain any other dependencies with external libraries. Can do it without a library?

crystalin:
@ChaituVR Unfortunately, there is the need to execute blake2 to retrieve the data for the given address. 
I could copy the blake2 library source instead of the minified version, would that work ?

ChaituVR:
If it is a small code, follows checklist and is reviewable then it should be fine I think but if it does some complicated calculations better to have it on a separate server and call the API from here. because it could potentially cause slow down to others 🙈

crystalin:
@ChaituVR 
I've replaced it with a pure js function. The code is fairly small and easily reviewable in my opinion.
Let me know if that is enough.

ChaituVR:
Still big i am worried about the computation required for this 🙈 also there is some error when we run test cases . can check?

crystalin:
I fixed the lint and slightly reduced blake2 to the minimum.
Is there anything else I could do ? The support of snapshot would be very valuable for the chain

src/strategies/moonbeam-free-balance/index.ts
```diff
@@ -0,0 +1,56 @@
...
+        const free = readLittleEndianBigInt(
+          result.substring(2 + 8 + 8 + 8 + 8, 2 + 8 + 8 + 8 + 8 + 32)
+        );
+
+        // Converts the bigint into number.
+        // Result won't be precise, but should be lower than real value.
+        return [getAddress(address), Number(free)];
+      })
+    )
+  );
```
ChaituVR, line 55: @crystalin It is not recommended to send one request per one address. try to use a multicall like other strategies to call `state_getStorage`

ChaituVR (changes requested):

crystalin:
@ChaituVR I made the change to support batches and merged master, let me know if I can do more

ChaituVR:
Hey @crystalin Sorry for the delay and sorry for not looking into blakejs before, https://www.npmjs.com/package/blakejs This looks great, we can use it directly instead of a separate file. sorry again :) let me know once it is added back. will merge this right away, Also better to avoid changes in `yarn.lock` don't push that change (i will add it) Thanks for your patience 🙏

src/strategies/moonbeam-free-balance/index.ts
```diff
@@ -0,0 +1,120 @@
...
+          balances[getAddress(addressBatch[(payload.id - 1) % batchSize])] =
+            free;
+        }
+        return;
+      },
+      (error) => {
+        throw error;
+      }
+    );
+  }
```
ChaituVR, line 118: Sorry one last change, don't mix up await and .then plz. strategies should always use await :) everything else looks good 👍

crystalin: Done.
There isn't much error handling directly there as I don't know if it is caught upstream in the code or not

ChaituVR (approved):

ChaituVR (approved):

crystalin:
@ChaituVR thank you

PR #1026, opened 2023-01-28
add axolittles dao strategy

AlliedToasters:
- adds axolittles dao voting strategy


ori-wagmi (approved):

src/strategies/axolittles-dao/index.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  const response = await fetch(api_url, {
+    method: 'GET',
+    headers: {
+      Accept: 'application/json',
+      'Content-Type': 'application/json'
+    }
+  });
+  const data = await response.json();
+  return data
```
ChaituVR, line 28: Hey @AlliedToasters You can directly use `api` strategy directly, no need to create a new strategy :) Can try?

ChaituVR (changes requested):

src/strategies/axolittles-dao/index.ts
AlliedToasters: Sure, trying!

PR #1029, opened 2023-01-29
Lodestar finance strategies [lodestar-staked-lp]

0xAppo:
Fixes # .

- Add staked LP strategy


src/strategies/lodestar-staked-lp/index.ts
```diff
@@ -0,0 +1,86 @@
...
+      (parseFloat(formatUnits(lpTokenReserves._reserve1, tokenDecimals)) /
+        parseFloat(formatUnits(lpTokenTotalSupply, lpTokenDecimals))) *
+      2;
+  } else {
+    tokenWeight = 0;
+  }
+
+  const multi = new Multicaller(network, provider, abi, { blockTag });
+
+  options.stakingPoolAddresses.forEach(stakingPoolAddress => {
```
ChaituVR, line 70: Hi @0xAppo Everything looks good, Can throw some error here if someone passes too many addresses? Thanks

ChaituVR (changes requested):

src/strategies/lodestar-staked-lp/index.ts
0xAppo: Hi there, would this method be acceptable?

ChaituVR (changes requested):
Need same changes as in here https://github.com/snapshot-labs/snapshot-strategies/pull/1030

0xAppo:
I think I fixed it, let me know if that works

src/strategies/lodestar-staked-lp/index.ts
```diff
@@ -0,0 +1,89 @@
...
+    if (options.stakingPoolAddresses.length >= 10) {
+      throw new Error('Too many stake pool addresses provided.')
+    }
```
ChaituVR, line 73: Can move this to before forEach?

0xAppo: Done.

ChaituVR (approved):

PR #1033, opened 2023-01-30
Feat/jbas nft [jbas-nft]

ppoliani:

New Custom Strategy

This is a custom strategy that defines the voting logic for the Japanese Born Ape Society project. 

One vote is assigned to each user that holds:

1. one unit of the JBAS [ERC721 collection](https://etherscan.io/token/0x56cA59ab1b3c7086b3c4aF417593fDeE566A3320)
2. and one unit of a specific token ID from the Japanese Apes Flower Shop [ERC1155 collection](https://etherscan.io/token/0x56cA59ab1b3c7086b3c4aF417593fDeE566A3320)

> The Japanese Apes Flower Shop has 7 different token ids [0, 6]

### DIFF
--- a/src/strategies/jbas-nft/index.ts
+++ b/src/strategies/jbas-nft/index.ts
@@ -1,0 +1,34 @@
+import { BigNumberish } from '@ethersproject/bignumber';
+import { formatUnits } from '@ethersproject/units';
+import { Multicaller } from '../../utils';
+
+export const author = 'bonustrack';
+export const version = '0.1.1';
+
+const abi = [
+  'function balanceOf(address account) external view returns (uint256)'
+];
+
+export async function strategy(
+  space,
+  network,
+  provider,
+  addresses,
+  options,
+  snapshot
+): Promise<Record<string, number>> {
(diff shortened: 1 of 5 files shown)


src/strategies/jbas-nft/index.ts
```diff
@@ -0,0 +1,96 @@
...
+
+      return acc;
+    },
+    {}
+  );
+
+  return Object.fromEntries(
+    Object.entries(balances).map(([address, balance]) => [
+      address,
+      Math.min(balance.erc721, balance.erc1155)
```
ChaituVR, line 93: Hey @ppoliani I think math strategy (with min operation) should work great in this case, instead of a separate Strategy. can try that?

ChaituVR (changes requested):

src/strategies/jbas-nft/index.ts
ppoliani: oh this is a good point. I did a bit of research before writing this strategy. Sorry for the PR. The math strategy works fine!

ChaituVR: NP :) Thanks for trying it 👍 Will close this PR then

PR #1041, opened 2023-02-03
[gains-vault-staked-balance] Gains Vault Staked Balance

onecri:
-  Create new strategy for Gains to vote on snapshot
- 
-


src/strategies/gains-vault-staked-balance/index.ts
```diff
@@ -0,0 +1,34 @@
...
+  );
+  const result: Record<string, BigNumberish> = await multi.execute();
+
+  return Object.fromEntries(
+    Object.entries(result).map(([address, res]) => [
+      address,
+      parseFloat(formatUnits(res[0], options.decimals))
+    ])
+  );
+}
```
ChaituVR, line 34: @onecri I think the `contract-call` strategy should work in this case, no need for a separate strategy :) Can try? let me know

ChaituVR (changes requested):

src/strategies/gains-vault-staked-balance/index.ts
onecri: HI @ChaituVR  i want to get our own separate strategy. Is it impossible?

ChaituVR: @onecri Is there any reason why?

ChaituVR: We wanted to reduce duplicate strategies as much as possible. so we added a checklist here https://github.com/snapshot-labs/snapshot-strategies#code

onecri: @ChaituVR there is no special reason, but we want to have our own.

ChaituVR: I understand that you want to have your own strategy easily identifiable by the space's name, however, we highly value the core development principles of keeping the code DRY. We are aiming to minimize the repetitive code and at the same time provide the best experience to our users (lack of confusion regarding the difference between two strategies with the same functionality, reusability of available tools, recognition of popular strategies among voters, etc). 

If the strategy requires custom implementation, then it makes total sense to create a new one however in your case we would appreciate it if you could use what has already been implemented and tested by multiple spaces.

onecri: Ok, got it

onecri: image

onecri: I am trying with `contract-call`, but i got this error. I believe this error is because the length of the return value of `userDeposit` function is 2 ([uint256, uint256]). Is there any way to get only 1st value from returned array? @ChaituVR

onecri: Ok, thank you

PR #1053, opened 2023-02-12
[umami-voting] Added unstaked Umami in the existing umami-voting strategy

umamianon:
- Added unstaked Umami in the total vote count for umami-voting.


ChaituVR (changes requested):
@EncryptedBunny WYT? Also is it better to change author to `UmamiDAO`?

umamianon:
> @EncryptedBunny WYT? Also is it better to change author to `UmamiDAO`?

Actually, someone has informed me we can add multiple strategies. Will just combo with basic erc20 vote. Closing PR, thanks.

PR #1057, opened 2023-02-14
Adds Hedgey delegate strategy [hedgey-delegate]

bark4mark:
- Creates a strategy to allow votes from a locked hedgey where the locked token has delegated to another address.

### DIFF
--- a/src/strategies/hedgey-delegate/index.ts
+++ b/src/strategies/hedgey-delegate/index.ts
@@ -1,5 +1,4 @@
-import { BigNumberish } from '@ethersproject/bignumber';
-import { formatUnits } from '@ethersproject/units';
+import { BigNumber, BigNumberish } from '@ethersproject/bignumber';
 import { Multicaller } from '../../utils';
 
 export const author = 'hedgey';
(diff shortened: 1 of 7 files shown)


src/strategies/hedgey-delegate/index.ts
```diff
@@ -0,0 +1,37 @@
+import { BigNumber, BigNumberish } from '@ethersproject/bignumber';
+import { Multicaller } from '../../utils';
+
+export const author = 'hedgey';
```
ChaituVR, line 4:
```suggestion
export const author = 'hedgey-finance';
```

src/strategies/hedgey-delegate/index.ts
```diff
@@ -0,0 +1,37 @@
...
+      options.hedgeyNFT
+    ])
+  );
+  const result: Record<string, BigNumberish> = await multi.execute();
+
+  return Object.fromEntries(
+    Object.entries(result).map(([address, votes]) => [
+      address,
+      BigNumber.from(votes).toNumber()
+    ])
```
ChaituVR, line 35: I think `contract-call` strategy should work in this case, instead of new strategy, can try it? let me know if you have any questions

ChaituVR (changes requested):
package-lock.json should not contain any changes

src/strategies/hedgey-delegate/index.ts
bark4mark: You are correct, that solved our problem, thank you for the advice. I'll close this PR

PR #1058, opened 2023-02-15
[izumi-veizi] add strategy for calculating votes according to amount of veiZi

LpcAries:
Fixes # .

- add new strategy for calculating votes according to amount of veiZi
- 
-


src/strategies/izumi-veizi/index.ts
```diff
@@ -0,0 +1,79 @@
+import { BigNumberish } from '@ethersproject/bignumber';
+import { formatUnits } from '@ethersproject/units';
+import { Multicaller } from '../../utils';
+import abi from './abi.json'
```
ChaituVR, line 4: It is not required to pass the entire json, just pass the abi of the method you are using :)

src/strategies/izumi-veizi/index.ts
```diff
@@ -0,0 +1,79 @@
...
+
+export const author = 'izumi';
```
ChaituVR, line 6:
```suggestion
export const author = 'izumiFinance';
```

src/strategies/izumi-veizi/index.ts
```diff
@@ -0,0 +1,79 @@
...
+                const decimalPoint = Number(Math.floor(parseFloat(formatUnits(point, options.decimals)) * 100) / 100)
+                const tmp = Object.keys(ans).find((t)=>{return t===address})
+                if (tmp !== undefined) {
+                    ans[address] = ans[address] + decimalPoint
+                } else {
+                    Object.assign(ans, {[address]:decimalPoint})
+                }
+          })   
+        }   
+    }))
```
ChaituVR, line 51: It should not call the contract once per address. https://github.com/snapshot-labs/snapshot-strategies#checklist-for-adding-a-new-strategy

ChaituVR (changes requested):

src/strategies/izumi-veizi/index.ts
LpcAries: Code has been updated, waiting for your reply

PR #1059, opened 2023-02-15
added strategy to call dcb tiers contract and get totaldeposited toke…

mj-dcb:
-  Include contract call strategy to get total deposited tokens from all staking pools

### DIFF
--- a/src/strategies/dcb-call-tiers/index.ts
+++ b/src/strategies/dcb-call-tiers/index.ts
@@ -1,0 +1,45 @@
+import { formatUnits } from '@ethersproject/units';
+import { multicall } from '../../utils';
+
+export const author = 'jaymaree';
+export const version = '0.1.0';
+
+function getArgs(options, address: string) {
+  const args: Array<string | number> = options.args || ['%{address}'];
+  return args.map((arg) =>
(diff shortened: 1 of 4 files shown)


mj-dcb:
@ChaituVR let me know if you need anything.

src/strategies/dcb-call-tiers/index.ts
```diff
@@ -0,0 +1,45 @@
...
+    response.map((value, i) => [
+      addresses[i],
+      parseFloat(
+        formatUnits(
+          options?.output ? value[options.output].toString() : value.toString(),
+          options.decimals
+        )
+      )
+    ])
+  );
```
ChaituVR, line 44: Hi @mj-dcb I think you can use contract-call strategy instead, can try? It is not required to create a new strategy if it works :)

ChaituVR (changes requested):

src/strategies/dcb-call-tiers/index.ts
mj-dcb: @ChaituVR, you're right, we could use contract-call right now, but we will change the strategy in the short future therefore, we wanted to add our base strategy with a working example.json file.

PR #1060, opened 2023-02-15
Add strategy support for STFX vesting tokens

JakeXBT:
Adds an addition strategy to `strategies` which supports STFX vesting tokens.

- Adds an addition strategy to `strategies` which supports STFX vesting tokens.


JakeXBT:
Closing - The strategy does not pass all unit tests.

PR #1077, opened 2023-02-28
[moonbeam-free-balance] Format moonbeam strategy to include decimals

crystalin:
Moonbeam free balance strategy is returning raw instead of formatted tokens value.
This PR fixes it

### DIFF
--- a/src/strategies/moonbeam-free-balance/examples.json
+++ b/src/strategies/moonbeam-free-balance/examples.json
@@ -3,7 +3,9 @@
     "name": "moonbeam-free-balance",
     "strategy": {
       "name": "moonbeam-free-balance",
-      "params": {}
+      "params": {
+        "decimals": 18
+      }
     },
     "network": "1284",
     "addresses": [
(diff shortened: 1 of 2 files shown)


crystalin:
@ChaituVR I realize we forgot the returned value needed to be formatted with decimals.
Can you take a look at it please ?

ChaituVR (approved):

crystalin:
Thank you :)

ChaituVR:
Thanks for your patience @crystalin 🙏 😃

ChaituVR:
HI @crystalin You are not using `snapshot` block here? it is causing confusion to users trying to vote. https://github.com/snapshot-labs/snapshot-strategies/blob/master/src/strategies/moonbeam-free-balance/index.ts

crystalin:
Yes that is right, I'll fix quickly.
Additionally, Moonbeam supports proxies and delegations. However those are in a "many-to-many" relationship. Ex:
> Account A gives proxy rights to Account B, C, D to perform "vote"

I didn't find a good way to support that on snapshot (without allowing voting multiple times with the same balance)

@ChaituVR do you have an idea ?

crystalin:
> HI @crystalin You are not using `snapshot` block here? it is causing confusion to users trying to vote. https://github.com/snapshot-labs/snapshot-strategies/blob/master/src/strategies/moonbeam-free-balance/index.ts

Should be fixed by https://github.com/snapshot-labs/snapshot-strategies/pull/1225

PR #1078, opened 2023-03-01
Eco voting power [eco-voting-power] - Use integers for Voting Power 

carlosfebres:
## Description
Use integers to avoid precision issues


ChaituVR:
I noticed there are two active proposals in ecollecitve space. if we change the strategies, it may affect the scores, better to wait until the vote is closed or archived

ChaituVR:
Also the tests are failing can have a look? try running `yarn test --starategy=eco-voting-power` in your local

carlosfebres:
@ChaituVR we made some changes to our subgraph. I updated the `examples.json` file with the new addresses. It should be good now.

ChaituVR (approved):

PR #1079, opened 2023-03-01
[delegation-with-overrides] feat: Add delegation-with-overrides

0xButterfield:
- Adds a new strategy, `delegation-with-overrides`. The use-case for this strategy is to forward delegated votes to an address defined in the strategy, via an overrides mapping param.

### DIFF
--- a/src/strategies/delegation-with-overrides/index.ts
+++ b/src/strategies/delegation-with-overrides/index.ts
@@ -4,6 +4,7 @@ import { getAddress } from '@ethersproject/address';
 
 export const author = '0xbutterfield';
 export const version = '0.1.0';
+export const dependOnOtherAddress = true;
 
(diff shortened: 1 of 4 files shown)


src/strategies/delegation-with-overrides/index.ts
```diff
@@ -0,0 +1,69 @@
+import { getDelegations } from '../../utils/delegation';
+import { getScoresDirect } from '../../utils';
+import { getAddress } from '@ethersproject/address';
+
+export const author = '0xbutterfield';
+export const version = '0.1.0';
+
```
ChaituVR, line 7:
```suggestion
export const dependOnOtherAddress = true;

```

ChaituVR (approved):

PR #1082, opened 2023-03-03
[GenomesDAO] add vegnome balance on arbitrum

Charles-repo:
Fixes # .

- 
- 
-


src/strategies/genomesdao/index.ts
```diff
@@ -22,6 +23,18 @@ export async function strategy(
 ): Promise<Record<string, number>> {
   const blockTag = typeof snapshot === 'number' ? snapshot : 'latest';
 
+  const viewWeb3onArb = snapshotMain.utils.getProvider('42161');
+
+  const multi4arbi = new Multicaller('42161', viewWeb3onArb, abi, {
+    blockTag: 65834212
+  });
```
ChaituVR, line 30: Hi @Charles-repo here we should use block number and network so it can be used for other networks and different block numbers

ChaituVR (changes requested):

src/strategies/genomesdao/index.ts
Charles-repo: @ChaituVR I'd like to add the balance on arbitrum to the voting power.
That's why I've used fixed parameters.

Charles-repo:
I'll try to add strategy into same space.

PR #1098, opened 2023-03-14
Adding new strategy lrc-nft-search-mult[lrc-nft-search-mult]

Raecaug:
Expanded version of the lrc-nft-dao-search voting strategy. It is meant to be used in conjunction with that strategy, to allow specific weighting of votes on certain NFTs by allowing a multiplier to be supplied, which will then be applied across all returned NFT amounts.

### DIFF
--- a/src/strategies/index.ts
+++ b/src/strategies/index.ts
@@ -365,6 +365,7 @@ import * as lrcL2SubgraphBalanceOf from './lrc-l2-subgraph-balance-of';
 import * as lrcL2NftBalanceOf from './lrc-l2-nft-balance-of';
 import * as lrcLPSubgraphBalanceOf from './lrc-lp-subgraph-balance-of';
 import * as lrcNFTDAOSearch from './lrc-nft-dao-search';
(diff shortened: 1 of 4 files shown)


ChaituVR (changes requested):

ChaituVR:
@Raecaug I see test cases are failing, make sure to test this strategy in your local, Refer https://github.com/snapshot-labs/snapshot-strategies#test-strategy-with-predefined-tests

Raecaug:
Have updated example to satisfy failing cases. One case continues to fail however, not due to any fault in the code. The AWS outage which was resolved a few hours ago appears to have affected the Loopring indexer. As such the test for the latest block fails:
image

Test cases on historical blocks, previously indexed before the outage do however report as expected: 
image

ChaituVR (approved):

ChaituVR:
Where is this subgraph located? can't find it here https://thegraph.com/hosted-service/subgraph/juanmardefago/loopring ?

PR #1099, opened 2023-03-15
brave arena strategy [bravearena]

BraveJosh:
Fixes # .

- add brave arena strategy

### DIFF
--- a/src/strategies/bravearena/index.ts
+++ b/src/strategies/bravearena/index.ts
@@ -1,0 +1,58 @@
+import fetch from 'cross-fetch';
+// const { getAddress } = require('@ethersproject/address');
+
+export const author = 'bravearena';
+export const version = '0.1.0';
+
+const API_URLS = {
+  42161: 'https://api.bravearena.io/staking/power',
+  421613: 'https://dev.bravearena.io/staking/power',
+  5: 'https://dev.bravearena.io/staking/power'
+};
+
+type VPListRequest = {
+  addresses: string[];
+  blockTime: number;
+  network: number
+};
+
+async function getVPList({
+  addresses,
+  network,
+  blockTime
+}: VPListRequest): Promise<number[]> {
+  const res = await fetch(API_URLS[network], {
+    method: 'POST',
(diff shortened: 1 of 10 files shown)


src/strategies/bravearena/index.ts
```diff
@@ -0,0 +1,58 @@
+import fetch from 'cross-fetch';
+// const { getAddress } = require('@ethersproject/address');
+
+export const author = 'bravearena';
```
ChaituVR, line 4: PR should be coming from this github's fork

src/strategies/bravearena/index.ts
```diff
@@ -0,0 +1,58 @@
...
+  const vpList = await getVPList({
+    addresses,
+    blockTime,
+    network: chainId?? network
+  });
+  const result = {};
+  for (let index = 0; index < addresses.length; index ++) {
+    result[addresses[index]] = vpList[index];
+  }
+  return result;
```
ChaituVR, line 57: Looks like this is same as `api-post` or `api-v2` strategy, Have you tried those? you can use them directly on your space :)

ChaituVR (changes requested):

src/strategies/bravearena/index.ts
BraveJosh: yes

src/strategies/bravearena/index.ts
BraveJosh: api-post doesn't have the option to get the block timestamp from the chain.

ChaituVR: Hey @BraveJosh Both api-post and api-v2 will send `snapshot` block in the payload, there is no way to accept this in your API?

If not, make sure to use `snapshot` instead of `latest` in 
```
const block = await provider.getBlock('latest');
```
else the voting power will be unpredictable

BraveJosh: Our voting power is based on the block timestamp like starsharks. So it doesn't depend on the block number.

BraveJosh: We decided to use the api-post strategy. So we will close this pull request.

PR #1101, opened 2023-03-17
[math] feat: add minus operation

anlia260:
# strategy.math support (minus)

-  update example.json address formatter to pass test
-  add Example simple contract call  minus


src/strategies/math/README.md
```diff
@@ -17,6 +17,8 @@ Currently supported operations are:
 | `a-if-lte-b`  | 3             | (x, a, b) = x <= b ? a : x                 |
 | `a-if-gt-b`   | 3             | (x, a, b) = x > b ? a : x                  |
 | `a-if-gte-b`  | 3             | (x, a, b) = x >= b ? a : x                 |
+| `sum`         | 2             | x + a                                      |
```
ChaituVR, line 20: Hi @anlia260 But using two strategies in a space will already return a sum, is there a reason why we need this?

ChaituVR (changes requested):

src/strategies/math/README.md
anlia260: sure, there is some reasons
1. using two strategies in a space will already return a sum is not too obvious ( detailed descriptions cannot be found in the documentation)
2. multiple strategies is less than eight, the math.sum is reduce sum that means multiple strategies expand in the feature ( but now just two)

anlia260: Got you worried, I will remove the sum now, waiting for me.

src/strategies/math/examples.json
```diff
@@ -1,4 +1,106 @@
...
+                  "constant": true,
+                  "stateMutability": "view"
+                }
+              }
+            }
+          }
+        ],
+        "operation": "minus"
+      }
+    },
```
ChaituVR, line 95: Can add smaller examples plz 😄 maybe like ticket strategy or something, Everything else looks fine, thanks for the updates

ChaituVR (changes requested):

ChaituVR:
Thanks for understanding and modifying it accordingly 🙏 :)

PR #1104, opened 2023-03-20
[galxe-loyalty-points] add galxe-loyalty-points strategy

HaynarCool:
add galxe-loyalty-points strategy


zzuziak:
Hey @HaynarCool 👋  Could you please provide a Readme for spaces (having non-dev users in mind) to explain what the strategy does and how to set it up?

src/strategies/galxe-loyalty-points/examples.json
```diff
@@ -0,0 +1,23 @@
+[
+  {
+    "name": "Galxe users loyalty points",
+    "strategy": {
+      "name": "galxe-loyalty-points",
+      "params": {
+        "symbol": "PTS",
+        "spaceId": 263,
```
ChaituVR, line 8:
```suggestion
        "spaceId": 263
```
Tests are failing because of this

src/strategies/galxe-loyalty-points/index.ts
```diff
@@ -0,0 +1,49 @@
...
+
+const graphqlUrl = 'https://graphigo.stg.galaxy.eco/query';
+
+export async function strategy(
+  space,
+  network,
+  provider,
+  addresses,
+  options,
+  snapshot
```
ChaituVR, line 14: There is no way to use the `snapshot` block in the graphql? if it is not used, results will be unpredictable

ChaituVR (changes requested):

src/strategies/galxe-loyalty-points/index.ts
HaynarCool: it now supports `snapshot`

ChaituVR (approved):

PR #1106, opened 2023-03-22
add erc115 weighted by token id [erc1155-weighted-by-id]

isaac-martin:
Fixes # .

- add new weighting strategy
-


ChaituVR (changes requested):
Hey @isaac-martin It is same as erc1155-balance-of-ids-weighted strategy no?

isaac-martin:
Thanks for the review @ChaituVR the strategy you mentioned multiplies all tokens by the same weight, whereas this new strategy allows you define a different weight to each token id. 

In the first example i provided i assign a weight of 25 to ID Number 1, a weight of 10 to id number 2 and a weight of 1 to id number 3.

If we passed in two wallets when calling this function
- wallet a (holds 1 x token id 1)
- wallet b (holds 1 x token id 2)
the result would be 25 votes for wallet a and 10 votes for wallet b

src/strategies/index.ts
```diff
@@ -735,6 +737,7 @@ const strategies = {
   'staked-uniswap-modifiable': stakedUniswapModifiable,
   'erc1155-balance-of-ids': erc1155BalanceOfIds,
   'erc1155-balance-of-ids-weighted': erc1155BalanceOfIdsWeighted,
+  'erc-1155-weighted-by-id': erc1155weighted,
```
ChaituVR, line 740: Thank you so much for explaining, can rename it to `erc1155-weighted-by-id` instead of `erc-1155-weighted-by-id` please :) that way it would be similar to other strategies names, everything else looks good 👍

ChaituVR (changes requested):

src/strategies/index.ts
isaac-martin: thanks @ChaituVR - updated

ChaituVR (approved):

ChaituVR:
Thanks for the changes, but tests are failing @isaac-martin Can try in your local? i think this method is hitting the gas cap when we pass 500 addresses?

isaac-martin:
> Thanks for the changes, but tests are failing @isaac-martin Can try in your local? i think this method is hitting the gas cap when we pass 500 addresses?

yeah it works for up to about 450 addresses, looking into it - the `balanceOfBatch` shouldn't be that gas intensive so not sure whats going on. Do you have any ideas?

PR #1117, opened 2023-03-28
Otterspace - new networks [otterspace-badges]

rahulrumalla:
1. Adds 2 new neworks - Mainnet & Optimism goerli
2. Udpates logic for using the badge with the highest weight when an address holds multiple badges
3. Updates exmaples' address to using checksums

### DIFF
--- a/src/strategies/otterspace-badges/index.ts
+++ b/src/strategies/otterspace-badges/index.ts
@@ -7,7 +7,7 @@ export const version = '1.0.1';
 const OTTERSPACE_SUBGRAPH_API_URLS_BY_CHAIN_ID = {
   '1': 'https://api.thegraph.com/subgraphs/name/otterspace-xyz/badges-mainnet',
   '5': 'https://api.thegraph.com/subgraphs/name/otterspace-xyz/badges-goerli',
-  '10': 'https://api.thegraph.com/subgraphs/name/otterspace-xyz/badges-optimism',
+  '10': 'https://api.thegraph.com/subgraphs/name/otterspace-xyz/badges-optimism-alpha',
   '420':
     'https://api.thegraph.com/subgraphs/name/otterspace-xyz/badges-optimism-goerli'
 };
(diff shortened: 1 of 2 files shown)


src/strategies/otterspace-badges/index.ts
```diff
@@ -5,8 +5,11 @@ export const author = 'otterspace-xyz';
 export const version = '1.0.0';
```
ChaituVR, line 5:
```suggestion
export const version = '1.0.1';
```

src/strategies/otterspace-badges/index.ts
```diff
@@ -75,9 +79,17 @@ function applyBadgeWeights(badges: [], options: any) {
...
-    badgeWeights[badgeAddress] = getBadgeWeight(options.specs, badge.spec.id);
+    // picks the highest weight when multiple badges are held by the same address
+    if (
+      !badgeWeights[badgeAddress] ||
+      badgeWeight > badgeWeights[badgeAddress]
+    ) {
+      badgeWeights[badgeAddress] = badgeWeight;
+    } else {
+      return;
+    }
```
ChaituVR, line 92: Hi @rahulrumalla There is a active proposal on cowgrants.eth space, that is using this space, can this change create a problem in someone's voting power? better to wait until the proposal is closed

ChaituVR (changes requested):

src/strategies/otterspace-badges/index.ts
ChaituVR: oh i think it may not cause a problem looking at the code, can confirm? :)

rahulrumalla: It shouldn't cause a problem

ChaituVR (approved):

rahulrumalla:
> @rahulrumalla Test cases are failing, can take a look? :)

Done!

PR #1121, opened 2023-04-05
Add RichQuack Staking amount strategy [richquack-staked-amount]

ephdtrg:
- Create new strategy [richquack-staked-amount] to calculate votes from staking amounts of RichQuack Staking

### DIFF
--- a/src/strategies/richquack-staked-amount/index.ts
+++ b/src/strategies/richquack-staked-amount/index.ts
@@ -1,0 +1,57 @@
+import { BigNumberish } from '@ethersproject/bignumber';
+import { formatUnits } from '@ethersproject/units';
+import { Multicaller } from '../../utils';
+
+export const author = 'ephdtrg';
+export const version = '0.1.0';
+
+const abi = [
+  {
+    inputs: [
+      {
+        internalType: 'address',
+        name: '',
+        type: 'address'
+      }
+    ],
+    name: 'stakeInfo',
+    outputs: [
+      {
+        internalType: 'uint256',
+        name: 'level',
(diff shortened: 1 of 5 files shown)


src/strategies/richquack-staked-amount/index.ts
```diff
@@ -0,0 +1,57 @@
...
+  addresses.forEach((address) =>
+    multi.call(address, options.stakingAddress, 'stakeInfo', [address])
+  );
+  const result: Record<string, BigNumberish> = await multi.execute();
+
+  return Object.fromEntries(
+    Object.entries(result).map(([address]) => [
+      address,
+      parseFloat(formatUnits(result[address][0], 18))
+    ])
```
ChaituVR, line 55: Hi @ephdtrg i think contract-call strategy should work for you, can try?

ChaituVR (changes requested):

src/strategies/richquack-staked-amount/index.ts
ephdtrg: Hi, @ChaituVR ! Thank you for response

Do you mean to use existing contract-call strategy for this workflow, instead of writing new strategy, correct?

ChaituVR: Yep @ephdtrg Can use it in your space directly :)

ephdtrg: Thank you for suggestion @ChaituVR !
Was not aware of this strategy, will check it out and close this PR if it works as intended  :)

PR #1123, opened 2023-04-07
[cap-voting-power] Added cap-voting-power strategy

unRealGamer28:
Fixes # .

- Adds a new strategy called 'cap-voting-power' which calculates the off-chain voting power for addresses with tokens vested in a contract.

### DIFF
--- a/src/strategies/cap-voting-power/index.ts
+++ b/src/strategies/cap-voting-power/index.ts
@@ -9,6 +9,7 @@ const abi = ['function totalSupply(uint256 t) external view returns (uint256)'];
 
 const maxLockupDuration = 4 * 365 * 24 * 60 * 60; // Maximum lockup duration is 4 years in seconds based on Curve VE contract
 const maxVestingDuration = 4 * 365 * 24 * 60 * 60; // Maximum vesting duration is 4 years in seconds
+const maxVestingAddressCount = 500; // Maximum number of vesting addresses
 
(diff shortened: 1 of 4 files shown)


src/strategies/cap-voting-power/index.ts
```diff
@@ -0,0 +1,267 @@
...
+export async function strategy(
+  space,
+  network,
+  provider,
+  addresses,
+  options,
+  snapshot
+): Promise<Record<string, number>> {
+  // Assertions/checks to validate options
+  options.vestingAddresses.forEach((vestingAddress) => {
```
ChaituVR, line 22: Hey @unRealGamer28 can you add a limit to `vestingAddresses` so it doesn't use all memory 😃

ChaituVR, line 22: Hey @unRealGamer28 can you add a limit to `vestingAddresses` so it doesn't use all memory 😃

ChaituVR (changes requested):

src/strategies/cap-voting-power/index.ts
unRealGamer28: Hey @ChaituVR  I added a limit in code and set it to 500 vesting addresses. I believe this is what you were looking for. But please let me know if you want another change! Thank you so much

ChaituVR (approved):

PR #1126, opened 2023-04-10
Add babywealthyclub-multi strategy

claibornej:
Fixes # .
add babywealthyclub-multi strategy

image

### DIFF
--- a/src/strategies/babywealthyclub-multi/index.ts
+++ b/src/strategies/babywealthyclub-multi/index.ts
@@ -1,0 +1,56 @@
+import { getProvider, getSnapshots } from '../../utils';
+import strategies from '..';
+
+export const author = 'claibornej';
+export const version = '0.1.0';
+
+export async function strategy(
+  space,
+  network,
+  provider,
+  addresses,
+  options,
+  snapshot
+) {
+  const promises: any = [];
+  const blocks = await getSnapshots(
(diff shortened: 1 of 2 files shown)


src/strategies/babywealthyclub-multi/index.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  const results = await Promise.all(promises);
+  const resp = results.reduce((finalResults: any, strategyResult: any) => {
+    for (const [address, value] of Object.entries(strategyResult)) {
+      if (!finalResults[address]) {
+        finalResults[address] = 0;
+      }
+      finalResults[address] += value;
+    }
+    return finalResults;
```
ChaituVR, line 52: Hi @claibornej I think you can use multiple strategies in your space than a separate strategy, the result of multiple strategies will be sum of them

ChaituVR (changes requested):

src/strategies/babywealthyclub-multi/index.ts
claibornej: I need to query the nft quantity of multiple chains

ChaituVR: @claibornej You can do same from your space settings, you can select different network for different strategy

PR #1128, opened 2023-04-10
[degenzoo-erc721-all-tokens] Strategy #2 : All tokens (holding/staking/animals)

aorfevre:
This strategy is tracking both chains assets of each wallets.
It is based on our own backend that will fetch all related assets score value. 

Assets tracked : 
Eth chain : 
* Holding amount
* Staked Amount

BNB Chain : 
* Holding amount
* Staked Amount
* Animals holding Staked tokens

### DIFF
--- a/src/strategies/degenzoo-erc721-all-tokens/index.ts
+++ b/src/strategies/degenzoo-erc721-all-tokens/index.ts
@@ -2,7 +2,7 @@ import fetch from 'cross-fetch';
 import { getAddress } from '@ethersproject/address';
 
 export const author = 'aorfevre';
-export const version = '0.1';
+export const version = '0.1.0';
 
 
 export async function strategy(
--- a/src/strategies/degenzoo-erc721-all-tokens/index.ts
+++ b/src/strategies/degenzoo-erc721-all-tokens/index.ts
@@ -1,0 +1,50 @@
+import fetch from 'cross-fetch';
+import { getAddress } from '@ethersproject/address';
+
+export const author = 'aorfevre';
+export const version = '0.1';
+
+
+export async function strategy(
+  space,
+  network,
+  provider,
+  addresses,
+  options,
+  snapshot
+): Promise<Record<string, number>> {
+
+  // initialize scores
+  const scores = {};
+  for (const address of addresses) {
+    scores[getAddress(address)] = 0;
+  }
+
+
+  async function getWalletScore(wallet) {
+    const url = `https://api.daomaker.com/get-dzoo-assets-by-wallet?wallet=${wallet}`;
+
+    const res = await fetch(url, {
+      method: 'GET',
+      headers: {
+        Accept: 'application/json',
+        'Content-Type': 'application/json'
+      }
+    });
+
+    const data = await res.json();
+    return data.total;
+  };
+  
+  const promises: any = [];
+  addresses.forEach(function (address) {
+    return promises.push(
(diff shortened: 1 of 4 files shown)


aorfevre:
hello,
Is everything ok ?

src/strategies/degenzoo-erc721-all-tokens/index.ts
```diff
@@ -0,0 +1,50 @@
+import fetch from 'cross-fetch';
+import { getAddress } from '@ethersproject/address';
+
+export const author = 'aorfevre';
+export const version = '0.1';
```
ChaituVR, line 5:
```suggestion
export const version = '0.1.0';
```

src/strategies/degenzoo-erc721-all-tokens/index.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  // initialize scores
+  const scores = {};
+  for (const address of addresses) {
+    scores[getAddress(address)] = 0;
+  }
+
+
+  async function getWalletScore(wallet) {
+    const url = `https://api.daomaker.com/get-dzoo-assets-by-wallet?wallet=${wallet}`;
```
ChaituVR, line 25: Hi @aorfevre This is not your API? maybe you could use `api` or `api-v2` strategy instead?

ChaituVR (changes requested):

src/strategies/degenzoo-erc721-all-tokens/index.ts
aorfevre: This is my api.

aorfevre: However, I did not use your api-v2 because it is targeting a single network.

DZOO is a cross chain project and the Voting power shall reflect that. 
This api : https://api.daomaker.com/get-dzoo-assets-by-wallet?wallet=${wallet} is checking assets of the user on all chains.

Do you still want me to use one of your other templates ?

aorfevre:
@ChaituVR 
See my comment here : https://github.com/snapshot-labs/snapshot-strategies/pull/1128#discussion_r1167624393

src/strategies/index.ts
```diff
@@ -434,7 +434,8 @@ import * as pdnBalancesAndVests from './pdn-balances-and-vests';
 import * as izumiVeiZi from './izumi-veizi';
 import * as lqtyProxyStakers from './lqty-proxy-stakers';
 import * as echelonWalletPrimeAndCachedKeyGated from './echelon-wallet-prime-and-cached-key-gated';
-import * as rdntCapitalVoting from './rdnt-capital-voting';
+import * as rdntCapitalVoting from './rdnt-capital-voting;
```
ChaituVR, line 437:
```suggestion
import * as rdntCapitalVoting from './rdnt-capital-voting';
```

src/strategies/degenzoo-erc721-all-tokens/index.ts
ChaituVR: So right now, it is sending one request per address won't scale, can send all in one request? so instead can use `api-v2` maybe just `network` param in your code? which should handle all networks right?

Also make sure your API consider `snapshot` block number to return voting power, else voting power can be unpredictable and won't be able to replay the voting power in future

aorfevre:
Definitly using the api-v2
I am closing that PR now.

PR #1130, opened 2023-04-12
Add offchain-delegation-2 Strategy for pre-distribution tokens

alexotsu:
This Strategy pulls token allocation information from a CSV. Necessary for temp checks that happen before distribution events.

- Add `/offchain-delegation-2` containing example.json, index.ts, and README.md
- Update index.ts to include `offchain-delegation-2` as `offchainDelegation2`

### DIFF
--- a/src/strategies/index.ts
+++ b/src/strategies/index.ts
@@ -408,6 +408,7 @@ import * as clqdrBalanceWithLp from './clqdr-balance-with-lp';
 import * as ninechroniclesStakedAndDcc from './ninechronicles-staked-and-dcc';
 import * as spreadsheet from './spreadsheet';
 import * as offchainDelegation from './offchain-delegation';
+import * as offchainDelegation2 from './offchain-delegation-2';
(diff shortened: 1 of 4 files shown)


ChaituVR (changes requested):
Hi @alexotsu I don't understand the changes, it is same as offchain-delegation strategy but you are not reading scores of delegations?

alexotsu:
> Hi @alexotsu I don't understand the changes, it is same as offchain-delegation strategy but you are not reading scores of delegations?

@ChaituVR yes - it is for tokens that haven’t been distributed to individual wallets yet for various reasons, but where projects still want to hold a vote. So the voting power itself is written down off-chain on the spreadsheet.

ChaituVR:
Thank you! Can rename it to something else? :) because it may confuse many others with delegation strategy

alexotsu:
Actually, I just realized that this is identical to the “[spreadsheet](https://github.com/snapshot-labs/snapshot-strategies/tree/master/src/strategies/spreadsheet)” strategy, so will plan to close unless you think there is some big reason to keep both

PR #1135, opened 2023-04-19
Multinetwork Support for Math [math]

RatanRSur:
I wanted to be able to be able to multiple a token balance on mainnet with one on polygon but that wasn't possible even when `network` was explicitly provided to the `operand`s.

### DIFF
--- a/src/strategies/math/index.ts
+++ b/src/strategies/math/index.ts
@@ -29,7 +29,7 @@ export async function strategy(
   // Recursively resolve operands
   const operandPromises: Promise<Record<string, number>>[] =
     strategyOptions.operands.map((item) =>
-      resolveOperand(item, addresses, space, network, provider, snapshot)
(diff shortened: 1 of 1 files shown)


src/strategies/math/index.ts
```diff
@@ -170,7 +169,7 @@ async function resolveOperand(
         strategyOperand.strategy.name
       ].strategy(
         space,
-        network,
+        strategyOperand.strategy.network,
```
ChaituVR, line 172: Hi @RatanRSur Can make it optional? like 
```js
strategyOperand.strategy.network || network
```

ChaituVR (changes requested):

src/strategies/math/index.ts
RatanRSur: Good idea. I hope it's ok that I used `??` instead to be more precise

ChaituVR (approved):

PR #1143, opened 2023-04-26
[erc1155-all-balances-of-with-erc721-key] Add strategy to count erc1155 votes only for holders of erc721 keys

darrylhansen:
Fixes # .

- Create new strategy [erc1155-all-balances-of-with-erc721-key] to calculate votes for holders of specified erc1155 tokens, but only count them if they have a required erc721 key.

### DIFF
--- a/src/strategies/erc1155-all-balances-of-with-erc721-key/index.ts
+++ b/src/strategies/erc1155-all-balances-of-with-erc721-key/index.ts
@@ -1,0 +1,92 @@
+import { multicall, getProvider } from '../../utils';
+import { strategy as erc721Strategy } from '../erc721';
+import { getSnapshots } from '../../utils';
+
+export const author = 'darrylhansen';
+export const version = '0.1.0';
+
+const erc1155Abi = [
+  'function balanceOf(address account, uint256 id) view returns (uint256)'
+];
+
+async function getERC1155Balances(network, provider, addresses, options, snapshot) {
+  const blockTag = typeof snapshot === 'number' ? snapshot : 'latest';
+  const response = await multicall(
+    network,
+    provider,
+    erc1155Abi,
+    addresses.flatMap((address: any) =>
+      options.tokenIds.map((tokenId: any) => [
+        options.address,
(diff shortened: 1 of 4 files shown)


src/strategies/erc1155-all-balances-of-with-erc721-key/index.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  return results.reduce((finalResults: any, networkResult: any) => {
+    for (const [address, value] of Object.entries(networkResult)) {
+      if (!finalResults[address]) {
+        finalResults[address] = 0;
+      }
+      finalResults[address] += value;
+    }
+    return finalResults;
+  }, {});
```
ChaituVR, line 91: Hey @darrylhansen You can use two different strategies in your space :) The result will be the sum of those two strategies, with no need of a new strategy

ChaituVR (changes requested):

src/strategies/erc1155-all-balances-of-with-erc721-key/index.ts
darrylhansen: Hi @ChaituVR, I couldn't figure out how to make two strategies do what we want because they are cumulative.

We want the ERC1155's to count as votes, but only if someone holds one of the ERC721's. Any strategies I found there was no way to exclude someone who held the ERC1155's but no ERC721.

ChaituVR: AH! Got it, then you can use voting validations :) 

https://docs.snapshot.org/user-guides/strategies/what-is-a-strategy-1
Go to space settings => voting => Validation
there you should be able to select `Basic` Validation, in which you can select `erc721` - This will be used to check if someone can vote or not

 and 

In strategies you can select erc1155 strategy -> This will be used for voting power

darrylhansen: Ah, slick 🤦. Okay, thanks very much!

PR #1146, opened 2023-04-28
[battlefly-vgfly-and-staked-gfly]Feature/battlefly dao voting upgrade for founders voting

Archethect:
Upgrade for the battlefly strategy that allows voting with Battlefly founders tokens.


src/strategies/battlefly-vgfly-and-staked-gfly/index.ts
ChaituVR, line 1: Also update the version plz :)
```
```

ChaituVR (changes requested):

src/strategies/battlefly-vgfly-and-staked-gfly/index.ts
Archethect: done!

ChaituVR (approved):

PR #1155, opened 2023-05-10
[hedgey-delegate] adds strategy for hedgey vesting contracts

bark4mark:
Adds hedgey-delegate

This is a simple strategy to call delegatedBalances and apply a multiplier value.

### DIFF
--- a/src/strategies/hedgey-delegate/index.ts
+++ b/src/strategies/hedgey-delegate/index.ts
@@ -1,0 +1,60 @@
+import { BigNumberish } from '@ethersproject/bignumber';
+import { formatUnits } from '@ethersproject/units';
+import { Multicaller } from '../../utils';
+
+export const author = 'hedgeyfinance';
+export const version = '1.0.0';
+
+const abi = [
+  {
+    inputs: [
+      {
+        internalType: 'address',
+        name: 'delegate',
+        type: 'address'
+      },
+      {
+        internalType: 'address',
+        name: 'token',
+        type: 'address'
+      }
+    ],
+    name: 'delegatedBalances',
+    outputs: [
+      {
+        internalType: 'uint256',
+        name: 'delegatedBalance',
+        type: 'uint256'
+      }
+    ],
+    stateMutability: 'view',
+    type: 'function'
+  }
+];
+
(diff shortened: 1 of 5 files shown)


src/strategies/hedgey-delegate/index.ts
```diff
@@ -0,0 +1,60 @@
...
+    ])
+  );
+  const result: Record<string, BigNumberish> = await multi.execute();
+
+  return Object.fromEntries(
+    Object.entries(result).map(([address, balance]) => [
+      address,
+      parseFloat(formatUnits(balance, options.decimals)) * options.multiplier
+    ])
+  );
```
ChaituVR, line 59: hi @bark4mark I think you can use the `math` strategy with `contract-call` in it :) Can try?

bark4mark: Would it be possible to have this strategy merged? I would like to keep the configuration minimal as I have to create documentation around voting and configuring the smart contracts.

src/strategies/hedgey-delegate/index.ts
```diff
@@ -0,0 +1,60 @@
+import { BigNumberish } from '@ethersproject/bignumber';
+import { formatUnits } from '@ethersproject/units';
+import { Multicaller } from '../../utils';
+
+export const author = 'hedgeyfinance';
```
ChaituVR, line 5:
```suggestion
export const author = 'hedgey-finance';
```

src/strategies/hedgey-delegate/index.ts
```diff
@@ -0,0 +1,60 @@
...
+      {
+        internalType: 'uint256',
+        name: 'delegatedBalance',
+        type: 'uint256'
+      }
+    ],
+    stateMutability: 'view',
+    type: 'function'
+  }
+];
```
ChaituVR, line 33:
```suggestion
const abi = [
  'function delegatedBalances(address delegate, address token) view returns (uint256 delegatedBalance)'
]
```

ChaituVR (changes requested):

ChaituVR (approved):

PR #1157, opened 2023-05-10
[staked-defi-balance] Fixed params structure to make it less confusing

taha-abbasi:
Release notes for staked-defi-balance 1.3.0:
-------------

1. updated `params` structure to have one `minStakedBalance` value across multiple contracts
2. Added logic to calculate and store balance in the loop for each contract then compare against `minStakedBalance` once looping is done and both values are already human-readable
3. `minStakedBalance` is not human-readable value
4. `./src/strategies/staked-defi-balance/schema.json` has also been updated

@ChaituVR I thought about these issues after you pointed out the problem, so the changes are raised here.


ChaituVR (approved):

critpto:
I will give it a go!

PR #1158, opened 2023-05-12
[passport-gated] Passport migration

sophiahdew:
- Implement the passport-gated and passport-weighted validation strategies to integrate with the Passport Scorer API instead of the SDK
- Include an input threshold minScore for passport-weighted


ChaituVR (changes requested):
Hello @sophiahdew can remove changes from .history folder? right now there are lot of changes on this PR 😄

sophiahdew:
Removed changes from .history folder

sophiahdew:
Hi @ChaituVR 

I noticed that you ran the validation tests using jest -i validation.test.ts --validation=Passport migration

I ran the tests on my end using yarn test:validation --validation=passport-gated and yarn test:validation --validation=passport-weighted

Additionally, a `.env` file is required at the project root with the following information before running this code.

- `NEXT_PUBLIC_GC_API_KEY=<your-api-key>`
- `NEXT_PUBLIC_GC_SCORER_ID=<your-scorer-id>`

Unfortunately, we don't have a test API key on our end, but you can make one by visiting https://scorer.gitcoin.co/. Here, you'll create both an API key and a Scorer. 

Additionally, here is a quick start guide that explains how to create the API key and Scorer in more detail https://docs.passport.gitcoin.co/building-with-passport/quick-start-guide

If you have any questions, please let me know

src/validations/passport-weighted/index.ts
```diff
@@ -14,35 +24,75 @@ export default class extends Validation {
...
+          console.log(`Your passport score (${roundedScore}) is lower than the threshold score (${THRESHOLD_NUMBER}).`);
+          return false;
+        } else {
+          console.log('You do not have a valid Gitcoin Passport. Create one by visiting https://passport.gitcoin.co/ ')
+          return false;
+          }
+      } catch (err) {
+        console.log('error: ', err)
+        return false;
+      }
```
ChaituVR, line 96: in case of an error, if we return false it can cause confusion, so can remove try-catch, we handle it somewhere else

ChaituVR (changes requested):

src/validations/passport-weighted/index.ts
sophiahdew: Made a new commit that changed return false to throw err so that if an error occurs, it won't be mistaken for a validation failure. This code will log the error as before, but then it will rethrow it to allow the error to propagate up the call stack to be handled by a global error handler.

src/validations/passport-gated/index.ts
```diff
@@ -13,43 +23,71 @@ export default class extends Validation {
...
+          signature,
+          nonce
+        })
+      })
+      const data = await response.json()
+      console.log('data:', data)
+    } catch (err) {
+      console.log('error: ', err)
     }
   }
```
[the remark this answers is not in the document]
sophiahdew: We updated the Passport API to no longer require signing a message on every request. I updated this in the latest commit.

package.json
```diff
@@ -41,22 +41,24 @@
     "blakejs": "^1.2.1",
     "copyfiles": "^2.4.1",
     "cross-fetch": "^3.1.5",
+    "dotenv": "^16.0.3",
     "eth-ens-namehash": "^2.0.8",
+    "ethers": "^6.3.0",
```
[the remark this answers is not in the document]
sophiahdew: Correct. Just removed it from package.jscon

sophiahdew:
Hi @ChaituVR Updated passport-gated to include a check for specific stamps. However, I'm not able to edit example.json to add a minScore parameter. It leads to an error message saying "must NOT have additional properties." Is this something you can edit on your end?

src/validations/passport-gated/README.md
```diff
@@ -1 +1,31 @@
...
+# Gitcoin Passport Gated Validation
+
+This repository provides a passport-gated validation strategy for Snapshot. The implementation integrates with the Gitcoin API to validate whether a user is authorized to vote on a proposal.
+
+## Prerequisites
+
+Before using this code, ensure that you have the following information stored in a `.env` file at the project root:
+
+- `NEXT_PUBLIC_GC_API_KEY=<your-api-key>`
+- `NEXT_PUBLIC_GC_SCORER_ID=<your-scorer-id>`
```
samuveth, line 10:
```suggestion
```

src/validations/passport-gated/README.md
```diff
@@ -1 +1,30 @@
...
+
+## Prerequisites
+
+Before using this code, ensure that you have the following information stored in a `.env` file at the project root:
+
+- `NEXT_PUBLIC_GC_API_KEY=<your-api-key>`
+
+## Overview
+
+This implementation uses the Gitcoin Passport API to check whether a user has a valid passport by looking for their passport score.
```
ChaituVR, line 13:
```suggestion
This implementation uses the Gitcoin Passport API to check whether a user has a valid passport by looking for their stamps.
```

src/validations/passport-gated/README.md
```diff
@@ -1 +1,30 @@
...
+
+## Modifications
+
+The original code utilized the Passport SDK to check if the user has a valid passport. However, with the introduction of the Passport API, we can now simplify the process by checking for a score.
+
+## Last Modified
+
+This code was last modified on May 11, 2023.
+
+Feel free to customize and extend this implementation to suit your specific needs.
```
ChaituVR, line 27:
```suggestion
## Code Explanation
[Soon] The main function in this codebase returns a binary score based on whether the user's passport is flagged as a likely Sybil.

## Modifications

The original code utilized the Passport SDK to check if the user has a valid passport. [Soon] However, with the introduction of the Passport API, we can now simplify the process by checking for a score.

## Last Modified

This code was last modified on June 1, 2023.

Feel free to customize and extend this implementation to suit your specific needs.
```

src/validations/passport-gated/index.ts
```diff
@@ -1,10 +1,16 @@
...
 import Validation from '../validation';
-import {
-  getPassport,
-  getVerifiedStamps,
-  hasValidIssuanceAndExpiration
-} from '../passport-weighted/helper';
+import dotenv from 'dotenv';
+dotenv.config({ path: '.env' });
+
+const API_KEY = process.env.NEXT_PUBLIC_GC_API_KEY
```
ChaituVR, line 6:
```suggestion
const API_KEY = process.env.PUBLIC_GC_API_KEY || '0cErnp4F.nRDEUU4Z8y5YyxcU32swrggDFNfWtXtI'
```

ChaituVR (approved):

PR #1166, opened 2023-05-23
[galxe-loyalty-points] feat: add snapshot id in params

HaynarCool:


src/strategies/galxe-loyalty-points/index.ts
```diff
@@ -38,7 +39,7 @@ export async function strategy(
       variables: {
         alias: parts[1],
         addresses: addresses,
-        snapshotId: typeof snapshot === 'number' ? snapshot : ''
+        snapshotId: options.snapshot_id
```
ChaituVR, line 42:
```suggestion
        snapshotId: options.snapshot_id ? options.snapshot_id : (typeof snapshot === 'number' ? snapshot : '')
```

src/strategies/galxe-loyalty-points/index.ts
```diff
@@ -19,6 +19,7 @@ export async function strategy(
   if (parts.length < 2) {
     throw error('invalid galxe space url');
   }
+  if (options.snapshot_id) throw error('pass valid snapshot_id');
```
ChaituVR, line 22:
```suggestion
```

PR #1171, opened 2023-05-26
[spaceid] feat: spaceid strategy

SID-Marcus:
Fixes # .

- 
- 
-

### DIFF
--- a/src/strategies/spaceid/index.ts
+++ b/src/strategies/spaceid/index.ts
@@ -1,0 +1,236 @@
+import { strategy as UniswapV3Strategy } from '../uniswap-v3';
+import { subgraphRequest } from '../../utils';
+import { BigNumber } from '@ethersproject/bignumber';
+import { WeiPerEther } from '@ethersproject/constants';
+import { getScoresDirect } from '../../utils';
+import { getAddress } from '@ethersproject/address';
+
+export const author = 'Marcus';
+export const version = '0.0.1';
+
+const SPACE_NAME = 'dao.spaceid.eth';
+const pancakeV3Subgraph =
+  'https://api.thegraph.com/subgraphs/name/messari/pancakeswap-v3-bsc';
+const snapshotSubgraph = {
+  '1': 'https://api.thegraph.com/subgraphs/name/snapshot-labs/snapshot',
+  '5': 'https://api.thegraph.com/subgraphs/name/snapshot-labs/snapshot-goerli',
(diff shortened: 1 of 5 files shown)


src/strategies/spaceid/index.ts
```diff
@@ -0,0 +1,236 @@
+import { strategy as UniswapV3Strategy } from '../uniswap-v3';
+import { subgraphRequest } from '../../utils';
+import { BigNumber } from '@ethersproject/bignumber';
+import { WeiPerEther } from '@ethersproject/constants';
+import { getScoresDirect } from '../../utils';
+import { getAddress } from '@ethersproject/address';
+
+export const author = 'Marcus';
```
ChaituVR, line 8:
```suggestion
export const author = 'SID-Marcus';
```

src/strategies/spaceid/index.ts
```diff
@@ -0,0 +1,236 @@
...
+  addresses,
+  options,
+  snapshot
+) {
+  const delegationPower = await getDelegationPower(
+    network,
+    provider,
+    addresses,
+    snapshot
+  );
```
ChaituVR, line 208: Hi @SID-Marcus We are not accepting any new strategies if voting power depend on other voters

ChaituVR (changes requested):

src/strategies/spaceid/index.ts
ChaituVR: Refer https://github.com/snapshot-labs/snapshot-strategies#code

SID-Marcus: The method is almost as same as  DELEGATION  strategy. We made some changes on getting delegation data from subgraph because the official one is too slow on querying. Delegation power is a necessary part of voting power in our space, so I would appreciate it if you could approve it.

BluesRockets: After our discussion, we decided to remove this part of this strategy. Instead, we will use API strategy to get delegation power.

SID-Marcus: After our discussion, we decided to remove this part of this strategy. Instead, we will use API strategy to get delegation power.

ChaituVR (approved):

ChaituVR:
@SID-Marcus Tests are failing, can take a look?

SID-Marcus:
@ChaituVR Could you approval for tests again

ChaituVR:
Untitled 2
@SID-Marcus

SID-Marcus:
Fixed it  @ChaituVR

PR #1190, opened 2023-06-13
[taraxa-delegation] Implements and adds Taraxa delegation strategy

VargaElod23:
Fixes # .

- Taraxa delegation strategy to calculate voting power.


src/strategies/taraxa-delegation/index.ts
```diff
@@ -0,0 +1,112 @@
...
+    abi,
+    provider
+  );
+
+  let newDelegations: ContractDelegation[] = [];
+  let page = 0;
+  let hasNextPage = true;
+  while (hasNextPage) {
+    try {
+      const allDelegations = await mainnetDpos!.getDelegations(address, page, {
```
ChaituVR, line 44: There is a lint error here

> Forbidden non-null assertion

src/strategies/taraxa-delegation/index.ts
```diff
@@ -0,0 +1,112 @@
+import { BigNumber, BigNumberish } from '@ethersproject/bignumber';
+import { Contract } from '@ethersproject/contracts';
+import { formatUnits } from '@ethersproject/units';
+
+export const author = 'taraxa';
+export const version = '0.1.0';
+export const dependOnOtherAddress = true;
```
ChaituVR, line 7: Hey @VargaElod23 Right now we are not accepting any strategies that depend on other voters addresses https://github.com/snapshot-labs/snapshot-strategies#code

ChaituVR (changes requested):

src/strategies/taraxa-delegation/index.ts
VargaElod23: Right, fixed.

src/strategies/taraxa-delegation/index.ts
```diff
@@ -0,0 +1,113 @@
+import { BigNumber, BigNumberish } from '@ethersproject/bignumber';
+import { Contract } from '@ethersproject/contracts';
+import { formatUnits } from '@ethersproject/units';
+import { getAddress } from '@ethersproject/address';
+
+export const author = 'Taraxa-project';
+export const version = '0.1.0';
+export const dependOnOtherAddress = false;
```
ChaituVR, line 8:
```suggestion
```

src/strategies/taraxa-delegation/index.ts
```diff
@@ -0,0 +1,112 @@
...
+    );
+    result[address] = delegation.stake;
+  }
+  return Object.fromEntries(
+    Object.entries(result).map(([address, stake]) => [
+      address,
+      parseFloat(formatUnits(stake, 18))
+    ])
+  );
+};
```
ChaituVR, line 93: Hi @VargaElod23 Looks like this will send one request per address? this will not scale :( https://github.com/snapshot-labs/snapshot-strategies#code,Try using muticall instead

ChaituVR (changes requested):

src/strategies/taraxa-delegation/index.ts
VargaElod23: Thing is that we cannot do that as our `getDelegations(address account, uint batch)` is possibly not returning all delegations as it returns in batches of 100. Would this be a problem @ChaituVR ? 
I was also wondering if we can implement proposal restrictions for creation on token holding. Is that something that is supported in snapshot?

ChaituVR: Asking the team if there could be some chance this way

VargaElod23: Hmm, would something like the following be useful:
 - Multicall for all addresses
 - Check for the ones which got `hasMore` as true
 - Multicall only for those addresses again until `hasMore` returns false.

VargaElod23: @ChaituVR WDYT about he approach above?

ChaituVR: What is `hasMore` 🙈 It is a method in your contract? It can cause how many extra multicalls?

VargaElod23: The method that we call returns delegations in a batch of 100 and a bool `returns (Delegation[] delegations, bool hasMore)` . In case there are more than 100 delegations for an address it returns the first 100 and true for the boolean so that you can programmatically loop through to get the remaining ones. Currently the address that has the most delegations on our mainnet has 20-something so we don't expect to have multiple multicalls for a long while. However, in the future it might happen that we need to do multiple multicalls.

VargaElod23:
Hi @ChaituVR ! At last we arrived to the hardfork that enables us to fully comply to your requirements on the multicalls. See the impl in the PR, but it should be good. 🚀

src/strategies/taraxa-delegation/index.ts
```diff
@@ -0,0 +1,34 @@
...
+    multi.call(address, options.address, 'getTotalDelegation', [address])
+  );
+  const result: Record<string, BigNumberish> = await multi.execute();
+  return Object.fromEntries(
+    Object.entries(result).map(([address, balance]) => [
+      getAddress(address),
+      parseFloat(formatUnits(balance, options.decimals))
+    ])
+  );
+}
```
ChaituVR, line 34: Hi @VargaElod23 Looks like this can be achieved by using `contract-call` strategy in your space, give it a try :)

VargaElod23: Fair point. I'm thinking about also including token holders in the score calculation. I just found the `eth-balance` strat, I'll most likely mix those two. Let's keep this PR open until.

VargaElod23:
Hi @ChaituVR . In the last V I added a `native balance + delegation score` strat.

ChaituVR (approved):

VargaElod23:
@ChaituVR I just tried out the strategy in the playground and I get: 
```
curl 'https://score.snapshot.org/api/scores' \
  -H 'authority: score.snapshot.org' \
  -H 'accept: application/json' \
  -H 'accept-language: en-GB,en-US;q=0.9,en;q=0.8' \
  -H 'content-type: application/json' \
  -H 'origin: https://snapshot.org' \
  -H 'referer: https://snapshot.org/' \
  -H 'sec-ch-ua: "Not/A)Brand";v="99", "Google Chrome";v="115", "Chromium";v="115"' \
  -H 'sec-ch-ua-mobile: ?1' \
  -H 'sec-ch-ua-platform: "Android"' \
  -H 'sec-fetch-dest: empty' \
  -H 'sec-fetch-mode: cors' \
  -H 'sec-fetch-site: same-site' \
  -H 'user-agent: Mozilla/5.0 (Linux; Android 6.0; Nexus 5 Build/MRA58N) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/115.0.0.0 Mobile Safari/537.36' \
  --data-raw '{"params":{"space":"","network":"841","snapshot":3325879,"strategies":[{"name":"taraxa-delegation","params":{"address":"0x00000000000000000000000000000000000000fe","symbol":"TARA","decimals":18}}],"addresses":["0xc6a808A6EC3103548f0b38d32DCb6a705B734c89","0xf3F6dBBE59AA219eAcE04955Bd6c3045ab4fF615","0x0dc0d841f962759da25547c686fa440cf6c28c61","0x4a179a005dcbe770c6970ee390a43d2284f67527","0x931d0b36e648906a27a67bc6057579984765e198","0xfc43217e71ec0a1cc480f3d210cd07cbde7374ec","0x21db400dcb1ef3bc3aee4f3d028ec1939b7fadd6","0x1daaa59a8d4a5c08080cadb65204d2d275838a99"]}}' \
  --compressed
```

```
{
    "jsonrpc": "2.0",
    "error": {
        "code": 500,
        "message": "unauthorized",
        "data": {
            "reason": "missing response",
            "code": "SERVER_ERROR",
            "requestBody": "{\"method\":\"eth_blockNumber\",\"params\":[],\"id\":42,\"jsonrpc\":\"2.0\"}",
            "requestMethod": "POST",
            "serverError": {
                "code": "ECONNRESET"
            },
            "url": "https://rpc.snapshot.org/841"
        }
    },
    "id": null
}
```

ChaituVR:
@VargaElod23 We were having few issues at that time, could you try again?

PR #1201, opened 2023-06-22
[math] Implement divide in MATH strategy

### LINKED ISSUE #1200: MATH strategy does not offer complete functionality
### Expected behavior

MATH strategy should allow all common mathematical operations

### Actual behavior

MATH strategy does not support division

### Steps to reproduce the behavior

Try using division with MATH strategy

NOTE: found out that there is good reason not to include addition, revised to just add division support

0xAurelius:
Fixes #1200 

- Add support for division to MATH strategy

### DIFF
--- a/src/strategies/math/README.md
+++ b/src/strategies/math/README.md
@@ -18,6 +18,7 @@ Currently supported operations are:
 | `a-if-gt-b`   | 3             | (x, a, b) = x > b ? a : x                  |
 | `a-if-gte-b`  | 3             | (x, a, b) = x >= b ? a : x                 |
 | `minus`       | 2             | x - a                                      |
+| `divide`      | 2             | x / a                                      |
 
 ## Examples
 
--- a/src/strategies/math/index.ts
+++ b/src/strategies/math/index.ts
@@ -43,6 +43,10 @@ export async function strategy(
   return finalResult;
 }
 
+function throwDivZero() {
+  throw Error("Cannot divide by zero!")
+}
+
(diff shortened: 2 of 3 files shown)


ChaituVR:
Hi @0xAurelius There is a reason why we didn't add `add` to the math strategy, it is because normally  if space is having `5` strategies, the result will be the sum of those `5` strategies

And then we have a limit of `8` strategies per space, if we have `add`, users will use this to bypass that rule to include more strategies😅

0xAurelius:
The use-case we have in mind is actually for permissioning proposal creation based on the total voting power a user has in the space - not summing up voting power for proposal voting

Since we have 5 sources of voting power, we want to allow someone to post a proposal if they have a total number of tokens above some threshold

Could we somehow limit this to proposal creation? Or does proposal creation support summing up the way they normal voting power strategies do?

Divide alone would help us anyway, happy to reduce to just that in this PR

ChaituVR:
Got it, makes sense why you added it, but we now have [proposal validation](https://docs.snapshot.org/user-guides/strategies/validation-strategies#proposal-validation-in-space-settings) now, try to setup this in your space, with basic validation I, it checks if you have that sum of voting power from strategies defined to create a proposal or not :) let me know if you face some issues doing so, 

But yea, happy to merge the divide one :)

0xAurelius:
OK great, I just updated this PR to only add a `divide` operation

ChaituVR (approved):

PR #1205, opened 2023-06-30
[delegate-registry-v2] Add the delegate registry v2 startegy

manboy-eth:
Add the delegate-registry-v2 strategy.


ChaituVR (dismissed):

ChaituVR:
@manboy-eth We need to add this strategy to src/strategies/index.ts

src/strategies/delegate-registry-v2/examples.json
```diff
@@ -0,0 +1,37 @@
+[
+  {
+    "name": "Example query",
+    "strategy": {
+      "name": "delegate-registry",
```
ChaituVR, line 5:
```suggestion
      "name": "delegate-registry-v2",
```

src/strategies/delegate-registry-v2/examples.json
```diff
@@ -0,0 +1,37 @@
...
+      "0x8d5F05270da470e015b67Ab5042BDbE2D2FEFB48",
+      "0x8d07D225a769b7Af3A923481E1FdF49180e6A265",
+      "0x8f60501dE5b9b01F9EAf1214dbE1924aA97F7fd0",
+      "0x9B8e8dD9151260c21CB6D7cc59067cd8DF306D58",
+      "0x17ea92D6FfbAA1c7F6B117c1E9D0c88ABdc8b84C",
+      "0x38C0039247A31F3939baE65e953612125cB88268",
+      "0x6cc5b30Cd0A93C1F85C7868f5F2620AB8c458190",
+      "0x7ef021f62E3E7975FBC21d3202C5A1F19D53bB47",
+      "0x1E1A51E25f2816335cA436D65e9Af7694BE232ad",
+      "0xD028d504316FEc029CFa36bdc3A8f053F6E5a6e4"
```
ChaituVR, line 33:
```suggestion
      "0xA478c2975Ab1Ea89e8196811F51A7B7Ade33eB11",
      "0xeF8305E140ac520225DAf050e2f71d5fBcC543e7",
      "0x1E1A51E25f2816335cA436D65e9Af7694BE232ad",
      "0x1F717Ce8ff07597ee7c408b5623dF40AaAf1787C",
      "0x1c7a9275F2BD5a260A9c31069F77d53473b8ae2e",
      "0x1d5E65a087eBc3d03a294412E46CE5D6882969f4"
```

ChaituVR:
@manboy-eth Can take a look at failing test cases, try running `yarn test --strategy=delegate-registry-v2 in your local

manboy-eth:
> @manboy-eth Can take a look at failing test cases, try running `yarn test --strategy=delegate-registry-v2 in your local

I appreciate your help, @ChaituVR.

I am a little unsure how to proceed. Here's how the voting strategy is supposed to work:

1. Our backend (https://delegate-registry-backend.vercel.app) will return the amount that is delegated to an address, provided the address is not delegating further.
2. This result is merged with other strategies in the Snapshot space, such as erc20-balance-of.
3. We plan to add a validation strategy that ensures only addresses that are not delegating can vote.

Does this approach make sense, or should we structure it differently? Is there a way to automatically force a space to enable a validation strategy if a voting strategy is enabled?

Also, since the strategy has to return at least one address with a score, is it possible to select what snapshot space should be used in the test? If not, could someone send some tokens (the ones referenced in the yam.eth's voting erc20 strategy) to `0xD028d504316FEc029CFa36bdc3A8f053F6E5a6e4`? This would allow me to make a delegation for the yam.eth space.

ChaituVR:
Hi @manboy-eth 
> Does this approach make sense,

Makes sense, but also you could return 0 voting power from this strategy itself and use this strategy in basic validation and check min score. https://docs.snapshot.org/user-guides/strategies/validation-strategies#voting-validation-in-space-settings then we don't need separate validation

> is there a way to automatically force a space to enable a validation strategy if a voting strategy is enabled?

There is no way to force, spaces should enable validation if they want to (maybe you can add this in README)

> since the strategy has to return at least one address with a score, is it possible to select what snapshot space should be used in the test?

You can easily create one on demo.snapshot.org, using goerli. it should be free :)

manboy-eth:
> Makes sense, but also you could return 0 voting power from this strategy itself and use this strategy in basic validation and check min score. https://docs.snapshot.org/user-guides/strategies/validation-strategies#voting-validation-in-space-settings then we don't need separate validation
I am not sure how I can make that work, since we need to be able to distinguish token holders who have delegated and token holders who have not delegated (but still has a score, from other strategies). In both cases, one or more other strategies in the space will return a non-zero value.

I need a validation strategy that can make it possible to vote as long as the delegation registry strategy does not return "0".

I think we have three scenarios (numbers 2 and 3, can probably be counted as the same):
1. When the strategy returns `0` it should not be possible to vote - the address has delegated
2. When the strategy returns `undefined`  it should be able to vote merging all the scores in the space - the address has not delegated and has no delegators delegating to it.
3. When the strategy returns a number it should be able to vote merging all the scores in the space - the address has delegations to it and possible a score from other strategies in the space.

> There is no way to force, spaces should enable validation if they want to (maybe you can add this in README)
Okay.

> You can easily create one on demo.snapshot.org, using goerli. it should be free :)
I meant the space used in the tests when I run `yarn test --strategy=delegate-registry-v2` it looks like it is hardcoded to use the yam.eth space.

manboy-eth:
I created and added the required validation strategy.

ChaituVR:
Hi @manboy-eth I meant adding basic validation directly to your space 😅 Don't have to create a seperate validation

manboy-eth:
> Hi @manboy-eth I meant adding basic validation directly to your space 😅 Don't have to create a seperate validation

But the problem is that basic validation does not work for our requirements (check the comment above: https://github.com/snapshot-labs/snapshot-strategies/pull/1205#issuecomment-1658076977). 
Because of this, we had to create a new validation strategy.

ChaituVR:
oh i get it now, thanks!

But what will be the voting power if delegate-registry-v2 returns `0` ? 🤔

ChaituVR:
Can't depend on `undefined` from strategies, better if strategy can decide on who got voting power or who don't have voting power

auryn-macmillan:
@manboy-eth it seems like we should just encapsulate everything in the delegation strategy, perhaps?

So rather than the current setup, where there is a delegation strategy alongside any number of other strategies...
```json
"strategies": [
    {
      "name": "gno"
    },
    {
      "name": "delegation",
      "params": {
        "strategies": [
          {
            "name": "gno"
          }
        ]
      }
    },
]
```

Perhaps everything should be encapsulated by the delegation strategy. This way there is no need for additional validation outside of the strategy.
```json
"strategies": [
    {
      "name": "delegation",
      "params": {
        "strategies": [
          {
            "name": "gno"
          }
        ]
      }
    },
]
```

manboy-eth:
> oh i get it now, thanks!
> 
> But what will be the voting power if delegate-registry-v2 returns `0` ? 🤔

Then the address has delegated its voting power and should not be able to vote.

manboy-eth:
This is the way it currently works: 
* `undefined` means "address not in the registry",  
* `0` means, address have delegated and 
* a positive number is the number of votes delegated to the address.

Regarding @auryn-macmillan 's suggested approach:
1. @ChaituVR: Can you make the UI for setting up and managing strategies handle this?
2. @ChaituVR: Does this have any impact on your caching systems etc.?
3. Just to make sure I understand, then we would compute everything like we do today but with two adjustments:
  a. When retrieving strategies during the creation of a snapshot, we will get it from the internal strategy settings.
  b. When returning the vote weight, we get our snapshot of vote weights from the delegation registry, then we get the vote weight from each strategy, merge the vote weights as intended, and return the results.

src/validations/strategy-zero-gated/index.ts
```diff
@@ -0,0 +1,30 @@
...
+      this.snapshot || 'latest'
+    );
+
+    if (scores == null || scores.length === 0) return true; // No score returned, allow user to vote
+
+    const score = scores[0] as any[];
+
+    if (score == null || score[0] == null || score.length === 0) return true; // No score returned, allow user to vote
+
+    return score[0] !== 0; // If score is 0, do not allow user to vote, if its greater than 0 or undefined/null, allow user to vote
```
ChaituVR, line 28: Don't think these conditions will work, in `getScoresDirect` we delete all addresses with `0` voting power, better to use basic validation and handle this in strategy itself

ChaituVR (changes requested):

ChaituVR:
> undefined means "address not in the registry",
0 means, address have delegated and
a positive number is the number of votes delegated to the address.

Better to handle this in strategy itself

ChaituVR:
> @ChaituVR: Can you make the UI for setting up and managing strategies handle this?

Sure, create a ticket on https://discord.snapshot.org/ will get in touch to create a test space for you
> @ChaituVR: Does this have any impact on your caching systems etc.?

Using a delegation strategy won't have any impact on caching
> Just to make sure I understand, then we would compute everything like we do today but with two adjustments:
a. When retrieving strategies during the creation of a snapshot, we will get it from the internal strategy settings.


Correct, whenever a proposal is created - we copy strategies to the proposal so it will use these strategies for ever

> b. When returning the vote weight, we get our snapshot of vote weights from the delegation registry, then we get the vote weight from each strategy, merge the vote weights as intended, and return the results.

didn't get it sorry 🙈

manboy-eth:
It sounds like we might have some miscommunication, so I just have to make sure we are on the same page before I start working on this:

So if I create a ticket in Discord, you will make the user interface for configuring any snapshot strategy inside of our strategy settings like Auryn proposes here: https://github.com/snapshot-labs/snapshot-strategies/pull/1205#issuecomment-1705795709. This seems like a potentially big and complicated task to me 🤔 
After you work on the user interface, users will be able to manage strategies (add, remove, or edit settings) nested inside our strategies params. This setting will not be at the normal strategy configuration location `strategies` but will rater be located inside of our strategy configuration like this `strategies[0].params.strategies` (the index is `0` because it will always be only one top-level strategy when the delegation registry strategy is added - the delegation registry strategy).

In other words, you will handle all these special cases (only applicable to the delegation registry strategy):
- When the delegation registry strategy is added to a space, the preexisting strategies will be moved inside of the delegation registry's params settings.
- When the delegation registry is removed, all the strategies inside of the delegation registry's params settings will be placed back in the top-level strategy location.
- When a strategy is added to a space with the delegation registry strategy already added, it will be added under the delegation registry's params. strategies`?
and so on.

This way we will be able to do all the vote weight computations for all strategies in the space, inside of our strategy.

ChaituVR:
I am sorry but i am having a hard time understanding your requirements, What interface you are talking about? Can you explain me an example situation of how you want to calculate your voting power?

manboy-eth:
> I am sorry but i am having a hard time understanding your requirements, What interface you are talking about? Can you explain me an example situation of how you want to calculate your voting power?

We wonder if we can have all the strategies inside of our strategy like @auryn-macmillan suggested (look at the strategy configuration he has suggested in his comment here: https://github.com/snapshot-labs/snapshot-strategies/pull/1205#issuecomment-1705795709).

I am talking about the user interface / GUI / frontend at "https://snapshot.org", and what I think would be required if we were going to do it as @auryn-macmillan suggested.

auryn-macmillan:
I don't think this necessarily requires UI changes in order to function as intended.
With the existing UI, you should already be able to configure the delegation strategy.

auryn-macmillan:
I think that works for the time being, we can improve on it in a subsequent PR.

src/validations/strategy-zero-gated/index.ts
```diff
@@ -0,0 +1,30 @@
+import Validation from '../validation';
+import { getProvider, getScoresDirect } from '../../utils';
+
+export default class extends Validation {
```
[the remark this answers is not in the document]
manboy-eth: The validation strategy is removed.

src/validations/strategy-zero-gated/index.ts
manboy-eth: The validation strategy is removed.

src/strategies/delegate-registry-v2/index.ts
```diff
@@ -0,0 +1,106 @@
...
+    `${options.backendUrl}/api/${space}/snapshot/${blockTag}/strategy-formatted-vote-weights`,
+    {
+      method: 'POST',
+      headers: {
+        Accept: 'application/json',
+        'Content-Type': 'application/json'
+      },
+      body: JSON.stringify({
+        addresses: addresses,
+        strategies: options.strategies
```
ChaituVR, line 43: Could you add a limit to `strategies` , throw a limit just in case more strategies are passed into it

src/strategies/delegate-registry-v2/index.ts
```diff
@@ -0,0 +1,106 @@
...
+  const response = await fetch(
+    `${options.backendUrl}/api/${space}/snapshot/${blockTag}/strategy-formatted-vote-weights`,
+    {
+      method: 'POST',
+      headers: {
+        Accept: 'application/json',
+        'Content-Type': 'application/json'
+      },
+      body: JSON.stringify({
+        addresses: addresses,
```
ChaituVR, line 42: To confirm, voting power of one address doesn't depend on other addresses right?

ChaituVR (changes requested):
Everything else looks good to me

src/strategies/delegate-registry-v2/index.ts
manboy-eth: Do you mean I should set a maximum for how many strategies can be passed?

For instance, by adding something like this at line 33?:
``` typescript
  if (options.strategies.length > 30)
    throw new Error('Maximum 30 strategies allowed');
```

src/strategies/delegate-registry-v2/index.ts
manboy-eth: No, each address should get the same score, no matter what other addresses are passed in the query.

src/strategies/delegate-registry-v2/index.ts
ChaituVR: Yes yes, should be ideally `8` strategies

ChaituVR (approved):

PR #1209, opened 2023-07-05
Add poap-unlock strategy [poap-unlock]

paulperegud:
Adds POAP-unlock voting strategy.

User has one vote if they have any POAPs and zero otherwise. If `eventIds` are specified, than only POAPs of particular event ids are taken into account.

### DIFF
--- a/src/strategies/index.ts
+++ b/src/strategies/index.ts
@@ -193,6 +193,7 @@ import * as unipoolSameToken from './unipool-same-token';
 import * as unipoolUniv2Lp from './unipool-univ2-lp';
 import * as unipoolXSushi from './unipool-xsushi';
 import * as poap from './poap';
+import * as poapUnlock from './poap-unlock';
 import * as poapWithWeight from './poap-with-weight';
 import * as poapWithWeightV2 from './poap-with-weight-v2';
 import * as uniswapV3 from './uniswap-v3';
@@ -652,6 +653,7 @@ const strategies = {
   'unipool-univ2-lp': unipoolUniv2Lp,
   'unipool-xsushi': unipoolXSushi,
   poap: poap,
+  'poap-unlock': poapUnlock,
   'poap-with-weight': poapWithWeight,
   'poap-with-weight-v2': poapWithWeightV2,
   'uniswap-v3': uniswapV3,
--- a/src/strategies/poap-unlock/README.md
+++ b/src/strategies/poap-unlock/README.md
@@ -1,0 +1,14 @@
+# POAP-unlock (erc721)
+
+Each POAP is implemented as an erc721 with a max supply tokens.
+
+Returns 1 if account owns any tokens. If `eventsIds` is passed, than it returns 1 if account owns tokens where the event id is included in `eventIds`.
+
+Here are some examples of parameters:
+
+```json
+{
+  "symbol": "POAP",
+  "eventIds": ["1213", "1293"]
+}
+```
--- a/src/strategies/poap-unlock/examples.json
+++ b/src/strategies/poap-unlock/examples.json
@@ -1,0 +1,19 @@
+[
+  {
+    "name": "Example query",
+    "strategy": {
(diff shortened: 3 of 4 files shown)


ChaituVR (changes requested):
Hi @paulperegud I see this is same as poap strategy except the check for minimum, you can use our voting validation feature for this, https://docs.snapshot.org/user-guides/strategies/validation-strategies#voting-validation-in-space-settings

ChaituVR:
Closing this PR for now, feel free to reopen if needed :)

vpabundance:
@ChaituVR I've been the one looking to set this up.  Under the link provided, this appears to be for Proposal Validation.  Can I ask how this would be applied to a strategy where users with 1, or multiple poaps have the same voting power?

ChaituVR:
Ah sorry, this one https://docs.snapshot.org/user-guides/strategies/validation-strategies#validation-strategy-example-basic 

> an I ask how this would be applied to a strategy where users with 1, or multiple poaps have the same voting power?
You can use it the way you want. after adding voting validation, use ticket strategy to get VP. so it will give one VP

vpabundance:
Apologies for my lack of understanding here. "The Basic validation strategy allows you to use existing Voting Strategies configured for your space or define a custom setup to determine if a user is eligible to create a proposal or cast a vote. "

When I go to set this up, all I am seeing is the ability to add which poap events allow a user to be eligible to vote.  When a user has both poaps in their wallet, they have x2 voting power.

Not seeing where we can limit that scenario to 1 under the link you provided. 

Screenshot 2023-07-11 at 12 48 29 PM

ChaituVR:
So voting validations are just for yes or no, if validation return true, then using space voting strategies, setup a `ticket` strategy like here https://docs.snapshot.org/user-guides/spaces/settings#voting-strategies it will give `1` voting power to everyone who is validated with voting validation

vpabundance:
Wouldn't this require then the use of passport? 
Screenshot 2023-07-11 at 12 55 48 PM

vpabundance:
Gitcoin is one of the 10 projects in this vote, and while we support them, it's not going to reflect well if voting requires the use of one of their products as this would reduce voters outside of their community.

paulperegud:
@ChaituVR This PR is something different from ticket option. Yes, both give user just 1 vote, but ticket option requires user to have gitcoin passport. Please reopen (I lack permissions to do that).

ChaituVR:
Hey guys we don't require gitcoin validation anymore, you can use basic validation too. in basic validation, you can use `poap` strategy and set the minimum score to your desired number of tokens required to vote, and the ticket strategy to give 1 voting power

Sorry those docs are outdated. @zuz could you update that plz :) https://github.com/snapshot-labs/snapshot-strategies/pull/1209#issuecomment-1631429196

vpabundance:
Thanks @ChaituVR , so to clarify, I am setting up 2 different strategies in my settings.  1 for poap strategy and one for ticket, and the combination of the two creates what we need?

vpabundance:
Having trouble getting what you are suggesting to work.  Here is what I have under: valdation>basic:

{
  "symbol": "POAP",
  "eventIds": [
    "128253",
    "128044"
  ]
}

And then additionally ticket with voting power of 1.

This still returns a voting power of 2 should a wallet have 2 poaps.

Are you able to look at the setup on https://snapshot.org/#/octantapp.eth/settings and see what I am missing here?
Thank you for any insight!
Screenshot 2023-07-12 at 3 30 04 PM
Screenshot 2023-07-12 at 3 30 11 PM

zzuziak:
Hey @vpabundance, the setup seems right, and the `ticket` strategy always gives only 1 VP. Did you check it on a new proposal after updating the settings? Keep in mind that the update in settings **affects only new strategies**.

vpabundance:
Thank you so much to everyone here who helped, I successfully created and tested the voting.  Cheers!

PR #1221, opened 2023-07-17
[hats-protocol-single-vote-per-org] Implements hats protocol strategy

Hotmanics:

added a strategy to have a flattened voting weight of 1 depending on if the addresses have at least 1 hat within a single hat tree


Hotmanics:
Apologies, the commit history is a bit messy towards the end. I attempted to fix merge conflicts by altering the yarn.lock file. Which is apparently a big no no. I am not sure of a way to fix this so would appreciate any advice. 

Should I leave the merge conflicts as is?

Thanks!

ChaituVR (changes requested):
Hi @Hotmanics Please revert changes from package lock and yarn lock

Hotmanics:
> Hi @Hotmanics Please revert changes from package lock and yarn lock

Ok I reverted the yarn.lock file and removed package-lock.json

src/strategies/hats-protocol-single-vote-per-org/index.ts
```diff
@@ -0,0 +1,180 @@
+import { subgraphRequest } from '../../utils';
+import { utils } from 'ethers';
+import { StaticJsonRpcProvider } from '@ethersproject/providers';
+import { multicall } from '../../utils';
+
+export const author = 'hotmanics';
+export const version = '0.1.1';
```
ChaituVR, line 7:
```suggestion
export const version = '0.1.0';
```

src/strategies/hats-protocol-single-vote-per-org/index.ts
```diff
@@ -0,0 +1,180 @@
...
+      obj.push({
+        address: this.paths[i],
+        value: r
+      });
+    });
+    this.calls = [];
+    this.paths = [];
+    return obj;
+  }
+}
```
ChaituVR, line 180:
```suggestion
```

src/strategies/hats-protocol-single-vote-per-org/index.ts
```diff
@@ -0,0 +1,134 @@
+import { subgraphRequest } from '../../utils';
+import { utils } from 'ethers';
+import { StaticJsonRpcProvider } from '@ethersproject/providers';
+import { multicall } from '../../utils';
```
ChaituVR, line 4:
```suggestion
import { multicall, Multicaller } from '../../utils';
```

src/strategies/hats-protocol-single-vote-per-org/index.ts
```diff
@@ -0,0 +1,134 @@
...
+      id: true,
+      hats: {
+        id: true,
+        wearers: {
+          id: true
+        }
+      }
+    }
+  };
+  const result = await subgraphRequest(url, params);
```
ChaituVR, line 33: @Hotmanics You have to pass `snapshot` into `block` arg here

src/strategies/hats-protocol-single-vote-per-org/index.ts
Hotmanics: I don't think that I want to import the Multicaller from the utils, as the implementation found in this file itself is a modified version of it.

src/strategies/hats-protocol-single-vote-per-org/index.ts
Hotmanics: same thing here, I am using a modified version of the Multicaller class.

ChaituVR: @Hotmanics Main idea is to avoid duplicate code as much as possible. I see you change and feel like we can modify the same data after executing the multicaller

Hotmanics: @ChaituVR From my understanding, the execute() function from the utils.ts file flattens the values.
For instance, if I have several calls like so:
```
    multi.call('0xaddress1', 'SmartContractAddress', 'isWearerOfHat', ['0xaddres1' '0xHatId1']);
    multi.call('0xaddress1', 'SmartContractAddress', 'isWearerOfHat', ['0xaddres1' '0xHatId2']);
    multi.call('0xaddress2', 'SmartContractAddress', 'isWearerOfHat', ['0xaddres2' '0xHatId1']);
```
Then the returned value from 'execute' will be something like so:
```
    {
       '0xaddress1': true,
       '0xaddress2': true
   }

```
When the desired result is something like:
```
    {
       '0xaddress1': true,
       '0xaddress1': true,
       '0xaddress2': true
   }
```

ChaituVR: It return entire value https://github.com/snapshot-labs/snapshot.js/blob/master/src/utils/multicaller.ts#L40C54-L40C71 you can modify result after getting it

Hotmanics: > It return entire value https://github.com/snapshot-labs/snapshot.js/blob/master/src/utils/multicaller.ts#L40C54-L40C71 you can modify result after getting it

I don't believe so unless I'm looking at incorrectly.

here is what my Multicaller returns. You can see that it returns multiple entries for 0xa6aF0566EF4eF7E8f38913f69d4e55c06F00A5aC.
Screenshot 2023-07-20 at 4 28 03 PM

here is what snapshot's Multicaller returns. You can see that it removes all extra entries pertaining to 0xa6aF0566EF4eF7E8f38913f69d4e55c06F00A5aC and gives one result.
Screenshot 2023-07-20 at 4 29 47 PM

Apologies if I am missing something here. This is just what I am seeing on my end.

ChaituVR:
Hi @Hotmanics Looks like test cases are failing can check>? also please resolve conflicts on yarn.lock. it should not have any differences when you see here https://github.com/snapshot-labs/snapshot-strategies/pull/1221/files

Hotmanics:
> Hi @Hotmanics Looks like test cases are failing can check>? also please resolve conflicts on yarn.lock. it should not have any differences when you see here https://github.com/snapshot-labs/snapshot-strategies/pull/1221/files

@ChaituVR I fixed the test cases, but I am not sure on how to resolve the yarn.lock issue. I am copying the raw file from snapshot-labs:master and pasting it into my yarn.lock. my local git doesn't recognize any changes and if you look at the diff on Github that you reference in the link, then you notice there are literally no changes aside from the removal of one set of text with a new set of text. Of which, both are completely identical. So idk what to do here.

Hotmanics:
Feel free to make commits to this PR if you think you can solve the yarn.lock issue. I have absolutely no clue why its recognizing a difference and am hoping to get this PR finished ASAP. Glad to hop on a google meets or something to just get this resolved.

src/strategies/hats-protocol-single-vote-per-org/index.ts
```diff
@@ -0,0 +1,191 @@
...
+  let result;
+
+  switch (network) {
+    case '1':
+      result = await subgraphRequestHats(
+        'https://api.thegraph.com/subgraphs/name/hats-protocol/hats-v1-ethereum',
+        snapshot,
+        options.humanReadableTreeId
+      );
+    case '10':
```
ChaituVR, line 77:
```suggestion
      result = await subgraphRequestHats(
        'https://api.thegraph.com/subgraphs/name/hats-protocol/hats-v1-ethereum',
        snapshot,
        options.humanReadableTreeId
      );
      break;
    case '10':
```

ChaituVR (approved):

PR #1222, opened 2023-07-17
New n3 passport voting strategy

johnmark13:
Adds a new voting strategy that currently just looks for the signer addresses of all passports and assigns them a voting weight of 1.

Commented out code left in as it would be required if we wanted to do a balance check for owner balances > 2 veNation, but would propose to remove all this and handle that elsewhere.


johnmark13:
Opened in error.

PR #1231, opened 2023-07-26
[wagumi-sbt]add wagumi-sbt method

ikisuke:
Additional snapshot strategies for Wagumi sbt holders


ikisuke:
I may also use api-v2, as it is currently a viable alternative.

However, we will need to add a dedicated strategy in the future and will contact you again at that time.

@ChaituVR

ChaituVR:
For sure :) Feel free to create new one when ever you are ready

PR #1242, opened 2023-08-02
[gelato-staking] add gelato-staking strategy

alxdca:
- Add a strategy that uses the Gelato Network staking contract 

To avoid double voting, the strategy only allows the manager to vote if the staker has set one.

### DIFF
--- a/src/strategies/gelato-staking/index.ts
+++ b/src/strategies/gelato-staking/index.ts
@@ -2,7 +2,7 @@ import { BigNumber } from '@ethersproject/bignumber';
 import { Multicaller } from '../../utils';
 import { formatUnits } from '@ethersproject/units';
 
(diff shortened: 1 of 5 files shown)


ChaituVR (approved):

src/strategies/gelato-staking/index.ts
```diff
@@ -0,0 +1,43 @@
+import { BigNumber } from '@ethersproject/bignumber';
+import { Multicaller } from '../../utils';
+import { formatUnits } from '@ethersproject/units';
+
+export const author = 'alxdca:feat';
```
ChaituVR, line 5:
```suggestion
export const author = 'alxdca';
```

ChaituVR (approved):

PR #1249, opened 2023-08-04
[sd-vote-boost-twavp-v2] add: sd-vote-boost-twavp-v2

clement-ux:
Fixes # .

- add new strategy for StakeDAO
- implement whitelist to bypass twavp

### DIFF
--- a/src/strategies/sd-vote-boost-twavp-v2/examples.json
+++ b/src/strategies/sd-vote-boost-twavp-v2/examples.json
@@ -1,0 +1,28 @@
+[
+  {
+    "name": "Stake DAO vote boost using TWAVP V2",
+    "strategy": {
+      "name": "sd-vote-boost-twavp-v2",
+      "params": {
+        "veToken": "0x5f3b5DfEb7B28CDbD7FAba78963EE202a494e2A2",
+        "liquidLocker": "0x52f541764E6e90eeBc5c21Ff570De0e2D63766B6",
+        "sdTokenGauge": "0x7f50786A0b15723D741727882ee99a0BF34e3466",
+        "symbol": "sdToken",
+        "decimals": 18,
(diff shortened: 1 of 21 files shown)


src/strategies/sd-vote-boost-twavp-v2/examples.json
```diff
@@ -0,0 +1,28 @@
...
+        "veToken": "0x5f3b5DfEb7B28CDbD7FAba78963EE202a494e2A2",
+        "liquidLocker": "0x52f541764E6e90eeBc5c21Ff570De0e2D63766B6",
+        "sdTokenGauge": "0x7f50786A0b15723D741727882ee99a0BF34e3466",
+        "symbol": "sdToken",
+        "decimals": 18,
+        "sampleSize": 10,
+        "sampleStep": 5,
+        "whiteListedAddress": [
+          "0x1c0D72a330F2768dAF718DEf8A19BAb019EEAd09"
+        ]
```
ChaituVR, line 16: Hi @clement-ux Throw an error if someone passes more addresses to this. so we can reduce memory issues :)

clement-ux: Hi @ChaituVR , what could be the maximum? Because I passed one as example, but there may be more than one.

clement-ux: For the moment it is only few, it will be adjusted on the way. But something like 10-20 maximumI guess. Can i leave it like this, or should I throw an error if this is above a certain maximum?

ChaituVR: yep, that would be nice :)

ChaituVR (approved):

PR #1250, opened 2023-08-04
Add Aave DAO proposal validation strategy

pyk:
- Add Aave proposal validation strategy

### DIFF
--- a/src/validations/aave/examples.json
+++ b/src/validations/aave/examples.json
@@ -1,0 +1,19 @@
+[
+  {
+    "name": "Example of Aave validation",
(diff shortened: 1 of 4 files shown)


ChaituVR (changes requested):
Hi @pyk Looks like it is same as Basic validation, can you try to use the same on your space than a new validation?

pyk:
Hi @ChaituVR,

Thank you so much for the review!

> Looks like it is same as Basic validation, can you try to use the same on your space than a new validation?

I implemented basic validation and added whitelist on top, based on the spec from the AAVE team. You're right, they're dropping this task and looking for an existing solution. I'll close this pull request for now.

PR #1255, opened 2023-08-18
fix: throw error when required options is missing

wa0x6e:
Related to https://github.com/snapshot-labs/score-api/issues/838

When the `options.tokenId` params is missing, this strategy will just throw a cryptic `Error: invalid BigNumber value (argument="value", value=undefined, code=INVALID_ARGUMENT, version=bignumber/5.7.0)` exception, caused by BigNumber trying to parse the tokenId which is `undefined`

This PR will throw a more meaningful error message when tokenId is not set


wa0x6e:
Closing, in favor of using schema for validation

PR #1266, opened 2023-08-28
[razor-network-voting] update strategy

rajkharvar:
- Update network and addresses in `examples.json`.
- Update strategy to support paginated data.


src/strategies/razor-network-voting/index.ts
```diff
@@ -34,43 +97,10 @@ export async function strategy(
...
-    // @ts-ignore
-    params.delegators.__args.block = { number: snapshot };
-  }
-
   const score = {};
 
-  // subgraph request 1 : it fetches all the details of the stakers and delegators.
-  const result = await subgraphRequest(RAZOR_NETWORK_SUBGRAPH_URL, params);
+  // subgraph request : it fetches all the details of the stakers and delegators.
+  const result = await getAllData(snapshot);
```
ChaituVR, line 103: How many network requests it can send? 😅

ChaituVR: Ideally, it should not send more than 5 requests https://github.com/snapshot-labs/snapshot-strategies#checklist-for-adding-a-new-voting-strategy-or-validation-strategy

rajkharvar: It's highly unlikely that we will surpass the upper limit of more than five requests.

ChaituVR: Then it should be cool :)

PR #1270, opened 2023-08-30
[passport-gated]: Update implementation to include Passport scoring mechanism

santteegt:
This PR mainly aims to update the `passport-gated` validation strategy to fully integrate with Gitcoin Passport API for validating users using a Passport stamp collection and scoring threshold. This work is a [collaboration](https://dm.raidguild.org/raids/adb5854a-94a2-4424-8a1f-c5c36b08093c) between RaidGuild and the Gitcoin Passport teams.

- Update the `passport-gated` strategy implementation to use a scorer mechanism to calculate a Passport score for evaluating the user's eligibility.
- Update documentation to include implementation details and setup requirements.
- Update the `validation.test.ts` suite to allow evaluating multiple examples per validation strategy. Changes also enable the inclusion of true negative samples.
- Remove `passport-weighted` validation strategy from the repository as it is now deprecated.

*UPDATES*:

### Demo

[Link to Video](https://youtu.be/eJQcVPEsNWc)

Proposed changes also require to update other components in the Snapshot stack:
- https://github.com/snapshot-labs/snapshot/pull/4199
- https://github.com/snapshot-labs/score-api/pull/876


plor:
I'm part of the RaidGuild team working on this and just wanted to weigh in on this to get the test to pass here. This PR adds a call to the passport API in order to score the passport to check whether it meets the threshold. To do this there are two new ENV variables required. These can be created at https://scorer.gitcoin.co/ so that the snapshot team can properly manage this secret key.

Basically the ENV variables needed are:
`PASSPORT_API_KEY` and `PASSPORT_SCORER_ID`

To create the `PASSPORT_SCORER_ID` a simple scorer can be created of the type `Unique Humanity`. I'm happy to discuss this further on the Discord server if more synchronous coordination is needed here.

santteegt:
Hey,

Test action is failing because the PR title was changed to `[passport-weighted]` but the strategy is under `passport-gated`. Let me know if I should update my PR to change the directory name or just need to adjust the PR title.

Regarding required env variables, it would be good to set them as Action secret variables on Github.

ChaituVR:
Hi @santteegt Thanks for the suggestions, 

for 
> To create the PASSPORT_SCORER_ID a simple scorer can be created of the type Unique Humanity

Do you mean every space will use the same `Unique humanity`?

santteegt:
> Hi @santteegt Thanks for the suggestions,
> 
> for
> 
> > To create the PASSPORT_SCORER_ID a simple scorer can be created of the type Unique Humanity
> 
> Do you mean every space will use the same `Unique humanity`?

Yeah, the Gitcoin team requested to use the `Unique Humanity` scorer by default.

ChaituVR:
Hi @santteegt Now that `scoreThreshold` is mandatory, it may break existing space's configurations, Is it possible to make it optional? or we should plan to announce that it won't work in the future, 

CC @zzuziak

zzuziak:
> Hi @santteegt Now that `scoreThreshold` is mandatory, it may break existing space's configurations, Is it possible to make it optional? or we should plan to announce that it won't work in the future,
> 
> CC @zzuziak

From what I remember we agreed that the threshold for Spaces already using the validation should be set to `0` 🤔

ChaituVR:
Could we default it to zero if undefined? 🤔

santteegt:
> Hi @santteegt Now that `scoreThreshold` is mandatory, it may break existing space's configurations, Is it possible to make it optional? or we should plan to announce that it won't work in the future,
> 
> CC @zzuziak

you're right. I'll set the parameter as optional with zero value by default

.github/workflows/validation.yml
```diff
@@ -18,6 +18,13 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2.3.4
+
+    - name: Create env file
+      run: |
+        touch .env
+        echo PASSPORT_API_KEY=${{ secrets.PASSPORT_API_KEY }}  >> .env
+        echo PASSPORT_SCORER_ID=${{ secrets.PASSPORT_SCORER_ID }}  >> .env
+
```
santteegt, line 27: @ChaituVR added this to the workflow so it can use the Passport keys you set as secret variables

ChaituVR: Yea already added them, but not sure why it doesn't read 🤔

ChaituVR: Works well in local

zzuziak:
Hey @santteegt, amazing work 🔥  Could you also possibly share screenshots or a short Loom recording with the Validation in UI?

.github/workflows/validation.yml
santteegt: tested locally using [https://github.com/nektos/act](act) and it's working. There might be an issue with how you're setting the secret env vars in the repo settings

santteegt:
> Hey @santteegt, amazing work 🔥 Could you also possibly share screenshots or a short Loom recording with the Validation in UI?

Sure! But don't know how to load the updated strategies on a local snapshot frontend though. I was exploring the frontend repo on how to do that but found that the latest validation strategies metadata is being fetched from a [https://score.snapshot.org/api/validations](API endpoint). Could you point me to a good guide on how to setup both the frontend and strategies locally?

santteegt:
Hey @zzuziak 

I added a link to the demo in the PR description. While recording the video I noticed that we also required to update other components in your stack. More specifically, the Scorer API and the frontend. I already submitted a couple PRs with those changes, you can find them listed at the top of this PR as well

src/validations/passport-gated/index.ts
```diff
@@ -27,46 +50,158 @@ function hasValidIssuanceAndExpiration(credential, proposalTs) {
...
+    const validStamps = await validateStamps(
+      currentAddress,
+      operator,
+      proposalTs,
+      requiredStamps
+    );
+    const validScore = await validatePassportScore(
+      currentAddress,
+      scoreThreshold
+    );
```
ChaituVR, line 203: HI @santteegt This should not be called if there is no `scoreThreshold`

ChaituVR: This address `0x24F15402C6Bb870554489b2fd2049A85d75B982f` got ENS stamp but it still returns `false` because of the passport score 🤔

src/validations/passport-gated/index.ts
```diff
@@ -27,46 +50,158 @@ function hasValidIssuanceAndExpiration(credential, proposalTs) {
...
+
+    if (scoreResponse.ok && scoreData.status === 'DONE') {
+      return evalPassportScore(scoreData, scoreThreshold);
+    }
+    console.log(
+      `[passport] Waiting for scorer... (${i}/${PASSPORT_SCORER_MAX_ATTEMPTS})`
+    );
+    await snapshot.utils.sleep(3e3);
+  }
+  return false;
```
ChaituVR, line 175: If we still wait for scorer after 2 attempts, we should return an error instead of `false`

ChaituVR (changes requested):
So for `validStamps` we are checking snapshot block time to check if user has stamps only before proposal creation, but for `validatePassportScore` we are not checking any time related score,  maybe we should make it optional for validStamps (default to check time)

santteegt:
> So for `validStamps` we are checking snapshot block time to check if user has stamps only before proposal creation, but for `validatePassportScore` we are not checking any time related score, maybe we should make it optional for validStamps (default to check time)

Checking passport stamps validity is required to ensure a member has a valid issued stamp before proposal creation / proposal voting timesstamp. Passport score however could be calculated at any time by independent Scorers (with `PASSPOT_SCORER_ID`) so it's not relevant to eval `last_score_timestamp` as soon the passport stamps have been issued before a validation event

0xAurelius:
Once this is finalized and merged, would it be reasonable for me to open a follow-up PR to add a "at-least-n" operator option and corresponding "n" parameter?

The use-case is that the DAO I contribute to would like to require at least n=2 of the approved stamps (regardless of Score)...

0xZakk:
Hey @ChaituVR, wondering what next steps are required to get this finalized and merged?

ChaituVR:
I think we are good with this. we planning to release this tomorrow (20th sept)

ChaituVR:
FYI, just waiting for a small doubt to be answered in our telegram chat, will merge it soon after that :)

0xZakk:
@santteegt - while testing, @ChaituVR ran into an issue where he updated his score in the Passport UI and then still wasn't able to vote with the strategy. We need to submit the address to be scored, then fetch the score. Even though this ends up hitting the API a lot, I think the best approach might be to submit the score each time (which will often give you the score back immediately, in the response object)

santteegt:
@0xZakk @ChaituVR thanks for the feedback. I made a quick update on the implementation to consider that scenario. The PR is ready for another review

0xZakk:
@santteegt thank you! That was fast! Looks good to me. The work day is finished for @ChaituVR, so he'll review it by tomorrow morning (our time)

src/validations/passport-gated/README.md
```diff
@@ -1,29 +1,74 @@
...
+### September 21, 2023
+
+Code is fully integrated with Passport API. In order to evaluate if an address is eligible for creating/voting on a proposa, It checks for passport stamp collection and minimum score using the `Unique Humanity` scoring mechanism, however it could be upgraded in the future to support other scoring algorithms.
+
+Unit tests now support multiple example Passports to evaluate different validation scenarios.
+
 ## Last Modified
 
-This code was last modified on June 1, 2023.
+This code was last modified on August 21, 2023.
```
ChaituVR, line 72:
```suggestion
This code was last modified on September 21, 2023.
```

src/validations/passport-gated/index.ts
```diff
@@ -27,46 +49,145 @@ function hasValidIssuanceAndExpiration(credential, proposalTs) {
...
+
+function evalPassportScore(scoreData: any, minimumThreshold = 0): boolean {
+  // scoreData.evidence?.type === 'ThresholdScoreCheck' -> Returned if using Boolean Unique Humanity Scorer (should not be used)
+  if (scoreData.evidence?.type === 'ThresholdScoreCheck') {
+    return (
+      Number(scoreData.evidence.rawScore) > Number(scoreData.evidence.threshold)
+    );
+  }
+  // scoreData.score -> Returned if using Unique Humanity Score
+  return Number(scoreData.score) > minimumThreshold;
```
ChaituVR, line 109: Threshold should include minimum value also right?? like if `scoreThreshold` is `17.62`, any validation with a score of `17.61` or lower would be considered invalid, while any validation with a score of `17.62` or higher would be considered valid. no?

here it is checking from `17.63` and more to be valid

```suggestion
  return Number(scoreData.score) >= minimumThreshold;
```

src/validations/passport-gated/index.ts
```diff
@@ -27,46 +49,145 @@ function hasValidIssuanceAndExpiration(credential, proposalTs) {
...
+  }
+
+  // Scorer done calculating passport score during submission
+  if (submittedPassport.ok && submissionData.status === 'DONE') {
+    return evalPassportScore(submissionData, scoreThreshold);
+  }
+
+  // Try to fetch Passport Score if still processing (submittedPassport.status === 'PROCESSING')
+  for (let i = 0; i < PASSPORT_SCORER_MAX_ATTEMPTS; i++) {
+    const scoreResponse = await fetch(GET_PASSPORT_SCORE_URI + currentAddress, {
```
ChaituVR, line 140: This API will return only DONE and PROCESSING? 🤔

ChaituVR (changes requested):
Everything else looks good :) 👍 Thanks for the quick fixes

src/validations/passport-gated/index.ts
santteegt: right. As it always hits the `/submit-passport` endpoint first, the `GET_PASSPORT_SCORE_URI` will be called only if the former returns `PROCESSING` status

ChaituVR (approved):
Thank you ❤️ 


@zzuziak  will take care of documentations and announcements very soon :)

PR #1271, opened 2023-08-30
[rareprotocol-staked-balance]: Staked RARE voting strategy

charlescrain:
Fixes # .

- Added a strategy, where the balance includes RARE that has been staked

### DIFF
--- a/src/strategies/rareprotocol-staked-balance/index.ts
+++ b/src/strategies/rareprotocol-staked-balance/index.ts
@@ -1,0 +1,43 @@
+import { BigNumberish } from '@ethersproject/bignumber';
+import { formatUnits } from '@ethersproject/units';
+import { Multicaller } from '../../utils';
+import { strategy as erc20BalanceOfStrategy } from '../erc20-balance-of';
+
+export const author = 'rareprotocol';
+export const version = '0.0.1';
+
+const abi = [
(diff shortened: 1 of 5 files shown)


src/strategies/rareprotocol-staked-balance/index.ts
```diff
@@ -0,0 +1,43 @@
...
+    provider,
+    addresses,
+    options,
+    snapshot
+  );
+  const blockTag = typeof snapshot === 'number' ? snapshot : 'latest';
+
+  const multi = new Multicaller(network, provider, abi, { blockTag });
+  addresses.forEach((address) =>
+    multi.call(address, options.rareStakingRegistryAddress, 'getTotalAmountStakedByUser', [address])
```
ChaituVR, line 33: Hi @charlescrain Better to use `erc20-balance-of` strategy with `contract-call` strategy on your space, the result will be the sum of both

ChaituVR (changes requested):

src/strategies/rareprotocol-staked-balance/index.ts
charlescrain: @ChaituVR Thanks for pointing that out will close!

charlescrain:
Closing because, this can be achieved with existing strategies

PR #1289, opened 2023-09-25
[dss-vest-balance-and-unpaid] Add DssVest vesting contract strategies

espendk:
[_This PR builds on PR #1290_]

This PR adds a strategy that allow voters to vote based on both their current balance of a token and vested but yet unclaimed tokens in a [DssVest](https://github.com/makerdao/dss-vest) token vesting contract.

The reason for having this as one strategy instead of using `erc20-balance-of` and `dss-vest-unpaid` in a space is to ensure that the UI will show one score for the token in question. If one uses two separate strategies, the UI breaks down the voting power as follows:

Screenshot 2023-09-26 at 07 55 09

This will be used by Mangrove DAO. We use a variant of DssVest for vestings, but think that a more general strategy might be useful to others.


ChaituVR (changes requested):
Hey @espendk better to create separate PR for each strategy

src/strategies/dss-vest-balance-and-unpaid/index.ts
```diff
@@ -0,0 +1,37 @@
...
+    network,
+    provider,
+    addresses,
+    { address: options.tokenAddress, ...options },
+    snapshot
+  );
+
+  for (const address of Object.keys(score)) {
+    score[address] += mgvScore[address];
+  }
```
ChaituVR, line 34: Hey @espendk You can use dss-vest-unpai and erc20-balance-of in your space, result will be sum of those two strategies, let me know if it doesn't work for you

espendk: Thanks for the suggestion :-) The thing is, we wish the UI to show the sum of these tokens as one, not two tokens.
If we use `dss-vest-unpaid` and `erc20-balance-of` as separate strategies in our space, the UI shows them separately:

billede

The reason this is important is that we'll be adding two other strategies with separate symbols (TBD, but maybe `Builders` and `Pods`) and we'd like the UI to show eg `2 MGV + 3 Builders + 5 Pods`.

espendk:
> Hey @espendk better to create separate PR for each strategy

Ok, I'll split them 👍 I'll keep this one for the `dss-vest-balance-and-unpaid` strat since the review comment pertains to that. Let me know if that works.

espendk:
I split `dss-vest-unpaid` out: https://github.com/snapshot-labs/snapshot-strategies/pull/1290

src/strategies/dss-vest-balance-and-unpaid/index.ts
ChaituVR: you will be using both `dss-vest-unpaid` and `dss-vest-balance-and-unpaid` in your space?

espendk: I suspect that we'll only be using `dss-vest-balance-and-unpaid`, but it's not clear at this time.

I mainly factored `dss-vest-unpaid` out into a separate strategy as it seemed generally useful to other projects that use DssVest for vesting. And those projects may not wish to mix it with `erc20-balance-of` like we do.

ChaituVR:
https://github.com/snapshot-labs/snapshot-strategies/pull/1290 is merged, could you update this branch :)

espendk:
> #1290 is merged, could you update this branch :)

Done 👍 :-)

ChaituVR (approved):

PR #1290, opened 2023-09-26
[dss-vest-unpaid] Add DssVest dss-vest-unpaid strategy

espendk:
This PR adds a strategy that allows voters to vote based on vested but yet unclaimed tokens in a [DssVest](https://github.com/makerdao/dss-vest) token vesting contract.

This will be used by Mangrove DAO. We use a variant of DssVest for vestings, but think that a more general strategy might be useful to others.

### DIFF
--- a/src/strategies/dss-vest-unpaid/index.ts
+++ b/src/strategies/dss-vest-unpaid/index.ts
@@ -6,6 +6,9 @@ import { Multicaller } from '../../utils';
 export const author = 'espendk';
 export const version = '1.0.0';
 
+// To avoid future memory issues, we limit the number of vestings supported by the strategy
+const MAX_VESTINGS = 500;
+
 const abi = [
   'function ids() external view returns (uint256)',
   'function usr(uint256 id) external view returns (address)',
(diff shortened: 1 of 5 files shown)


espendk:
I split `dss-vest-unpaid` out: https://github.com/snapshot-labs/snapshot-strategies/pull/1290

src/strategies/dss-vest-unpaid/index.ts
```diff
@@ -0,0 +1,55 @@
...
+): Promise<Record<string, number>> {
+  const blockTag = typeof snapshot === 'number' ? snapshot : 'latest';
+
+  // Get the number of vestings
+  const dssVestContract = new Contract(options.address, abi, provider);
+  const idCount = await dssVestContract.ids();
+
+  // Get the vesting addresses and unpaid amounts
+  const multi = new Multicaller(network, provider, abi, { blockTag });
+  for (let id = 1; id <= idCount; ++id) {
```
ChaituVR, line 31: Could we add a limit to idCount here, to avoid memory issues in future, you can return an error if we get more than what you expect

ChaituVR (changes requested):

src/strategies/dss-vest-unpaid/index.ts
espendk: Sure. Added in [`369cbdc` (#1290)](https://github.com/snapshot-labs/snapshot-strategies/pull/1290/commits/369cbdc8442cbecac1151cf6bd19fbbe5c771b1f)

ChaituVR (approved):

PR #1301, opened 2023-10-05
[api-v2] Send secret in header to know requests are coming from snapshot

ChaituVR:
More details: https://discord.com/channels/955773041898573854/1146484874644050103/1154670952983048212

We need to send a secret in header for APIs (similar to webhook for API creator to know that these requests are coming from snapshot only

TODO:
- [ ] Add `SNAPSHOT_API_STRATEGY_SALT` to score-api in digital ocean


Todmy (approved):
utACK

PR #1309, opened 2023-10-11
[passport-gated]: update required parameters

santteegt:
The Gitcoin passport team requested to update the `stamps` property as an optional parameter in the `passport-gated` validation strategy.

- Make `stamps` property as optional parameter
- Update `operator` parameter to fit different scenarios
- Update unit tests

Related PRs in other repos:

- https://github.com/snapshot-labs/snapshot/pull/4279

### DIFF
--- a/src/validations/passport-gated/index.ts
+++ b/src/validations/passport-gated/index.ts
@@ -167,8 +167,10 @@ export default class extends Validation {
   async validate(currentAddress = this.author): Promise<boolean> {
     const requiredStamps = this.params.stamps || [];
     const operator = this.params.operator;
-    const scoreThreshold = this.params.scoreThreshold || 0;
-    if (!operator) throw new Error('Operator is required');
+    const scoreThreshold = this.params.scoreThreshold;
+
+    if (scoreThreshold === undefined) throw new Error('Score threshold is required');
+    if (requiredStamps.length > 0 && !operator) throw new Error('Operator is required');
 
     const provider = snapshot.utils.getProvider(this.network);
     const proposalTs = (await provider.getBlock(this.snapshot)).timestamp;
--- a/src/validations/passport-gated/schema.json
+++ b/src/validations/passport-gated/schema.json
@@ -9,8 +9,9 @@
         "stamps": {
           "type": "array",
           "title": "Stamps",
+          "description": "List of Eligible Passport Stamps (Optional).",
           "uniqueItems": true,
-          "minItems": 1,
+          "minItems": 0,
           "maxItems": 27,
           "items": {
             "type": "string",
@@ -129,7 +130,7 @@
         "operator": {
           "type": "string",
           "title": "Approval operator",
-          "description": "Control how many or which stamps are required to vote.",
+          "description": "Control how many stamps are required to become eligible.",
           "anyOf": [
             {
               "const": "AND",
@@ -138,19 +139,23 @@
             {
               "const": "OR",
               "title": "Require at least one stamp"
+            },
+            {
+              "const": "",
+              "title": "Any (Stamps MUST be empty)"
             }
           ]
         },
         "scoreThreshold": {
           "type": "number",
           "title": "Passport Score Threshold",
-          "description": "Minimum Passport score threshold (>) to be eligible as a valid Passport",
+          "description": "Minimum Passport score threshold (>=) to be eligible as a valid Passport",
           "minimum": 0,
           "maximum": 100,
           "default": 0
         }
       },
-      "required": ["stamps", "operator"],
+      "required": ["scoreThreshold"],
       "additionalProperties": false
     }
   }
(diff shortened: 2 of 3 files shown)


santteegt:
This is an open source contribution. I'd appreciate it if you guys add the `hacktoberfest-accepted` label to this PR

src/validations/passport-gated/schema.json
```diff
@@ -138,19 +139,23 @@
...
           "title": "Passport Score Threshold",
-          "description": "Minimum Passport score threshold (>) to be eligible as a valid Passport",
+          "description": "Minimum Passport score threshold (>=) to be eligible as a valid Passport",
           "minimum": 0,
           "maximum": 100,
           "default": 0
         }
       },
-      "required": ["stamps", "operator"],
+      "required": ["scoreThreshold"],
```
samuveth, line 158: What exactly is `scoreThreshold` and why is it required now?

santteegt: `scoreThreshold` is the minimum score a Passport must fulfil to become eligible. The Passport team requested to change the required parameters for this strategy, I think to cover cases where a user can have any valid stamp (>0). Before, the only options were `at least one (>=1) (OR operand)` and `all stamps (m == n) (AND operand)`. @0xZakk could give a more detailed explanation of why they're looking to implement these changes

samuveth: When you say "any valid stamp (>0)," does that mean they're looking to include cases where, say, 2 out of 5 stamps would make you eligible? Would love to get a bit more clarity on this, maybe @0xZakk can weigh in?

lebraat: Very sorry for the delayed response. 

It's a best practice to require a score threshold as opposed to specific Stamps, so we made the recommendation for this update. 

Requiring a specific Stamp will not work for all use cases, especially if there isn't a specific Stamp that makes sense for their use case. Because of this, we suggested that we make any Stamp selection optional. 

The 'require at least one Stamp' is not something we've recommended in the past, but I could see a potential use case there. I would guess that this option will rarely be used.

zzuziak: This is to ensure the transition of the validation, so that spaces which are using stamps at the minute will not see an error caused by lack of minimum threshold set (which is a new param for the validation). 

Is that correct @lebraat ?

lebraat: Truth. We don't want any of the early adopters to experience any errors. 

Regardless, risk is low that this will cause much user confusion/frustration considering it's still marked as 'beta'. We have held off on doing a substantial marketing push until this PR is merged.

Is there a way for you all to pull a list of Spaces that are utilizing this functionality?

ChaituVR: Hi @lebraat Sorry for the delayed response, right now there is no way to get spaces with validation but you can query proposals using our API https://hub.snapshot.org/graphql

```graphql
{
  proposals(first: 1000, where: { validation: "passport-gated" }) {
    title
    id
    space {
      id
    }
    validation {
      name
      params
    }
    created
  }
}
```

ChaituVR: These changes look's good to me, we are just worried about space's questions in our support 😄 but we can go with this if you think you had enough marketing push, let me know.

ChaituVR (approved):

src/validations/passport-gated/schema.json
lebraat: Feel free to send support messages my way so I can get a good understanding of what those look like. I can also work with you all to make adjustments to the docs to make things more clear, but I believe the updates in this PR will make the integration much more intuitive for users.

samuveth (approved):

src/validations/passport-gated/index.ts
```diff
@@ -167,8 +167,10 @@ export default class extends Validation {
   async validate(currentAddress = this.author): Promise<boolean> {
     const requiredStamps = this.params.stamps || [];
     const operator = this.params.operator;
-    const scoreThreshold = this.params.scoreThreshold || 0;
-    if (!operator) throw new Error('Operator is required');
+    const scoreThreshold = this.params.scoreThreshold;
+
+    if (scoreThreshold === undefined) throw new Error('Score threshold is required');
+    if (requiredStamps.length > 0 && !operator) throw new Error('Operator is required');
```
samuveth, line 173: Maybe `Operator is required when selecting required stamps`

lebraat: Something we can definitely consider, but wouldn't consider that a blocker for this work. What are some next steps we can take to merge this one?

ChaituVR:
```suggestion
    if (requiredStamps.length > 0 && !operator) throw new Error('Operator is required when selecting required stamps');
```

ChaituVR: I will change this in a separate PR (no access to add commit to this PR)

PR #1319, opened 2023-10-14
chore: auto run linter on new commit

wa0x6e:
## 🧿 Current issues / What's wrong ?

Current lint task is using `--fix`, which will not raise any errors.

## 💊 Fixes / Solution

- Remove the `--fix` from the `yarn lint` command
- Add a new `yarn lint:fix` task to keep old auto-fixing behavior
- Rename the lint git workflow to fix-lint (should be obsolete, but kept in case)
- Add a new lint github action, which will run on new commit, and warn on lint and typescript errors
- Fix lint errors on some strategies


ChaituVR (approved):

ChaituVR (approved):
utACk

PR #1327, opened 2023-10-19
Added Balance of At Scaled as well as erc721 collateral held 

publu:
Fixes # .

- fixed erc721 held (qidao vaults)
- added balance of at scaled
-

### DIFF
--- a/src/strategies/ve-balance-of-at-scaled/index.ts
+++ b/src/strategies/ve-balance-of-at-scaled/index.ts
@@ -1,0 +1,35 @@
+import { BigNumberish } from '@ethersproject/bignumber';
+import { formatUnits } from '@ethersproject/units';
+import { Multicaller } from '../../utils';
+
+export const author = 'publu';
+export const version = '0.0.1';
+
+const abi = [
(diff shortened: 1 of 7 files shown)


src/strategies/erc721-collateral-held/index.ts
ChaituVR, line 1: Hi @publu One PR should contain changes of one strategy only.

src/strategies/ve-balance-of-at-scaled/index.ts
```diff
@@ -0,0 +1,35 @@
...
+    multi.call(address, options.address, 'balanceOfAt', [address, blockTag])
+  );
+  const result: Record<string, BigNumberish> = await multi.execute();
+
+  return Object.fromEntries(
+    Object.entries(result).map(([address, balance]) => [
+      address,
+      parseFloat(formatUnits(balance, options.decimals)) * options.scaleValue
+    ])
+  );
```
ChaituVR, line 34: You could you math strategy for this @publu let me know if you have some trouble doing so :)

ChaituVR (changes requested):

PR #1341, opened 2023-11-10
[sd-reverse-delegation] new strategy: sd reverse delegation

pierremarsotlyon1:
Create a reverse delegation strategy

- the delegation address will acquire the whole voting power
- if an user will vote or delegate to another address than the delegation address, the delegation address voting power will decrease

### DIFF
--- a/src/strategies/botto-dao-base/README.md
+++ b/src/strategies/botto-dao-base/README.md
@@ -1,0 +1,33 @@
+# Botto Dao Governance strategy
+
+It returns the BOTTO balance of the voters participating in the governance and in the liquidity mining program
+
+Here is an example of parameters:
+```json
+[
+  {
+    "name": "Botto Dao Governance strategy",
+    "strategy": {
+      "name": "botto-dao",
+      "params": {
+        "token": "0x9dfad1b7102d46b1b197b90095b5c4e9f5845bba",
+        "stakingAddress": "0x19CD3998f106eCC40eE7668c19C47e18b491e8a6",
+        "miningAddress": "0xf8515Cae6915838543bCD7756F39268CE8F853Fd",
+        "liquidityAddress": "0x9ff68f61ca5eb0c6606dc517a9d44001e564bb66",
+        "symbol": "BOTTO",
+        "decimals": 18
+      }
+    },
+    "network": "1",
+    "addresses": [
+      "0xf2b10b41961f6bc3801e7946dabe5572158a78a5",
+      "0xea32cf979bdaf8d3bb4121d58515d4623a27f3e0",
+      "0x7173d417cfa1cdec794fa21f4ca8a71d181b2ca0",
+      "0x58787f1659b2fd8853918dab6d4a1565569e6044",
+      "0x7dbc5f88986db946f1d5987edfbf9c6917230fa2"
+    ],
+    "snapshot": 13808330
+  }
+]
+
+```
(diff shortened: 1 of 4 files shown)


src/strategies/sd-reverse-delegation/index.ts
```diff
@@ -0,0 +1,138 @@
+import { getDelegations } from '../../utils/delegation';
+import { getScoresDirect } from '../../utils';
+import { multicall } from '../../utils';
+import { formatUnits } from '@ethersproject/units';
+import { BigNumber } from '@ethersproject/bignumber';
+import { getAddress } from '@ethersproject/address';
+
+export const author = 'pierremarsotlyon1';
+export const version = '0.1.0';
+export const dependOnOtherAddress = true;
```
ChaituVR, line 10: Hi @pierremarsotlyon1 Right now we are not accepting strategies that depend on other addresses to calculate voting power. these types of strategies are usually not scalable, you see some other way to handle this without depending on other voters

ChaituVR (changes requested):

src/strategies/sd-reverse-delegation/index.ts
ChaituVR: https://github.com/snapshot-labs/snapshot-strategies#code

pierremarsotlyon1: Hey @ChaituVR 
I didn't really understand the issue here
I'm just calculating the voting power of each voters to calculate the voting power of another voter
Which is basically like calculate the voting power of each voters like all the other strategies

For example, if we have these voters : [0x1, 0x2, 0x3] and the strategy will want to calculate the voting power of 0x3 (based on strategy parameters), I will calculate the voting power of 0x1 and 0x2 to assign the voting power of 0x3

What's the issue here ?

ChaituVR: Yes we do have other 15 old strategies that depend on other addresses, but we stopped accepting new ones as they are not scalable and causing lot of trouble, one alternative way is running seperate server to return the voting power of addresses and using the strategy `api-v2-override`

pierremarsotlyon1: But there is something that i don't understand
Currently, with your current default delegation strategy, how you reduce the delegation address voting power if there are some voters who delegated to the delegation address ?

ChaituVR: > Currently, with your current default delegation strategy, how you reduce the delegation address voting power if there are some voters who delegated to the delegation address?

We calculate voting power for all addresses again (that's the pain problem 🙈 ), and then we will just ignore the voting power of the delegator who voted from the delegate's voting power

pierremarsotlyon1: Is it possible to calculate in a external server the voting power of each user and then compute the delegation address voting power in the strategy ?

the delegation address voting power will still be based on the voter addresses but at least the user voting power will be compute externally
So basically, the strategy will just fetch voting powers with an endpoint and then do plus/minus stuff

ChaituVR: Your strategy can do anything within limits mentioned here https://github.com/snapshot-labs/snapshot-strategies#code just that depending on other addresses is not scalable for us

pierremarsotlyon1: Ok, we're blocked with what we want to achieve and your requirements.
Is it really not possible to push this strat ?

Otherwise, is it possible to use your delegation strategy (like currently we have) but use a custom delegation contract address by the strategy settings ? 
Or is it possible to add this logic on your side ?

Basically, our "custom" delegation contract will be the same as yours but with new functions

PR #1345, opened 2023-11-14
[delegate-registry-v2] Add optional v1 support

manboy-eth:
Adds optional support for "delegate registry v1" delegations in the "delegate registry v2" strategy.

### DIFF
--- a/src/strategies/delegate-registry-v2/index.ts
+++ b/src/strategies/delegate-registry-v2/index.ts
@@ -12,6 +12,7 @@ const DEFAULT_BACKEND_URL = 'https://delegate-registry-backend.vercel.app';
 type Params = {
   backendUrl: string;
   strategies: Strategy[];
+  delegationV1VChainIds?: number[]; // add this to include v1 delegations
 };
 
 /*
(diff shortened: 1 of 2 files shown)


src/strategies/delegate-registry-v2/index.ts
```diff
@@ -12,6 +12,7 @@ const DEFAULT_BACKEND_URL = 'https://delegate-registry-backend.vercel.app';
 type Params = {
   backendUrl: string;
   strategies: Strategy[];
+  delegationV1VChainIds?: number[]; // add this to include v1 delegations
```
ChaituVR, line 15: We are not using this anywhere?

ChaituVR (changes requested):
Also PR should be coming from `gnosis` github as they are the author, can fork this  from there? or better to create a new strategy

src/strategies/delegate-registry-v2/index.ts
manboy-eth: It is just forwarded to the backend in line 47 and used there.

ChaituVR: Thanks @manboy-eth missed it

ChaituVR:
I see you are original author of the strategy

ChaituVR:
@manboy-eth Test cases are failing can take a look?
image

manboy-eth:
Closing since I opened a PR from Gnosis: https://github.com/snapshot-labs/snapshot-strategies/pull/1353

I will also address the test error in that PR

PR #1353, opened 2023-11-19
[delegate-registry-v2] Add optional v1 support

manboy-eth:
Adds optional support for "delegate registry v1" delegations in the "delegate registry v2" strategy.

### DIFF
--- a/src/strategies/delegate-registry-v2/examples.json
+++ b/src/strategies/delegate-registry-v2/examples.json
@@ -10,17 +10,19 @@
           {
             "name": "erc20-balance-of",
             "params": {
+              "symbol": "ST",
+              "address": "0xE666Ad68a6e2897CD06A9ff378ED8b0d71093398",
+              "network": "5",
+              "decimals": 18
+            },
+            "network": "5"
+          },
+          {
+            "name": "erc20-balance-of",
+            "params": {
               "address": "0x6b175474e89094c44da98b954eedeac495271d0f",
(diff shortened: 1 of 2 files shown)


ChaituVR (approved):

ChaituVR:
Thanks @manboy-eth some test cases still fail 🙈 can take a look? Looks like no address in examples is returning a positive score
image

ChaituVR:
image
Can also add some addresses that have voting power at the `latest` block  🙈 I tried but couldn't find them

manboy-eth:
> image Can also add some addresses that have voting power at the `latest` block 🙈 I tried but couldn't find them

Aha, yes this is a problem :/ 
I can not easily make that test pass because there are no v2-delegations for the yam.eth context/space on mainnet or gnosis chain, and the yam.eth space does not have the delegate-registry-v2 strategy with v1 delegations enabled (via the `delegationV1VChainIds` strategy pram).

Making this test pass requires one of the following conditions:
1. The delegate-registry-v2 strategy is enabled on the space, and an address holding the token that is used in that space needs to delegate to another address on the mainnet or gnosis chain.
2. The delegate-registry-v2 strategy is enabled on the space, and the delegationV1VChainIds param set (to for instance, `[1, 100]`) on the mainnet or gnosis chain.

ChaituVR:
Is it possible for you to create a test pass to test?

ChaituVR (approved):

PR #1357, opened 2023-11-23
[plearn] Add plearn strategy 

jeeraponnnnnnnn:
- Add a voting strategy for the Plearn project, which calculates voting power based on holdings of the project(pools, token).


src/strategies/plearn/index.ts
```diff
@@ -0,0 +1,214 @@
...
+        addresses.map((address: any) => [
+          item.address,
+          'userInfo',
+          [address],
+          { blockTag }
+        ]),
+        { blockTag }
+      )
+    )
+  );
```
ChaituVR, line 139: Is it possible to limit the number of `lockedPoolAddresses` `foundingInvestorPoolAddresses`, `pendingWithdrawalAddresses` and make sure number of requests doesn't cross 5 https://github.com/snapshot-labs/snapshot-strategies#code

ChaituVR (changes requested):
Also some of test cases are failing, can take a look at them plz? :)

src/strategies/plearn/index.ts
ChaituVR: @jeeraponnnnnnnn

jeeraponnnnnnnn: @ChaituVR 
I've implemented the changes as you suggested. Now, I've limited the arrays for lockedPoolAddresses, foundingInvestorPoolAddresses, and pendingWithdrawalAddresses to have no more than 5 addresses each. Thanks

ChaituVR: But that will max 15 network calls, we can only allow upto max 5 network calls https://github.com/snapshot-labs/snapshot-strategies#code

jeeraponnnnnnnn: @ChaituVR  I've updated code, could you please take another look at it?

ChaituVR (approved):

ChaituVR:
@jeeraponnnnnnnn could you resolve conflicts please :)

PR #1364, opened 2023-11-29
[cake] Update cake strategy for only veCake

chef-jojo:
-  Now cake strategy will only count for veCake balance after bsc block `34371669`


zzuziak (approved):
utACK

PR #1378, opened 2023-12-21
[aavegotchi-agip] Aavegotchi

cinnabarhorse:
Fixes # .

- Add new wearable IDs 
- Update tests


ChaituVR:
I see a few active proposals on aavegotchi space
if we merge this those will be affected
better to wait until those proposals are closed

cinnabarhorse:
The active proposals are likely to fail due to lack of quorum, because they are affected by the issue this PR fixes. @ChaituVR

cinnabarhorse:
Please merge if you don't see any other issues, thanks!

ChaituVR:
But users already voted may have old voting power while new ones will have new voting power 😅

cinnabarhorse:
But the VP will only increase, will not decrease @ChaituVR

ChaituVR:
If we update the voting strategy the users that already voted on current proposals will keep the same voting power but new ones will have voting power based on the updated strategy

It can cause a lot of confusion

ChaituVR:
You think you can communicate this information to users?

cinnabarhorse:
Yes we will communicate the update in our DAO channel. 

If users that have already voted vote again, will the VP be updated?

ChaituVR:
Not by default, but if we remove cache then if users vote again, VP will be updated.

ChaituVR (approved):

ChaituVR:
@cinnabarhorse we just flushed the cache, make sure to communicate with your users to vote again :)

PR #1387, opened 2024-01-10
[hopr-stake-and-balance-qv] Update "HOPR Stake and Balance QV" to version `0.2.0`

QYuQianchen:
Closes https://github.com/hoprnet/hopr-devrel/issues/353.

- This strategy replaces the old "staking season" program with the new "Safe Staking" program, where HOPR nodes' are tied to voting accounts via Safes.
- Update the calculation of tokenholders' balances to be `(B1 + B2 + sum((B_i + S_i) * F_i)) ^ exponent` (see README file for details of each variable)
- Vote power is no longer calculated purely quadratically and can be set with `exponent`


ChaituVR:
Hi @QYuQianchen I feel like it is better to create a new strategy because it looks like a lot of things change, the name should be same.?

QYuQianchen:
Hi @ChaituVR Thanks for raising this concern. There are indeed a lot of changes there. However, as we are deprecating the old staking contracts, the existing strategy (version "0.1.0") will not be used in the future.

ChaituVR (approved):

PR #1405, opened 2024-02-08
[urbit-galaxies] Add Urbit Galaxy Voting - compatible with any Azimuth deployment

kennyrowe:
This strategy allows for the use of a voting proxy with a fallback on the owner of the Urbit galaxy.


src/strategies/intex.ts
ChaituVR, line 1: Could you please delete this file :)

ChaituVR (changes requested):

ChaituVR (approved):

PR #1408, opened 2024-02-15
[giveth] Multichain Supply Weighted Strategy

divine-comedian:
Fixes # .

- Add new strategy 'multichain-supply-weighted'
- fetch circulating supply of token from a REST API
- use 'multichain' logic to get the total balance of a given token across many chains 
- divide the sum of balances by the circulating supply
- apply a weight to the quotient to get the final voting power result


ChaituVR:
Hi @divine-comedian `multichain` strategy will be deprecated shortly. idea is not to allow more than 8 strategies in a space, https://docs.snapshot.org/user-guides/strategies/voting-strategies#how-to-set-up-a-strategy

divine-comedian:
@ChaituVR I did notice this one https://github.com/snapshot-labs/snapshot-strategies/issues/688 but looks like it hasn't moved for over a year. 

I'm a bit stumped if that's the case. Do you have any ideas on this one? It's a bit limiting because I need to get the result of multiple strategies and divide it by the result of another. 

The goal is to get the total balance of an address across all chains and divide it by the circulating supply.

PR #1427, opened 2024-03-11
strategy added  Helios [stake-mine-liquid-helios]

asim07:
Fixes # .

- we have added new strategy for our customised platform voting


.yarnrc.yml
```diff
@@ -0,0 +1 @@
+nodeLinker: node-modules
```
ChaituVR, line 1: Could you revert this file, please

yarn.lock
ChaituVR, line 1: Revert this one also

ChaituVR (changes requested):
We are good to go once you revert these :)

asim07 (commented):
file deleted as oper requirement

yarn.lock
ChaituVR, line 1: We need to revert this file too @asim07

ChaituVR, line 1: We should not delete it, In your local try running `git checkout origin/master -- yarn.lock`

PR #1439, opened 2024-03-19
[metronome] Add Metronome strategy

rokso:
Fixes # None

- Adding new strategy for Metronome

### DIFF
--- a/src/strategies/metronome/index.ts
+++ b/src/strategies/metronome/index.ts
@@ -1,0 +1,15 @@
+import { strategy as votingStrategy } from '../vesper';
+
+export const author = 'rokso';
+export const version = '1.0.0';
+
+export async function strategy(
+  space,
+  network,
+  provider,
+  addresses,
+  options,
+  snapshot
+) {
+  return votingStrategy(space, network, provider, addresses, options, snapshot);
+}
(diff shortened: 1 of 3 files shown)


src/strategies/metronome/index.ts
```diff
@@ -0,0 +1,15 @@
...
+export async function strategy(
+  space,
+  network,
+  provider,
+  addresses,
+  options,
+  snapshot
+) {
+  return votingStrategy(space, network, provider, addresses, options, snapshot);
+}
```
ChaituVR, line 15: Hi @rokso You can directly use `vesper` strategy on your space. let me know if you have some trouble

ChaituVR (changes requested):

src/strategies/metronome/index.ts
rokso: @ChaituVR now that I look into it, I am not sure how it will be fetching strategy address deployed on-chain? How does it tie to voting power address added in examples.json?

rokso: I am fine using erc20-balanceof strategy as well for given snapshot block but how to tie voting power address or token address to existing strategy?

rokso: @ChaituVR I think I know what you are talking, let me have admin try the config in UI. I will update you how that goes. Thanks

ChaituVR: Cool, let me know if you face some issues :) All i mean is to use vesper on your space settings

rokso: Just to confirm, we should be able to use vesper strategy but metronome voting power contract in space settings in UI, correct?

rokso: I don't have access to space and hence I can't do/verify that myself.

ChaituVR: > I don't have access to space and hence I can't do/verify that myself.

I mean you should do it on a space that you have access 😄 even if you add a new strategy, only space admins can use your new strategy

rokso: @ChaituVR Admin was able to use Vesper strategy with new config so all good and this PR is not needed. Closing it.

PR #1449, opened 2024-04-03
[far-staking-balance-of] far-staking-balance-of strategy

cameron-brooks:
Fixes # .

- far-staking-balance-of strategy


cameron-brooks:
> @cameron-brooks I think you can just use a `contract-call` strategy on your space directly. no need of a separate strategy

You are right, I have implemented it using contract call. Please close or delete this PR, thank you.

PR #1452, opened 2024-04-09
Improve `hopr-stake-and-balance-qv` strategy

QYuQianchen:
According to feedback from community members, here are some improvements and bug fixes to the voting strategy.

- Improve error catching in the cross-chain block-number conversion so that the fallback block number can be used whe the subgraph is down.
- Specify type definition for number types. 
- Make subgraph util functions use an explicit block number.
- Update the "safe stake query id" and "latest version string" in the example.json file

### DIFF
--- a/src/strategies/hopr-stake-and-balance-qv/examples.json
+++ b/src/strategies/hopr-stake-and-balance-qv/examples.json
@@ -6,7 +6,7 @@
       "params": {
         "tokenAddress": "0xf5581dfefd8fb0e4aec526be659cfab1f8c781da",
         "symbol": "HOPR",
-        "fallbackGnosisBlock": 27852687,
+        "fallbackGnosisBlock": 33114482,
         "subgraphStudioProdQueryApiKey": null,
         "subgraphStudioDevAccountId": null,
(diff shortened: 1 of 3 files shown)


ChaituVR (changes requested):
I see an active proposal at https://snapshot.org/#/vote.hopr.eth/proposal/0xc7c172d20247df0c255a05f0773e321ef80c98af4f7a7bacec4e9a3a2435ab87 which is using this strategy. i would suggest to merge this after the end of this proposal

QYuQianchen:
Thanks @ChaituVR for the review. Sounds good 👍. A related question: if the strategy implementation gets updated during an ongoing voting, will the scores being calculated according to the newer version or will the is the strategy anchored at a specific IPFS version/commit so it's calculated with an older version?

ChaituVR:
> ill the scores being calculated according to the newer version or will the is the strategy anchored at a

Scores will be calculated with newer version but we have many cache systems, so if someone already tried to vote, they may see older voting power. so we always check if there is a active proposal in case someone want to update existing strategy, or other way is to create a new strategy and use it on new proposals ;)

PR #1481, opened 2024-05-20
[safe-vested] Optimize Strategy: safe-vested 

cristovaoth:
### Summary
This PR improves calculation complexity, making the strategy faster:

- **Allocations Map**: Builds an allocations map in memory once (takes approximately 50ms).
- **Efficient Lookups**: Use the map for retrieving allocations during calculations. For 5000+ addresses, this reduces execution time by around 20 seconds.
- **Optimised Result Construction**: Avoids creating new objects at every step of the result reducer, using simple assignment instead. For +5000 addresses, this saves around 7 seconds.

### Discussion
An additional optimization to discuss with the SAFE team involves the size of the allocations file, currently around 120MB and loaded each time the strategy runs. We were able to trimmed it to approximately 17MB. Two possible options to optimize this:
1. Source the trimmed file alongside the strategy.
2. Make the trimmed file available at http:/safe-claiming-app

### DIFF
--- a/src/strategies/safe-vested/examples.json
+++ b/src/strategies/safe-vested/examples.json
@@ -9,7 +9,7 @@
       }
     },
     "network": "1",
-    "snapshot": 15800000,
+    "snapshot": 17778299,
     "addresses": [
       "0x9970dcab40e29a84D1020DeaEa443dCE1d8471b7",
       "0x1230B3d59858296A31053C1b8562Ecf89A2f888b",
@@ -18,7 +18,7 @@
       "0xf2565317F3Ae8Ae9EA98E9Fe1e7FADC77F823cbD",
       "0x37b828802FAeA4244d176Da386CF632F5Bbc414F",
       "0xA8FFD6B87388F8d5FACfDa0147d9B0Da511539b6",
-      "0xd7539FCdC0aB79a7B688b04387cb128E75cb77Dc"
+      "0x0Ca678b984186b0117501C00d4A6B4F8F342D06D"
     ]
   }
 ]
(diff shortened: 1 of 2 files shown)


ChaituVR:
Hi @cristovaoth Tests are failing, could you please take a look, try running `yarn test --strategy=safe-vested` in your local. looks like example.json is missing addresses that have voting power

cristovaoth:
Hey @ChaituVR 

I've included one address with voting power, which fixes all the tests locally except for the one related to slowness. This issue is being addressed in this PR.

We are currently consulting with the SAFE team to confirm whether it is acceptable for them to source the trimmed allocations JSON file next to the strategy.

Would this be acceptable for Snapshot? The file will be approximately 17 MB.

ChaituVR (approved):

ChaituVR:
> We are currently consulting with the SAFE team to confirm whether it is acceptable for them to source the trimmed allocations JSON file next to the strategy.

> Would this be acceptable for Snapshot? The file will be approximately 17 MB.

Hi @cristovaoth Sounds like a big file but you can add it if it is important :) 

or alternate way is to call it only once and store it in a global variable ([before strategy function](https://github.com/snapshot-labs/snapshot-strategies/blob/master/src/strategies/safe-vested/index.ts#L37)) if JSON is available we can just reuse and if not we will fetch it again

cristovaoth:
Hi @ChaituVR , I agree. I think global variable would be a better approach, especially because SAFE people retain control over the datasource. However, this requires that your services are long-lived (serverless lambdas wouldn't be suitable).

If all is a go, I will address this in a follow up PR

cristovaoth:
Just created a subsequent PR

PR #1483, opened 2024-05-25
[safe-vested] cache allocations

cristovaoth:
It loads the allocations JSON once and stores a trimmed version in cache using a global variable inside the strategy file.


cristovaoth:
@ChaituVR here is the subsequent PR, that implements the optimization the way we agreed on

ChaituVR (approved):

PR #1492, opened 2024-06-05
[garden-stakes] Add garden-stakes strategy

Revantark:
- Add garden-staker strategy which uses votes from garden staker contract  to calculate the vote power.

### Background

Getting the votes from the contract is not straightforward. First, we need to fetch the nonce of the user in the contract, which represents the number of stakes the user has. Then, we calculate an ID called stakeId, which is the hash of the user address and the nonce. Finally, we can get the user's votes by calling the `stakes` function with the stakeId. Users can have multiple stakes, so we need to iterate over all the stakes to get the total votes of the user.

### DIFF
--- a/src/strategies/garden-stakes/index.ts
+++ b/src/strategies/garden-stakes/index.ts
@@ -1,6 +1,6 @@
 import { keccak256 } from '@ethersproject/keccak256';
 import { pack } from '@ethersproject/solidity';
-import { Multicaller, call } from '../../utils';
+import { Multicaller } from '../../utils';
 import { BigNumber } from '@ethersproject/bignumber';
 
 export const author = 'gardenfi';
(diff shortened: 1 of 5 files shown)


src/strategies/garden-stakes/index.ts
```diff
@@ -0,0 +1,77 @@
...
+): Promise<Record<string, number>> {
+  const blockTag = typeof snapshot === 'number' ? snapshot : 'latest';
+
+  const multi = new Multicaller(network, provider, abi, { blockTag });
+  for (const address of addresses) {
+    const nonce = await getNonce(
+      provider,
+      address,
+      options.gardenStakerAddress
+    );
```
ChaituVR, line 34: Could we please move this out of the loop?

ChaituVR (changes requested):

src/strategies/garden-stakes/index.ts
Revantark: Updated to use multi-call

ChaituVR (approved):

PR #1509, opened 2024-06-19
[balance-of-with-max] Updated balance-of-with-max to add weight

thomasscovell:
Updated balance-of-with-max strategy to add an optional weight

- added weight as an option to multiply token total by
- example weight set to 1


thomasscovell:
@ChaituVR as suggested I have updated this strategy to include weight rather than adding a new strategy. Thanks.

ChaituVR (approved):

PR #1513, opened 2024-06-24
[te-academy] add te-academy snapshot strategy

marthendalnunes:
Adds Snapshot Strategy for Token Engineering Academy voting mechanism.

### DIFF
--- a/src/strategies/te-academy/index.ts
+++ b/src/strategies/te-academy/index.ts
@@ -1,0 +1,40 @@
+import fetch from 'cross-fetch';
+
+export const author = 'te-academy';
+export const version = '0.0.1';
+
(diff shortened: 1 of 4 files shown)


src/strategies/te-academy/index.ts
```diff
@@ -0,0 +1,40 @@
+import fetch from 'cross-fetch';
+
+export const author = 'te-academy';
```
ChaituVR, line 3:
```suggestion
export const author = 'marthendalnunes';
```

ChaituVR (changes requested):
Hey @marthendalnunes I think you can use `api-v2` strategy here, could you try using it?

PR #1515, opened 2024-06-26
[csv] add strategy CSV

bonustrack:


src/strategies/csv/index.ts
```diff
@@ -0,0 +1,62 @@
...
+  _space,
+  _network,
+  _provider,
+  addresses,
+  options
+): Promise<Record<string, number>> {
+  const result = await parseCSV(
+    options.csv,
+    options.lineSeparator,
+    options.fieldSeparator
```
ChaituVR, line 47: These fields are missing in the schema file, and will always be same

PR #1523, opened 2024-07-08
[swarm-stacking] new strategy

UnnaturalCabbage:
Hey there guys,

We have [SMT Stacking platform](https://smt.swarm.com/staking) that stores user tokens inside contracts and not in their wallets obsly. Therefore, we simply wanna make things fair and count these tokens too. I guess it's not the first case you have with stacking, so pls guide me if i missed somewhere

### DIFF
--- a/src/strategies/swarm-stacking/index.ts
+++ b/src/strategies/swarm-stacking/index.ts
@@ -3,7 +3,7 @@ import { formatUnits } from '@ethersproject/units';
 import { subgraphRequest } from '../../utils';
 import { getAddress } from '@ethersproject/address';
 
-export const author = 'swarm';
+export const author = 'SwarmMarkets';
 export const version = '1.0.0';
 
 // 137 - Polygon network number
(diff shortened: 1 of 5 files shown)


src/strategies/swarm-stacking/index.ts
```diff
@@ -0,0 +1,88 @@
+import { BigNumber } from '@ethersproject/bignumber';
+import { formatUnits } from '@ethersproject/units';
+import { subgraphRequest } from '../../utils';
+import { getAddress } from '@ethersproject/address';
+
+export const author = 'swarm';
```
ChaituVR, line 6:
```suggestion
export const author = 'SwarmMarkets';
```

ChaituVR (changes requested):

src/strategies/swarm-stacking/index.ts
UnnaturalCabbage: @ChaituVR yep good point, I've updated it rn

ChaituVR (approved):

PR #1529, opened 2024-07-15
Add strategy erc721-with-landtype-multiplier

monish-nagre:
Fixes # .

- This strategy returns the balances of the voters for a specific ERC721 NFT with an arbitrary multiplier based on the type of land they own.
Types Of Land :
Mega contributes 25000 VP 
Large contributes 10000 VP
Medium contributes 4000 VP
Unit contributes 2000 VP

### DIFF
--- a/src/strategies/erc721-with-landtype-multiplier/index.ts
+++ b/src/strategies/erc721-with-landtype-multiplier/index.ts
@@ -1,0 +1,97 @@
+import { multicall } from '../../utils';
+import { BigNumber } from '@ethersproject/bignumber';
+
+export const author = 'monish-nagre';
+export const version = '0.1.0'; 
+
+const abi = [
+  'function balanceOf(address owner) public view returns (uint256)',
+  'function tokenOfOwnerByIndex(address owner, uint256 index) public view returns (uint256)',
+  'function landData(uint256 tokenId) public view returns (uint256 landId, string landType, string x, string y, string z)'
+];
+
+// Voting power based on land type
+const landTypeVotingPower: { [key: string]: number } = {
+  'Mega': 25000,
+  'Large': 10000,
+  'Medium': 4000,
+  'Unit': 2000
+};
+
+export async function strategy(
+  space: string,
+  network: string,
+  provider: any,
+  addresses: string[],
+  options: any,
+  snapshot: number | string
+): Promise<{ [address: string]: number }> {
+  const blockTag = typeof snapshot === 'number' ? snapshot : 'latest';
+
+  try {
+    // Step 1: Get the balance of each address
+    const balanceCalls = addresses.map((address: string) => [options.address, 'balanceOf', [address]]);
+    const balanceResponse = await multicall(network, provider, abi, balanceCalls, { blockTag });
+
+    // Check if balanceResponse is an array and has valid data
+    if (!Array.isArray(balanceResponse) || balanceResponse.length !== addresses.length) {
+      throw new Error('Balance response is not valid');
+    }
+
+    // Parse balance response
+    const balances = balanceResponse.map((response: any) => BigNumber.from(response[0]).toNumber());
+    console.log('Balance response:', balances);
+
+    // Step 2: Get all token IDs for each address
+    const tokenCalls: [string, string, [string, number]][] = [];
+    addresses.forEach((address: string, i: number) => {
+      const balance = balances[i];
+      for (let j = 0; j < balance; j++) {
+        tokenCalls.push([options.address, 'tokenOfOwnerByIndex', [address, j]]);
+      }
+    });
+
+    if (tokenCalls.length === 0) {
+      return {};
+    }
+
+    const tokenResponse = await multicall(network, provider, abi, tokenCalls, { blockTag });
+
+    // Check if tokenResponse is an array and has valid data
+    if (!Array.isArray(tokenResponse)) {
+      throw new Error('Token response is not an array');
+    }
+
+    // Parse token response
+    const tokenIds = tokenResponse.map((response: any) => BigNumber.from(response[0]).toString());
+    console.log('Token response:', tokenIds);
+
+    // Step 3: Get land type for each token ID
+    const landDataCalls: [string, string, [BigNumber]][] = tokenIds.map((tokenId: string) => [options.address, 'landData', [BigNumber.from(tokenId)]]);
+    const landDataResponse = await multicall(network, provider, abi, landDataCalls, { blockTag });
+
+    // Check if landDataResponse is an array and has valid data
+    if (!Array.isArray(landDataResponse) || landDataResponse.length !== tokenIds.length) {
+      throw new Error('Land data response is not valid');
+    }
+
+    // Step 4: Calculate voting power based on land type
+    const votingPower: { [address: string]: number } = {};
+    let tokenIndex = 0;
+    addresses.forEach((address: string, i: number) => {
+      votingPower[address] = 0;
+      const balance = balances[i];
+      for (let j = 0; j < balance; j++) {
+        const landType = landDataResponse[tokenIndex].landType;
+        votingPower[address] += landTypeVotingPower[landType] || 0;
+        tokenIndex++;
+      }
+    });
+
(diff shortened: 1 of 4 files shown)


monish-nagre:
Actually , i'm confuse that my request is accept or not , because the testcase are run as i expected voting power. 
the error is for example.json address range like 3 and etc , i think that doesn't effect my strategy.
Please conform me. 
My request is ready to merge / having issue ?

test/strategy.test.ts
```diff
@@ -78,74 +78,17 @@ describe.each(examples)(
...
-            .map((v) => v.toLowerCase())
-            .includes(address.toLowerCase())
-        )
-      ).toBe(true);
-      // Check if all scores are numbers
-      expect(
-        Object.values(scores[0]).every((val) => typeof val === 'number')
-      ).toBe(true);
-    });
-
```
ChaituVR, line 101: HI @monish-nagre We should remove existing testcases, could you revert this file?

src/strategies/erc721-with-landtype-multiplier/index.ts
```diff
@@ -0,0 +1,97 @@
...
+      votingPower[address] = 0;
+      const balance = balances[i];
+      for (let j = 0; j < balance; j++) {
+        const landType = landDataResponse[tokenIndex].landType;
+        votingPower[address] += landTypeVotingPower[landType] || 0;
+        tokenIndex++;
+      }
+    });
+
+    console.log('Voting power:', votingPower);
```
ChaituVR, line 91:
```suggestion
```

src/strategies/erc721-with-landtype-multiplier/index.ts
```diff
@@ -0,0 +1,97 @@
...
+    console.log('Balance response:', balances);
+
+    // Step 2: Get all token IDs for each address
+    const tokenCalls: [string, string, [string, number]][] = [];
+    addresses.forEach((address: string, i: number) => {
+      const balance = balances[i];
+      for (let j = 0; j < balance; j++) {
+        tokenCalls.push([options.address, 'tokenOfOwnerByIndex', [address, j]]);
+      }
+    });
```
ChaituVR, line 52: if my balance is in millions, we will send million calls to this multicall? 😄 is there some other way?

src/strategies/erc721-with-landtype-multiplier/index.ts
```diff
@@ -0,0 +1,97 @@
...
+    const tokenResponse = await multicall(network, provider, abi, tokenCalls, { blockTag });
+
+    // Check if tokenResponse is an array and has valid data
+    if (!Array.isArray(tokenResponse)) {
+      throw new Error('Token response is not an array');
+    }
+
+    // Parse token response
+    const tokenIds = tokenResponse.map((response: any) => BigNumber.from(response[0]).toString());
+    console.log('Token response:', tokenIds);
```
ChaituVR, line 67:
```suggestion
```

src/strategies/erc721-with-landtype-multiplier/index.ts
```diff
@@ -0,0 +1,97 @@
...
+        tokenIndex++;
+      }
+    });
+
+    console.log('Voting power:', votingPower);
+
+    return votingPower;
+  } catch (error) {
+    return {};
+  }
```
ChaituVR, line 96: we don't need try catch in strategy, it was handled seperately

ChaituVR (changes requested):

test/strategy.test.ts
monish-nagre: Thanks for reply , yes i do

src/strategies/erc721-with-landtype-multiplier/index.ts
monish-nagre: yes

src/strategies/erc721-with-landtype-multiplier/index.ts
ChaituVR: I think we should throw some error if it crosses a certain number, else it may eat all memory, or if there is way to get total supply of the tokens, and if it goes beyond a certain number we can make this strategy to stop working

monish-nagre: add abi :          [  'function totalSupply() external view returns (uint256)' ]

 const MAX_SUPPLY_THRESHOLD = totalSupply;         // Set max supply threshold dynamically
 
 // Check if the number of calls exceeds the maximum threshold
    if (tokenCalls.length > MAX_SUPPLY_THRESHOLD) {
      throw new Error(`Number of token calls (${tokenCalls.length}) exceeds the maximum threshold (${MAX_SUPPLY_THRESHOLD})`);
    }
    
  may i add these changes , it's okay ? i mean is satisfied for strategy

ChaituVR: I think we should set a static MAX_SUPPLY_THRESHOLD, also this may allow few users and block few users 😅 

I think it would be better if we could:
- At the beginning of the strategy, Get totalSupply of contract
- If it is more than 10k or something a bit more
- Return error
- Else continue

monish-nagre: Yes I Understood , 
sorry for above response : ( My current maximum supply of ERC721 tokens is 5000 and the balance is unlikely to be high due to the expensive nature of the lands. ) its not my current supply , is my max limit that the land mint i.e 5000.


In my contract , the maximum LAND MINTING supply is 5000, and currently we sell (land-count) is 150 .
Soo we should set the MAX_SUPPLY_THRESHOLD to 5000.

monish-nagre: Thanks for reply ...
soo now , should i push the updated code ?

ChaituVR: Yes please

monish-nagre: without try-catch my code runs :

Test strategy "erc721-with-landtype-multiplier" with example index 0
    ✓ Strategy name should be lowercase and should not contain any special char expect hyphen (2 ms)
    ✓ Strategy name should be same as in examples.json
    ✓ Addresses in example should be minimum 3 and maximum 20 (1 ms)
    ✓ Must use a snapshot block number in the past (1147 ms)
    ✓ Strategy should run without any errors (416 ms)
    ✕ Should return an array of object with addresses (2 ms)
    ✓ Should take less than 10 sec. to resolve (1 ms)
    ✕ File examples.json should include at least 1 address with a positive score (1 ms)
    ✓ Returned addresses should be checksum addresses (1 ms)
    ✕ Voting power should not depend on other addresses (1016 ms)
  
Test strategy "erc721-with-landtype-multiplier" with example index 0 (latest snapshot)
    ✓ Strategy should run without any errors (2738 ms)
    ✓ Should return an array of object with addresses (3 ms)
  
Test strategy "erc721-with-landtype-multiplier" with example index 0 (with 500 addresses)
    ○ skipped Should work with 500 addresses
    ○ skipped Should take less than 20 sec. to resolve with 500 addresses
  
Other tests with example index 0
    ○ skipped Check schema (if available) is valid with examples.json
    ○ skipped Check schema (if available) all arrays in all levels should contain `items` property
    ○ skipped Strategy should work even when strategy symbol is null
  
Others:
    ✓ Author in strategy should be a valid github username (575 ms)
    ✓ Version in strategy should be a valid string

monish-nagre: Thanks for support ... @ChaituVR 
Yes i pushed the updated code now .

monish-nagre:
Please @ChaituVR , check the updated code.

ChaituVR:
It should pass all test cases @monish-nagre 

```
 call revert exception [ See: https://links.ethers.org/v5-errors-CALL_EXCEPTION ] (method="balanceOf(address)", data="0x", errorArgs=null, errorName=null, errorSignature=null, reason=null, code=CALL_EXCEPTION, version=abi/5.6.4)
 ```

PR #1539, opened 2024-07-24
[rocketpool-node-operator-delegate-v8] Added rocket-pool-node-operator-delegate-v8-strategy

nickdoherty:
Added an updated strategy that fixes an issue with v7.

### DIFF
--- a/src/strategies/rocketpool-node-operator-delegate-v8/index.ts
+++ b/src/strategies/rocketpool-node-operator-delegate-v8/index.ts
@@ -23,6 +23,16 @@ export async function strategy(
   const blockTag =
     typeof snapshot === 'number' ? snapshot : 'latest';
 
+  const validPrefix = 'rocketpool-node-operator-v';
+  if (
+    !options.strategies.some((s) => {
+      const parsedStrategy = JSON.parse(JSON.stringify(s));
+      return parsedStrategy.name.startsWith(validPrefix);
+    })
+  ) {
+    return {};
+  }
+
   const req = await fetch(
     'https://api.rocketpool.net/mainnet/delegates/block/' + blockTag
   );
--- a/src/strategies/rocketpool-node-operator-delegate-v8/index.ts
+++ b/src/strategies/rocketpool-node-operator-delegate-v8/index.ts
@@ -1,0 +1,126 @@
+import fetch from 'cross-fetch';
+import { getScoresDirect, Multicaller } from '../../utils';
+import { getAddress } from '@ethersproject/address';
+
+export const author = 'rocket-pool';
+export const version = '0.1.8';
+export const dependOnOtherAddress = true;
+
+const signerRegistryContractAddress =
+  '0xc1062617d10Ae99E09D941b60746182A87eAB38F';
+const signerRegistryAbi = [
+  'function signerToNode(address) external view returns (address)'
+];
+
+export async function strategy(
+  space,
+  network,
+  provider,
+  addresses,
+  options,
+  snapshot
+): Promise<Record<string, number>> {
+  const blockTag =
+    typeof snapshot === 'number' ? snapshot : 'latest';
+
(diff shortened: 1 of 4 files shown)


src/strategies/rocketpool-node-operator-delegate-v8/index.ts
```diff
@@ -0,0 +1,126 @@
+import fetch from 'cross-fetch';
+import { getScoresDirect, Multicaller } from '../../utils';
+import { getAddress } from '@ethersproject/address';
+
+export const author = 'rocket-pool';
+export const version = '0.1.8';
+export const dependOnOtherAddress = true;
```
ChaituVR, line 7: Hi @nickdoherty We are not accepting any new strategies that depend on other addresses https://github.com/snapshot-labs/snapshot-strategies?tab=readme-ov-file#code you could use delegation strategy on your space instead

ChaituVR (changes requested):

src/strategies/rocketpool-node-operator-delegate-v8/index.ts
nickdoherty: Hey @ChaituVR, we have used the delegation strategies in the past.

Rocket Pool has moved to on-chain governance, but we still intend on using Snapshot for signalling. To achieve this, our on-chain delegation has to match snapshot.

As we are a permissionless protocol, using Snapshot's delegation will allow sinister users to create an inconsistency between on-chain delegation and snapshot delegation.

For example, they may delegate to someone on-chain, but delegate to someone different using snapshot delegation.

This strategy allows us to make them consistent.

v8 of this strategy corrects the override vote behaviour.

I didn't want to be dishonest, so I left in the override addresseses (the last two addresses in the examples) and added the dependOnOtherAddress flag.

If the flag is the issue, I can update the PR with the flag removed.

We'd love it if you could make an exception for this case, you can only use this strategy within the Rocket Pool ecosystem.

ChaituVR: Di you mean you are using snapshot delegation in your contract?

nickdoherty: We can't use snapshot delegation.

Our governance has moved on-chain (using our contracts), however we still intend to use snapshot for signalling.

Our users can delegate their vote power on-chain to other users. We are trying to align the on-chain delegation & vote power with snapshot, so that there aren't any inconsistencies and when they go to vote on a proposal using snapshot, their on-chain vote power matches what they see in snapshot.

Using snapshot delegation, it allows inconsistencies.

For example, user A delegates to user B on chain. User A then delegates to user C using snapshot's delegation. They now have differing vote power on-chain and when they use snapshot.

Previous iterations (v6/v7) of this strategy ensured the consistency, however some our users expressed a need to override their delegate (similar to how the snapshot delegation system works) so we updated this strategy (v8) to allow delegate overrides.

However for it to pass the tests it needs the flag, or the overriding delegates removed from the examples.json file.

We didn't want to be dishonest in how it worked so we left the overrides in and were unaware that you weren't accepting new strategies with the flag.

Let me know if you have any further questions.

src/strategies/rocketpool-node-operator-delegate-v8/index.ts
```diff
@@ -0,0 +1,126 @@
...
+  const delegations: Record<string, string[]> = Object.fromEntries(
+    nodeData.map((node) => [node.nodeAddress, node.delegators])
+  );
+
+  if (Object.keys(delegations).length === 0) return {};
+
+  const scores = (
+    await getScoresDirect(
+      space,
+      options.strategies,
```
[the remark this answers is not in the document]
nickdoherty: Added check for valid strategies based on prefix.

ChaituVR (approved):

PR #1547, opened 2024-08-13
[candy-lockv1-token] feat: Add candy-lockv1-strategy

DramaCrypto:
New strategy for the Candy token lock v1 contracts in the Candy Defi , it returns the balances of the locked Candy tokens.

### DIFF
--- a/src/strategies/candy-lockv1-token/index.ts
+++ b/src/strategies/candy-lockv1-token/index.ts
@@ -3,7 +3,7 @@ import { formatUnits } from '@ethersproject/units';
 import { Multicaller } from '../../utils';
 
 export const author = 'dramacrypto';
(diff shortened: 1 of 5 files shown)


ChaituVR (approved):

src/strategies/candy-lockv1-token/index.ts
```diff
@@ -0,0 +1,65 @@
+import { BigNumber } from '@ethersproject/bignumber';
+import { formatUnits } from '@ethersproject/units';
+import { Multicaller } from '../../utils';
+
+export const author = 'dramacrypto';
+export const version = '0.1.1';
```
ChaituVR, line 6:
```suggestion
export const version = '0.1.0';
```

PR #1550, opened 2024-08-14
[candy-lock-token] feat: Add new strategy "candy-lock-token"

DramaCrypto:


src/strategies/candy-lock-token/index.ts
```diff
@@ -0,0 +1,166 @@
...
+    options.decimals
+  );
+  const v3Scores = await v2_3_scores(
+    network,
+    provider,
+    addresses,
+    snapshot,
+    options.v3_address,
+    options.decimals
+  );
```
ChaituVR, line 158: Looks like they are using `2` multicalls each, which makes it `6` multicalls, but we have a limit of `5` multicalls per strategy https://github.com/snapshot-labs/snapshot-strategies?tab=readme-ov-file#code

ChaituVR (changes requested):

src/strategies/candy-lock-token/index.ts
DramaCrypto: I got it @ChaituVR 
Let me update it!

ChaituVR (approved):

PR #1572, opened 2024-09-09
[sd-gauge-less-vote-boost] Add veSDT boost for MS

pierremarsotlyon1:
Fixes : Add veSDT boost for MS


ChaituVR:
Looks like there are few open proposals at https://snapshot.org/#/sdfxs.eth it is better to wait until these proposals to be closed

ChaituVR (approved):

PR #1578, opened 2024-09-13
Add erable-governance-v1 strategy

vk4vd:
Fixes: N/A

- Added a new custom strategy for Erable Governance, called erable-governance-v1.
- This strategy computes voting power based on the token balances of users, both from their wallets and from staked balances, where staked tokens receive a 2x voting multiplier.
- Strategy follows the ERC20 token balance format.


ChaituVR:
Hi @vk4vd Looks like there are conflicts. could you resolve them? also make sure there are no changes in package-lock or yarn.lock

PR #1583, opened 2024-09-19
[subgraph-split-delegation] New strategy: Subgraph split delegation

xavikh:
The subgraph-split-delegation calculates new Vp for the addresses based on the delegation ratios provided by a Subgraph endpoint and an initial strategy Vp.


src/strategies/subgraph-split-delegation/index.ts
```diff
@@ -0,0 +1,132 @@
+import { getAddress } from '@ethersproject/address';
+import { subgraphRequest, getScoresDirect } from '../../utils';
+import { Strategy } from '@snapshot-labs/snapshot.js/dist/src/voting/types';
+
+export const author = 'aragon';
+export const version = '0.1.0';
+
+const DEFAULT_BACKEND_URL =
+  'https://api.studio.thegraph.com/query/87073/split-delegation/version/latest';
```
ChaituVR, line 9: Hi @xavikh could you share the subgraph id for this? these test URLs are rate limited

src/strategies/subgraph-split-delegation/index.ts
```diff
@@ -0,0 +1,132 @@
...
+  const blockTag = typeof snapshot === 'number' ? snapshot : 'latest';
+  const block = await provider.getBlock(blockTag);
+
+  const subgraphData: { members: Member[] } = await subgraphRequest(
+    options.subgraphUrl,
+    {
+      members: {
+        __args: {
+          block: { number: block.number }
+        },
```
ChaituVR, line 49: I think this subgraph will return only 1000 members at once by default.

ChaituVR (changes requested):

src/strategies/subgraph-split-delegation/index.ts
xavikh: We haven't published the subgraph yet. We were planing on passing the final URL through the strategy options.

src/strategies/subgraph-split-delegation/index.ts
xavikh: You are right! I added a filter to only ask for the addresses needed but also added pagination just in case. Thanks for the heads up

src/strategies/subgraph-split-delegation/index.ts
```diff
@@ -0,0 +1,174 @@
...
+        delegatees: {
+          __args: {
+            where: {
+              context: space,
+              expirationTimestamp_gte: block.timestamp
+            }
+          },
+          delegatee: {
+            address: true
```
ChaituVR, line 160: Not too sure, but i think if a address has more than 1000 delegations this may ignore them?

ChaituVR: If the chuck of addresses have more than 1000 delegators

xavikh: Yeah, I think you are right. I'm going to paginate the delegations too. Thanks!

ChaituVR (approved):

PR #1584, opened 2024-09-19
[polygon-self-staked-pol] New strategy: polygon-self-staked-pol

xavikh:
This strategy calculates the voting power based on the active stake of delegators for the Polygon network

### DIFF
--- a/src/strategies/index.ts
+++ b/src/strategies/index.ts
@@ -1,6 +1,7 @@
 import { readFileSync } from 'fs';
 import path from 'path';
 
+import * as polygonSelfStaked from './polygon-self-staked-pol';
 import * as delegatexyzErc721BalanceOf from './delegatexyz-erc721-balance-of';
 import * as urbitGalaxies from './urbit-galaxies/index';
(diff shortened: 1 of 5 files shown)


xavikh:
I see the strategy is taking more than 20s to run for 500 addresses. I'm trying to look for a different solution.

ChaituVR:
> I see the strategy is taking more than 20s to run for 500 addresses. I'm trying to look for a different solution.

It should be fine for now. I will merge it but feel free to modify if you have any idea to optimize :)

ChaituVR (approved):

ChaituVR:
Looks like https://github.com/aragon/polygon-snapshot-strategy/blob/e003689b5743f0f74dc56fa92adacb04bf710079/src/strategies/polygon-self-staked-pol/index.ts#L39 will not scale well if `nftCounter` returns a high value

xavikh:
Yes, we are considering a different solution with an wrapper contract that aggregates the Vp. That might reduce the strategy to a simple contract call.

PR #1600, opened 2024-10-04
fix: use customFetch instead of fetch

wa0x6e:
Closes https://github.com/snapshot-labs/workflow/issues/236

We already have an existing helper fetch function, supporting timeout

This PR replaces all direct calls to `fetch` from the cross-fetch lib, to our custom fetch helper, to take advantage of the timeout feature.

Now, all fetch calls from score-api have a max timeout of 20s


bonustrack (approved):
utACK

src/strategies/safe-vested/index.ts
```diff
@@ -1,4 +1,3 @@
-import fetch from 'cross-fetch';
```
ChaituVR, line 1: But we use fetch here https://github.com/snapshot-labs/snapshot-strategies/blob/eca18b39bf5093087ee6fdf98a6349dc882fdb8b/src/strategies/safe-vested/index.ts#L45

ChaituVR: i will update it to use customFetch, but JIC if not needed LMK

ChaituVR (approved):
utAck

PR #1604, opened 2024-10-09
[moxie] Adding Strategy For Moxie Voting Power Calculation

Hrishikesh-Thakkar:
- Adding a custom strategy for calculating moxie voting power which involves multiple subgraphs and RPC Calls for Moxie Token
- Adding Moxie as a strategy in the list of strategies

### DIFF
--- a/src/strategies/index.ts
+++ b/src/strategies/index.ts
@@ -456,6 +456,7 @@ import * as pom from './pom';
 import * as superboring from './superboring';
 import * as erableGovernanceV1 from './erable-governance-v1';
 import * as snxMultichain from './snx-multichain';
+import * as moxie from './moxie'
 
 const strategies = {
   'delegatexyz-erc721-balance-of': delegatexyzErc721BalanceOf,
@@ -922,7 +923,8 @@ const strategies = {
   pom,
   superboring,
   'erable-governance-v1': erableGovernanceV1,
-  'snx-multichain': snxMultichain
+  'snx-multichain': snxMultichain,
+  'moxie': moxie
 };
 
 Object.keys(strategies).forEach(function (strategyName) {
--- a/src/strategies/index.ts
+++ b/src/strategies/index.ts
@@ -926,7 +926,7 @@ const strategies = {
   'erable-governance-v1': erableGovernanceV1,
   'world-liberty-financial-erc20-balance-of-votes': worldLibertyFinancial,
   'snx-multichain': snxMultichain,
-   'moxie': moxie
+  'moxie': moxie
 };
 
 Object.keys(strategies).forEach(function (strategyName) {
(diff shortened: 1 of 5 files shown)


src/strategies/moxie/index.ts
```diff
@@ -0,0 +1,190 @@
...
+export const author = 'Hrishikesh-Thakkar';
+export const version = '0.0.1';
+
+const abi = [
+  'function balanceOf(address account) external view returns (uint256)'
+];
+
+const MOXIE_VESTING_SUBGRAPH_URL = "https://api.studio.thegraph.com/query/88457/moxie-vesting/version/latest";
+const MOXIE_PROTOCOL_SUBGRAPH_URL = "https://api.studio.thegraph.com/query/88457/moxie-protocol/version/latest";
+const MOXIE_LIQUIDITY_POOL_SUBGRAPH_URL = "https://api.studio.thegraph.com/query/88457/moxie-liquidity/version/latest";
```
ChaituVR, line 15: Hi @Hrishikesh-Thakkar These URLs are not meant for production use, could you publish and share the URLs of the subgraphs? i can update the code with our proxy server

src/strategies/moxie/index.ts
```diff
@@ -0,0 +1,190 @@
...
+      allAddressesScoreMap[getAddress(address)] += formattedBalance;
+    });
+  };
+
+  // Fetch data from both subgraphs in parallel
+  await Promise.all([
+    fetchSubgraphData(MOXIE_PROTOCOL_SUBGRAPH_URL, protocolSubgraphQuery, processProtocolData, "portfolios"),
+    fetchSubgraphData(MOXIE_LIQUIDITY_POOL_SUBGRAPH_URL, liquidityPoolSubgraphQuery, processLiquidityPoolData, "userPools"),
+    fetchBalances(),
+  ]);
```
ChaituVR, line 177: It looks like the strategy is crossing the limit of 5 API calls/multicalls https://github.com/snapshot-labs/snapshot-strategies?tab=readme-ov-file#code You see some other way around?

ChaituVR (changes requested):

src/strategies/moxie/index.ts
Hrishikesh-Thakkar: Ah I thought the 5 requests included pagination. So would fetching all balances in a single multi-call count as 1 call?

ChaituVR: Yep

src/strategies/index.ts
```diff
@@ -923,8 +924,10 @@ const strategies = {
   pom,
   superboring,
   'erable-governance-v1': erableGovernanceV1,
+  'snx-multichain': snxMultichain,
```
ChaituVR, line 927:
```suggestion
```

Hrishikesh-Thakkar: Done 👍

ChaituVR (approved):

PR #1614, opened 2024-10-19
first commit

jscrui:
Fixes # .

Added an strategy called: 

ERC20-STAKED-AT 

This strategy is designed to calculate the total amount of tokens a user has staked in a staking contract. Originally developed for **Paal AI**, it can be adapted for use with any other staking contract, provided it is compatible with this structure.

### DIFF
--- a/src/strategies/erc20-staked-at/index.ts
+++ b/src/strategies/erc20-staked-at/index.ts
@@ -1,0 +1,44 @@
+
+import { multicall } from '../../utils';
+import { formatUnits } from '@ethersproject/units';
+
+export const author = 'jscrui';
+export const version = '0.1.1';
+
+const abi = ['function shares(address account) view returns (uint256, uint256)'];
+
+/**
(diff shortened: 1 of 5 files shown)


src/strategies/erc20-staked-at/index.ts
```diff
@@ -0,0 +1,44 @@
...
+    );
+    
+    return Object.fromEntries(
+        response.map((value, i) => {                        
+            // Format the voting power using the token decimals
+            const votingPower = parseFloat(formatUnits(value[0], options.decimals)); // Convert using formatUnits
+            //console.log(`${addresses[i]}: ${votingPower} (${symbol})`);
+            return [addresses[i], votingPower];
+        })
+    );
```
ChaituVR, line 43: Hi @jscrui You can use `contract-call` strategy directly in your space instead of a separate strategy

ChaituVR (changes requested):

PR #1621, opened 2024-10-29
[price-per-share-balance-of] New strategy: Price Per Share Strategy

carlosgj94:
This strategy calculates the voting power based on the price per shares of that same contract.

### DIFF
--- a/src/strategies/price-per-share-balance-of/index.ts
+++ b/src/strategies/price-per-share-balance-of/index.ts
@@ -1,0 +1,41 @@
+import { BigNumber } from '@ethersproject/bignumber';
+import { formatUnits } from '@ethersproject/units';
+import { Multicaller } from '../../utils';
+
+export const author = 'bonustrack';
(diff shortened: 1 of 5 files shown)


src/strategies/price-per-share-balance-of/index.ts
```diff
@@ -0,0 +1,41 @@
...
+  const result: Record<string, BigNumber> = await multi.execute();
+
+  const pricePerShare = result[`${options.address}-price-per-share`];
+
+  return Object.fromEntries(
+    Object.entries(result).map(([address, balance]) => [
+      address,
+      parseFloat(formatUnits(balance.mul(pricePerShare), options.decimals))
+    ])
+  );
```
ChaituVR, line 40: Hi @carlosgj94 I think you can use `math` strategy in your space instead of a new strategy. you tried?

ChaituVR (changes requested):

PR #1624, opened 2024-11-05
[fountainhead] Add Fountainhead strategy

d10r:
This strategy can be used for a system which allows gradual unlocking of tokens via streams.
It sums up unlocked, locked and staked tokens.


ChaituVR:
Hi @d10r What's your space? 🤔 

Looks like no space is using your previous strategy
https://snapshot.org/#/strategy/superboring

ChaituVR (approved):

PR #1627, opened 2024-11-07
[coinage-coop] feat: add coinage media coop strategy 

isaac-martin:
Adds a custom strategy that only allows active coinage.media coop members to vote
Fetches a custom api endpoint that checks wallets against our user database


ChaituVR (changes requested):
Hi @isaac-martin I think you can use `api` or `api-v2` directly in your space instead of a new strategy, you tried?

isaac-martin:
> Hi @isaac-martin I think you can use `api` or `api-v2` directly in your space instead of a new strategy, you tried?

youre right @ChaituVR i didnt even see those strategies

PR #1628, opened 2024-11-08
adding kmon staking strategy

umbecanessa:
Added a strategy to count KMON token staked as voting power

### DIFF
--- a/src/strategies/kmon-staked-balance/index.ts
+++ b/src/strategies/kmon-staked-balance/index.ts
@@ -1,0 +1,67 @@
+import { multicall } from '../../utils';
+
+export const author = 'umbecanessa';
+export const version = '0.1.1';
+
+const stakingAbi = [
+  {
+    "inputs": [
+      {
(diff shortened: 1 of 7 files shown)


src/strategies/kmon-staked-balance/index.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  console.log('Executing multicall...');
+  const res = await multicall(network, provider, stakingAbi, calls, { blockTag });
+  console.log('Multicall response:', res);
+
+  // Return the balances as voting power
+  const formattedResult = Object.fromEntries(
+    res.map((value, i) => [
+      addresses[i],
+      parseFloat(value) / 10 ** options.decimals // Format balance using the token decimals
```
ChaituVR, line 61: Hi @umbecanessa You can use our strategy https://snapshot.org/#/strategy/math in your space directly without any need of new strategy, you tried? you have to use `divide` operation

PR #1635, opened 2024-11-20
Erc721 multi registry weighted max

aktoriukas:
New Strategy

This is to calculate the balance of the wallets in a multi-registry contract.
Calculation is weighted with a maxBalance parameter.

Here is an example of parameters:

```{
    "symbol": "Total",
    "tokens": [
    "0x3106a0a076BeDAE847652F42ef07FD58589E001f",
    "0xb986Fd52697f16bE888bFAD2c5bF12cd67Ce834B",
    "0xb2d4F230298Cf68164F6C5DD994068Cbb4C3D335"
    ],
    "maxBalance": 200000,
    "weights": [1, 5, 10]
}
```

### DIFF
--- a/src/strategies/erc721-multi-registry-weighted-max/README.md
+++ b/src/strategies/erc721-multi-registry-weighted-max/README.md
@@ -1,0 +1,16 @@
+This is to calculate the balance of the wallets in a multi-registry contract.
+Calculation is weighted with a maxBalance parameter.
+
+Here is an example of parameters:
+
+```{
+    "symbol": "Total",
+    "tokens": [
+    "0x3106a0a076BeDAE847652F42ef07FD58589E001f",
+    "0xb986Fd52697f16bE888bFAD2c5bF12cd67Ce834B",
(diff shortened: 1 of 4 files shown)


ChaituVR (dismissed):
Hi @aktoriukas I think you can use math strategy in your space instead of a new strategy https://snapshot.org/#/strategy/math did you try?

aktoriukas:
@ChaituVR 
There is not much DOCS on what the `math` strategy suppose todo.

From a brief look i can't see how can i set-up multiple token with different weights and max value using `math` strategy.

Also, why would there be `erc721-multi-registry-weighted` if `math` could do it then ?

as my proposed strategy of `erc721-multi-registry-weighted-max` is just extension of the `erc721-multi-registry-weighted` with a max value added.

ChaituVR:
hi @aktoriukas 

> edit: there is plus in the math strategy.

I don't get it 🤔 there is no plus operation on math strategy, you need to use erc721-multi-registry-weighted with 
a-if-gte-b operation

PR #1642, opened 2024-11-27
[synfutures-v3-voting-power] Add synfutures v3 votingPower strategy

bijianing97:
Summary of changes
- Added synfutures v3 votingPower strategy, all tests are passing for the new strategy, which is called `synfutures-v3-voting-power`

### DIFF
--- a/src/strategies/synfutures-v3-voting-power/index.ts
+++ b/src/strategies/synfutures-v3-voting-power/index.ts
@@ -1,0 +1,43 @@
+import { getAddress } from '@ethersproject/address';
+// import { formatUnits } from '@ethersproject/units';
+import { customFetch } from '../../utils';
+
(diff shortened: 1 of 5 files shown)


src/strategies/synfutures-v3-voting-power/index.ts
ChaituVR, line 1: Hey @bijianing97 i feel like you could use api-v2 strategy on your space instead of creating a new strategy. you tried it?

ChaituVR (changes requested):

src/strategies/synfutures-v3-voting-power/index.ts
bijianing97: Yeah, thanks for you advice, `api-v2` can meet our need, I will close the pull request

PR #1645, opened 2024-12-01
Add erc721-balance-or-delegate Strategy

SheeranJL:
**Description:**
This PR introduces a new Snapshot strategy named erc721-balance-or-delegate.

The strategy allows voting power to be calculated based on the following criteria:

1. A wallet holding at least one NFT from the specified ERC-721 contract.
2. A wallet being delegated voting power by another wallet that holds the NFT.

**Features:**

- Supports both direct NFT ownership and delegated voting power.
- Ensures that voting power is granted to the delegator, the delegatee, or both, depending on the setup.

**Implementation:**

- balanceOf Function: Determines if a wallet directly holds an NFT.
- delegates Function: Checks if a wallet has been delegated voting power by another wallet.
- Unique Voters Set: Combines eligible voters from both direct ownership and delegation into a single set to avoid duplication.

### DIFF
--- a/src/strategies/erc721-balance-or-delegated/README.md
+++ b/src/strategies/erc721-balance-or-delegated/README.md
@@ -1,0 +1,29 @@
+# ERC721 Balance or Delegate Strategy
+
+This strategy allows users to vote if they either hold an ERC721 NFT or have voting power delegated to them from that same NFT contract. It combines direct NFT ownership and delegated voting power to determine the final voting score for addresses, with a maximum of 1 in voting power in both cases.
+
+## How it Works
+
+The strategy uses the following logic:
+1. **Direct Ownership**: An address with a balance greater than 0 for the specified ERC721 contract gets a voting score.
+2. **Delegation**: If an address has voting power delegated to it, it also gets a voting score.
+3. **Combined Voting**: Both the delegator and delegate can vote independently if they meet the conditions above.
+
+## Parameters
+
+| Parameter   | Type   | Description                                                                 |
+|-------------|--------|-----------------------------------------------------------------------------|
+| `address`   | string | The ERC721 contract address to query balances and delegations from.         |
+| `symbol`    | string | A string symbol to identify the strategy.                                   |
+| `decimals`  | number | Number of decimals, usually set to 0 for ERC721 as it's non-fungible tokens.|
+
+## Example Parameters
+
+Below is an example configuration for the strategy:
+
+```json
+{
+  "address": "0x4E911626Fa378cC95fB1A26Cc272c070eF79e4b7",
+  "symbol": "HVAXVC",
+  "decimals": 0
+}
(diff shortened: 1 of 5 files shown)


src/strategies/erc721-balance-or-delegated/README.md
```diff
@@ -0,0 +1,29 @@
+# ERC721 Balance or Delegate Strategy
+
+This strategy allows users to vote if they either hold an ERC721 NFT or have voting power delegated to them from that same NFT contract. It combines direct NFT ownership and delegated voting power to determine the final voting score for addresses, with a maximum of 1 in voting power in both cases.
```
ChaituVR, line 3: Hi @SheeranJL I did some review and looks like instead of a separate strategy, you could use our in built feature of voting validation https://docs.snapshot.box/user-guides/strategies/validation-strategies#voting-validation-in-space-settings 

And use `erc721`  and `contract-call` strategies inside basic validation and `ticket` strategy to give `1` voting power for eligible voters

ChaituVR (changes requested):

src/strategies/erc721-balance-or-delegated/README.md
SheeranJL: Thanks for that @ChaituVR i'll take a look. 

If i create a Voting Validation to achieve this, then I would no longer need to have a Strategy within the 'Strategies' section since the Voting Validation would be checking my rules - however, I don't believe you can choose NOT have at least one Strategy defined in the Strategies section. Is this correct?

Also, is it possible to test a Voting Validation via the UI before enabling it, much like you can test strategies within the 'Strategies' section?

ChaituVR: you need to use `basic` voting validation 
image

and then with minimum score of `1`, select custom strategies 
image

SheeranJL: Hey @ChaituVR - I played around with this Basic Vote Validation setup as per your instructions.

A summary of what I configured: 

1. Ticket strategy with 1 Vote Power 
2. Contract call strategy for **getVotes** on my NFT contract
3. ERC721 call to get holdings of NFT for a given address. 

I then also added points 2, and 3, to the main Strategies tab as their own Strategies alongside the Basic setup since I cannot remove all of the Strategies because it says you need at least one. Duplicating this seemed to have worked and assigned 1 point to all users regardless of their combination of delegation/balance most likely due to the Ticket Validation setup.

I think i'm good now unless you notice anything odd about the way I did this. Thanks so much for your help 👍

https://snapshot.org/#/hvax.eth

Screenshot 2024-12-02 at 14 27 15

ChaituVR: hi @SheeranJL 

- ticket strategy will give every wallet with `1` voting power, you need to remove it from the voting validation and keep the other two (unless intentional)
image

- And you wanted to give `1` voting power to every eligible voter right? so in strategies, you should add a `ticket` strategy and remove other
image

PR #1651, opened 2024-12-09
[botto-dao-base] version 1 botto dao base

agustinjch:
Fixes # .

- Added botto base contracts as a new strategy
- Added examples.json


src/strategies/botto-dao-base/index.ts
```diff
@@ -0,0 +1,57 @@
...
+        const stakedBotto = _formatUnits(balances[adr + '-stakedBotto'] || 0);
+        return [adr, stakedBotto];
+      })
+    );
+
+    return [result];
+  } catch (error) {
+    console.error('Error fetching Base balances:', error);
+    return [defaultScores];
+  }
```
ChaituVR, line 56: We should not add a try-catch because even if there is a network issue, score will be a `0` for someone who have voting power which is wrong

ChaituVR (changes requested):

src/strategies/botto-dao-base/index.ts
agustinjch: makes sense. I just removed it. Thanks for pointing it out

ChaituVR (approved):

src/strategies/botto-dao-base/index.ts
```diff
@@ -0,0 +1,52 @@
...
+): Promise<Record<string, number>[]> {
+  const blockTag = typeof snapshot === 'number' ? snapshot : 'latest';
+  const _formatUnits = (value) =>
+    parseFloat(formatUnits(value, options.decimals));
+
+  // Create default scores object
+  const defaultScores = Object.fromEntries(
+    addresses.map((address) => [address, 0])
+  );
+
```
ChaituVR, line 28:
```suggestion
```

ChaituVR:
@agustinjch It Looks like some tests are failing, could you take a look?

PR #1683, opened 2025-01-24
[spooky-lp-sonic] Updated the network ID for the spooky-lp-sonic strategy and adjusted parameters in the example.json file.

BounebRayan:

- Updated the network ID from 1 to 146 for the spooky-lp-sonic strategy.
- Adjusted parameters in the example.json file.


ChaituVR:
@BounebRayan Could you resolve conflicts, also if `poolIds` is passed in options, then this strategy can send more than 5 requests, you need to throw an error in that case

BounebRayan:
@ChaituVR thanks for the feedback. The poolIds is not passed in options anymore.

ChaituVR:
> @ChaituVR thanks for the feedback. The poolIds is not passed in options anymore.

Looks like this is from previous changes. all Good :)

ChaituVR (approved):

PR #1698, opened 2025-02-14
Strategy for hedgey vesting and lockup contracts

nerfZael:


src/strategies/hedgey-locked/index.ts
```diff
@@ -0,0 +1,48 @@
...
+): Promise<Record<string, number>> {
+  const blockTag = typeof snapshot === 'number' ? snapshot : 'latest';
+
+  const multi = new Multicaller(network, provider, abi, { blockTag });
+
+  for (const contract of options.contracts) {
+    for (const address of addresses) {
+      multi.call(contract + "/" + address, contract, 'lockedBalances', [address, options.token]);
+    }
+  }
```
ChaituVR, line 28: May face memory issues if someone pass too many contracts. could you throw an error if someone pass contracts above a limit

PR #1701, opened 2025-03-06
[sablier-v2] refactor: add subgraphs for Sablier

maxdesalle:
Add new subgraphs for Sablier.

### DIFF
--- a/src/strategies/sablier-v2/configuration.ts
+++ b/src/strategies/sablier-v2/configuration.ts
@@ -24,91 +24,64 @@ const chains = {
 
 const deployments = {
   [chains.ethereum]: {
-    subgraph: [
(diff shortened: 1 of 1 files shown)


ChaituVR (changes requested):
Looks like build/tests are failing for this, as your strategy expects only one subgraph and you are trying to pass a array here


Also you are trying to pass subgraph URLs like this https://thegraph.com/explorer/subgraphs/<SUBGRAPH-ID> but you can use our subgraph service like https://subgrapher.snapshot.org/subgraph/arbitrum/<SUBGRAPH-ID>

maxdesalle:
Should be fixed now @ChaituVR, thanks!

ChaituVR (approved):

PR #1713, opened 2025-04-01
[passport-gated] fix: default scoreThreshold to 0

ChaituVR:
Fixes https://github.com/snapshot-labs/workflow/issues/456

The scoreThreshold was defaulted to `0` previously but later updated to be a required field
https://github.com/snapshot-labs/snapshot-strategies/commit/2615e58789c21757e89f2427009bdfe6c2b576f9

But some spaces are not updating space settings for their reasons, this PR will default it to `0` like before

### DIFF
--- a/src/validations/passport-gated/index.ts
+++ b/src/validations/passport-gated/index.ts
@@ -173,10 +173,8 @@ export default class extends Validation {
   async validate(currentAddress = this.author): Promise<boolean> {
     const requiredStamps = this.params.stamps || [];
     const operator = this.params.operator;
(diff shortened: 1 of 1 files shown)


wa0x6e (approved):
utAck

wa0x6e:
Not sure why it was moved to a required field, should be fine when taking into account this is validation is almost not used

ChaituVR:
> Not sure why it was moved to a required field, should be fine when taking into account this is validation is almost not used

Looks like to go for best practice https://github.com/snapshot-labs/snapshot-strategies/pull/1309#discussion_r1383993375

PR #1714, opened 2025-04-05
[split-delegation] throw error if there is a error in delegate API

ChaituVR:


wa0x6e (approved):
They start returning this for un-finalized block

```
{
    "error": "Block\"22202998\" @ Ethereum is too recent"
}
```

If we don't have a special error code for vp not ready, then just throwing the original error is fine

PR #1716, opened 2025-04-12
feat: add BYOSecTest strategy using getVotingPower()

seb-uniris:
Fixes # .

- Adds custom Snapshot strategy, based on getVotingPower(address) from the ERC20Votes Token contract on Polygon.
- 
-


ChaituVR (changes requested):
Hi @seb-uniris Looks like same can be possible by using `contract-call` strategy directly on your space settings, you tried?

PR #1733, opened 2025-05-15
feat(strategy): add eden-online-override strategy

nicolas-law:
# Description

This PR adds a new strategy, `eden-online-override`, which enables voting eligibility for both:

- **Niji Warrior token holders**: Voting power is calculated as the square root of the balance of NFTs owned.
- **Niji Warrior AI agents**: Each NFT has a unique EOA generated in a TEE environment; these agent addresses are retrieved and allowed to vote on Eden Online’s Snapshot space (`edenonline.eth`).

## How does this strategy work?

- Fetches the list of Niji Warrior agent addresses via `options.addressUrl`.
- If a Niji Warrior agent votes, it is granted a voting power of 1.
- If the owner of the NFT also votes, all votes from their NFTs’ agent addresses are nullified (overridden by the owner’s vote).


src/strategies/eden-online-override/examples.json
```diff
@@ -0,0 +1,24 @@
+[
+  {
+    "name": "Eden Online Override",
+    "strategy": {
+      "name": "eden-online-override",
+      "params": {
+        "symbol": "UBINIJI",
+        "address": "0x66efaf92df6456c3cb810012b2de3fb223c25d0d",
+        "decimals": "0",
```
bonustrack, line 9: This should be a number

```suggestion
        "decimals": 0,
```

nicolas-law: Updated this type

PR #1738, opened 2025-05-28
feat: filter out addresses not supported by strategies and validation

wa0x6e:
Toward https://github.com/snapshot-labs/workflow/issues/562

This PR adds a new property to the strategy and validation object, to define the list of protocols supported (`evm` and/or `starknet`).

It will default to `['evm']` by default, for backward compatibility with existing strategies.

Address not supported will be filtered out before processing the strategy/validation

### Note

- [x] Waiting for https://github.com/snapshot-labs/snapshot.js/pull/1149
- [x] Waiting for https://github.com/snapshot-labs/snapshot.js/pull/1152

### DIFF
--- a/src/strategies/rocketpool-node-operator-delegate-v8/index.ts
+++ b/src/strategies/rocketpool-node-operator-delegate-v8/index.ts
@@ -12,9 +12,11 @@ const signerRegistryAbi = [
   'function signerToNode(address) external view returns (address)'
 ];
 
-const snapshotSecretHeader = sha256(
-  `https://api.rocketpool.net/mainnet/delegates/block/${process.env.SNAPSHOT_API_STRATEGY_SALT}`
-);
+function getSnapshotSecretHeader() {
+  return sha256(
+    `https://api.rocketpool.net/mainnet/delegates/block/${process.env.SNAPSHOT_API_STRATEGY_SALT}`
+  );
+}
 
 export async function strategy(
   space,
@@ -42,7 +44,7 @@ export async function strategy(
     'https://api.rocketpool.net/mainnet/delegates/block/' + blockTag,
     {
       headers: {
-        'X-Snapshot-API-Secret': snapshotSecretHeader
+        'X-Snapshot-API-Secret': getSnapshotSecretHeader()
       }
     }
   );


src/validations/validation.ts
```diff
@@ -25,7 +34,50 @@ export default class Validation {
...
+    );
+  }
+
+  private validateStrategiesLength(): boolean {
+    if (
+      this.hasInnerStrategies &&
+      this.params.strategies?.length > MAX_STRATEGIES_LENGTH
+    ) {
+      throw new Error(`Max number of strategies exceeded`);
+    }
```
ChaituVR, line 80: We can remove this check now, but we can keep it if we want

src/constants.ts
```diff
@@ -0,0 +1,4 @@
+import { Protocol } from './types';
+
+export const DEFAULT_SUPPORTED_PROTOCOLS: Protocol[] = ['evm'];
+export const MAX_STRATEGIES_LENGTH = 8;
```
ChaituVR, line 4:
```suggestion
export const MAX_STRATEGIES_LENGTH = 10;
```

wa0x6e:
Tests are now passing thanks to the dependency update

src/utils/vp.ts
```diff
@@ -54,7 +57,10 @@ export async function getVp(
       if (addresses.length === 0) return {};
     }
 
-    addresses = addresses.map(getAddress);
+    addresses = getFormattedAddressesByProtocol(
+      addresses,
+      strategy.supportedProtocols ?? DEFAULT_SUPPORTED_PROTOCOLS
```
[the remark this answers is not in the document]
wa0x6e: This may raise an error for people using the SDK, and passing a strategies object directly

src/validations/karma-eas-attestation/index.ts
```diff
@@ -70,7 +70,8 @@ export default class extends Validation {
   public description =
     'Use EAS attest.sh to determine if user can create a proposal.';
   public proposalValidationOnly = true;
-  async validate(): Promise<boolean> {
+
+  protected async doValidate(): Promise<boolean> {
```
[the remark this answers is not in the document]
wa0x6e: Here, we're using the template design pattern, there are some widely accepted convention about naming for this abstract function:

Node.js Core
```
// EventEmitter pattern
emit() → _emit()
write() → _write()
```

Express.js
```
// Router implementation
handle() → doHandle()
process() → doProcess()
```

Angular Framework
```
ngOnInit() → doOnInit()
validate() → doValidate()
```

src/utils/vp.ts
```diff
@@ -54,7 +57,10 @@ export async function getVp(
...
+    );
```
ChaituVR, line 63: Sorry for not noticing earlier 🙈 With previous changes, we were not  modifying the length of `addresses`. If an address is not valid, it throws an error; however, now with getFormattedAddressesByProtocol it filters addresses. Can this cause any problems? 


I am mostly worried if we don't send back a response to an address, it will asume `0` as voting power

ChaituVR: it maybe very rare case to see invalid address, but better to throw instead of filtering 

Would it cause any problem if we throw error  instead of filtering? we could write a reusuable function for `validateAddressType`

ChaituVR (commented):
Everything else looks good

src/utils/vp.ts
wa0x6e: now returning an error on invalid address/address from unsupported protocols

ChaituVR (commented):
looking good, just the test case is failing

ChaituVR (approved):
utAck

PR #1739, opened 2025-06-03
[puffer-getpastvotes] Puffer erc20-votes strategy

bxmmm1:
Create a new voting strategy that uses `getPastVotes` from ERC20Votes.


ChaituVR (changes requested):
Hi @bxmmm1 you have to use `snapshot` block to calculate voting power, else it can behave weridly

bxmmm1:
Hey @ChaituVR, could you please check if this solution is okay? Cheers.

PR #1746, opened 2025-06-09
[delegated-ape] new delegated ape strategy

ChaituVR:
Related to https://github.com/snapshot-labs/workflow/issues/572


src/strategies/delegated-ape/index.ts
```diff
@@ -0,0 +1,112 @@
...
+export const author = 'snapshot-labs';
+export const version = '0.1.0';
+
+const abi = [
+  'function delegation(address delegator, bytes32 id) view returns (address delegate)',
+  'function getDelegators(address delegate, bytes32 id) view returns (address[])',
+  'function getEthBalance(address account) public view returns (uint256 balance)'
+];
+
+const CONTRACT_ADDRESS = '0xdd6b74123b2ab93ad701320d3f8d1b92b4fa5202';
```
ChaituVR, line 16: @bonustrack, do you think this should be hardcoded like this or passed in as parameters?

bonustrack:
Delegation contract and ID should be a part of the params

PR #1754, opened 2025-06-16
[forte-staking] Forte staking strategy

oscarsernarosero:


src/strategies/forte-staking/index.ts
```diff
@@ -0,0 +1,74 @@
...
+    Object.entries(stakeResult).map(([address, rawBatches]) => [
+      address,
+      calculateVotingPower(rawBatches, kycResult[address])
+    ])
+  );
+}
+
+function calculateVotingPower(rawBatches: batch[], kycLevel: number): number {
+  let votingPower: bigint = 0n;
+  rawBatches.forEach((batch) => {
```
cirsteve, line 61: This `forEach` could be replaced with `reduce` and ``votingPower` being replaced with `accumulator` so line 61 would be `rawBatches.reduce((acc, batch) => {` and the return value would be `return acc + (stake * BigInt(4 * (daysStaked > 1 ? daysStaked - 1 : 0))) / 1461n;`

cirsteve (commented):
Looks great, 1 small comment

src/strategies/forte-staking/index.ts
oscarsernarosero: Great suggestion. Implemented.

src/strategies/forte-staking/index.ts
```diff
@@ -0,0 +1,76 @@
...
+function calculateVotingPower(rawBatches: batch[], kycLevel: number): number {
+  let preKYCPower: bigint = rawBatches.reduce(
+    (votingPower: bigint, batch: batch): bigint => {
+      const today: number = +new Date();
+      const stakeDate: number = +new Date(Number(batch[timeStamp]._hex) * 1000); // * 1000 to convert to ms
+      const daysStaked: number = Math.floor(
+        (today - stakeDate) / (60 * 60 * 24 * 1000) // (...) / (1 day in ms)
+      );
+      const stake: bigint = BigInt(batch[stakeAmount]._hex);
+      return (votingPower +=
```
cirsteve, line 68: this works however `votingPower` as the accumulator is assigned to whatever is returned by the called function, so `votingPower` does not need to be updated and the called function can return `votingPower + (stake * BigInt(4 * (daysStaked > 1 ? daysStaked - 1 : 0))) / 1461n)`

ChaituVR (approved):

src/strategies/forte-staking/index.ts
```diff
@@ -0,0 +1,112 @@
+import { BigNumber } from '@ethersproject/bignumber';
+import { Multicaller } from '../../utils';
+import { formatUnits } from '@ethersproject/units';
+
+export const author = 'forte';
```
ChaituVR, line 5: Should be a github related to forte 😅  it will help identify you if you want to update this strategy in future

oscarsernarosero: 😅 ok I wasn't aware this had the purpose of giving update privileges. I changed it to my GH user

PR #1760, opened 2025-06-23
fix: get supported protocols from hardcodes strategy module

wa0x6e:
Similar to https://github.com/snapshot-labs/snapshot-strategies/pull/1759


ChaituVR (approved):
utAck

PR #1762, opened 2025-06-25
[overtime] Overtime strategy

vladanthales:
- Create a new strategy for Overtime governance (ex Thales).


src/strategies/overtime/index.ts
```diff
@@ -0,0 +1,46 @@
+import { customFetch } from '../../utils';
+import { strategy as erc20BalanceOfStrategy } from '../erc20-balance-of';
+
+export const author = 'thales-markets';
+export const version = '1.0.0';
+export const dependOnOtherAddress = true;
```
ChaituVR, line 6: Hi @vladanthales We are not accepting new strategies that override due to scaling issue,  Instead you can use `api-v2-override` strategy in your space to call the API instead, let me know if you have any questions

ChaituVR (changes requested):

src/strategies/overtime/index.ts
vladanthales: I have checked `api-v2-override`, but I don't see how I can use that here (that strategy uses API to get scores for addresses). In my strategy, my API returns `EOA`/`smartAccount` map (since the voter can't vote with a smart account, I want to assign that score to EOA). So, score is still based on `balance-of` strategy, just I need map from API.

vladanthales: Okay, I double-checked and removed that part. No need to override `smartAccount` voting power, since user can't vote with smartAccount. Code updated.

PR #1770, opened 2025-07-02
fix: remove delegation support from getVp

wa0x6e:
Splitting https://github.com/snapshot-labs/snapshot-strategies/pull/1767 into multiple PR, for easier review.

---

This PR focuses on only removing delegation support from `getVp`, without any other code refactoring (will come in later PR)

DO no merge after approval, all other dependents PR toward refactoring `getVp` will be merged into this branch first, before deploying to master.

### Test

```
yarn jest test/vp.test.ts
```

PR #1771, opened 2025-07-02
refactor: use `getScoresDirect` to get scores in `getVp`

wa0x6e:
Continuing on https://github.com/snapshot-labs/snapshot-strategies/pull/1770

---

Since the delegation support dropping from `getVp`, the core logic to get scores from strategies is the same as `getScoresDirect`

This PR refactors `getVp` to re-use `getScoresDirect` as much as possible to retrieve score, and avoid duplication.

### Changes

- re-use `getScoresDirect` to get score
- move `getVp` to utils.ts (file is becoming too small to warrant its own file)
- move vp.test.ts to integration/utils.test.ts
- since `getScoresDirect` do not validate the addresses, we lose this feature also on `getVp` (temporary)

The next PR will focus on adding back the address validation to `getScoresDirect`


bonustrack (approved):
utACK

ChaituVR (approved):
tAck

